PR #29, opened 2020-04-13
Enforce no-cache behaviour via Cache-Control header

### LINKED ISSUE #4: Do not cache frontend assets
**Describe the bug**
Today the application returns `cache-control public, max-age=14400` for `bundle.js`. This can cause issues with new releases.

**To Reproduce**
1. Go to Network tab in Chrome devtools
2. Inspect the returned headers

**Expected behavior**
We don't want cache behaviors from our browsers.

### LINKED ISSUE #5: Landing page redirects to user page after sign out on Internet Explorer
**Describe the bug**
User is redirected to login page when signing out, but is redirected back to the user page when he visits the landing page

Even though the user can then see the user page with the links displayed, the user cannot create or modify links while logged out

**To Reproduce**
1. Sign in on IE
2. Sign out
3. Click the logo
4. User redirected to user page

**Expected behavior**
Landing page displayed

 - OS: Windows 10
 - IE 11

yong-jie:
## Problem

Browsers are unnecessarily caching resources and server responses.

This results in `bundle.js` being momentarily outdated on new redeployments. IE's behaviour of caching responses without revalidation also creates problems because even after signing out, it still mistakenly thinks it is still logged in due to outdated auth-check responses.

Closes #4 and closes #5 .

## Solution

Send `Cache-Control: no-store` header for every outgoing response. This is implemented using a middleware that is inserted very early in the pipeline, right before `express.static` is being used. This allows even `bundle.js` to be not be cached.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -55,6 +55,12 @@ initDb()
     // Initialise nodemailer
     initMailer()
 
+    // Site-wide cache control
+    app.use((_, res, next) => {
+      res.header('Cache-Control', 'no-store')
+      next()
+    })
+
     // To serve from build
     app.use(express.static('dist'))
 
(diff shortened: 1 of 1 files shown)


yong-jie:
Have confirmed that this solves the sign-out-not-working issue on IE.

PR #41, opened 2020-04-16
Release 1.13.4 - Send Sentry source-maps on travis build step

yong-jie:
# Login page
- [x] It should respond with invalid email when email does not end with .gov.sg
- [x] Invalid OTP should not log the user in
- [x] Valid OTP should log the user in
- [x] Resend OTP should send a new OTP to user, and invalidate previous OTP
- [x] After trying to enter wrong OTP 3 times, it should respond with OTP not found/expired (a new OTP must be requested)
- [x] Visiting /user should redirect to login page when not logged in
# Sessions
- [x] Shows the homepage if user does not have an existing session
- [x] Redirects to /user if user has an existing session (ie logged in previously on the same browser)
# URL creation
- [x] It should populate the short url input box on the create url modal with a random string when the refresh icon on the short url input box is pressed
- [x] It should show the new short url on the users’ links table when a new link is created
- [x] It should close the create url modal on user page when a new link is created
- [x] It should prevent creation of short urls pointing to long urls hosted on blacklisted domains
# Redirection
- [x] It should redirect to the long url when short url is active
- [x] It should redirect to a 404 page when short url is inactive
- [x] On google analytics, a page view should be recorded for a successful redirection - Page Path: short url , Page Title: long url (Reports > Realtime > Content > Pageviews)
# Editing/Toaster logic
- [x] It should set short url active or inactive immediately when the toggle is switched (any caching for that short url is cleared)
- [x] It should redirect to the amended url immediately when the long url is changed (any caching for that short url is cleared)
- [x] Url is updated/saved when user enters a new url, then click away from the input field
- [x] Url is updated/saved when user enters a new url, then keys “Enter”
- [x] Url is reverts to original when user enters a new url, then keys “Escape”
- [x] Error validation (red underline + helperText) appears when value in edit long url textfield is invalid
- [x] Info Toaster (teal/grey coloured) appears when url is saved
- [x] Error Toaster (red coloured) appears when input url is invalid
- [x] Toasters to disappear when user clicks on the X only
- [x] Toasters to disappear after 5sec
- [x] Toasters to not disappear on clickaway (ie to prevent premature closure when user clickaway to save url)
# QR
- [x] Scan the QR code to check that it points to the correct long url
- [x] Downloaded PNG QR code should be of 1000px x 1000px
- [x] Downloaded SVG QR code should work
- [x] QR Code should scale as per screen size (width below 600px to scale smaller accordingly)
- [x] On mobile browsers: Download PNG button > open in new tab > should return a PNG image
# Logging
- [x] UrlHistory table is populated in database

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ notifications:
   email:
     recipients:
       - $EMAIL_RECIPIENT
-    on_success: always
+    on_success: change
     on_failure: always
 
 install:
--- a/README.md
+++ b/README.md
@@ -89,7 +89,11 @@ After these have been set up, set the environment variables according to the tab
 |SESSION_SECRET|Yes|For hashing browser sessions, e.g. `change-this`|
 |VALID_EMAIL_GLOB_EXPRESSION|Yes|The glob expression used to test if a provided email address is valid. For safety, we have disabled the use of negations, ext-glob, glob stars (`**`) and braces, e.g. `*@youremaildomain.com`|
 |GA_TRACKING_ID|No|The Google Analytics tracking ID, e.g. `UA-12345678-9`|
+|SENTRY_AUTH_TOKEN|No|To get relevant permissions to upload the source maps.|
 |SENTRY_DNS|No|The Sentry DNS used for bug and error tracking. e.g. `https://12345@sentry.io/12345`|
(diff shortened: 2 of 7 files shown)


kylerwsm:
Confirmed on Sentry that sourcemap releases for production and staging environments are working as intended.

PR #47, opened 2020-04-23
Expose pre-signed upload url endpoint

yong-jie:
## Feature

Expose a new endpoint that allows authenticated users to generate pre-signed urls for file uploads.

## Caveats
This feature works with user-level IAM credentials, but the pre-signed URL generated on elastic beanstalk staging does not work, pending some help regarding permission settings for the EB's role.

## Tests

Before merging, the problem of elastic beanstalk pre-signed links not working should be first fixed.

- `AWS_S3_BUCKET` : The bucket name used for storing file uploads.

**New dependencies**:

- `aws-sdk` : Used to generate S3 pre-signed upload links

### DIFF
--- a/src/server/api/user.ts
+++ b/src/server/api/user.ts
@@ -108,6 +108,13 @@ function validateState(req: Express.Request, res: Express.Response, next: Expres
 
 /**
  * Make sure all parameters needed for the pre-signed url request are present.
+ *
+ * @param {string} fileType - File type of the file that is being uploaded.
+ * This must be declared here so that subsequent PUT requests to the pre-signed URL
+ * will pass header checks.
+ * @param {string} key - Name of the entry to be inserted to the S3 bucket. Ensure
+ * that the key does not collide with other files before declaring it here. Otherwise,
+ * it will cause an existing file of the same key to be overridden.
  */
 function validatePresignedUrlRequest(
   req: Express.Request,
--- a/src/server/api/user.ts
+++ b/src/server/api/user.ts
@@ -11,6 +11,7 @@ import { logger } from '../config'
 import { redirectClient } from '../redis'
 import blacklist from '../resources/blacklist'
 import { isHttps, isValidShortUrl } from '../../shared/util/validation'
+import { generatePresignedUrl } from '../util/aws'
 
 const router = Express.Router()
 
@@ -106,6 +107,23 @@ function validateState(req: Express.Request, res: Express.Response, next: Expres
 }
 
 /**
+ * Make sure all parameters needed for the pre-signed url request are present.
+ */
+function validatePresignedUrlRequest(
+  req: Express.Request,
+  res: Express.Response,
+  next: Express.NextFunction,
+) {
+  if (!req.body.fileType || !req.body.key) {
+    res.badRequest(
+      jsonMessage('Some or all required arguments are missing: fileType, key.')
+    )
+    return
+  }
+  next()
+}
+
+/**
  * Endpoint for a user to create a short URL.
  */
 router.post('/url', validateUrls, async (req, res) => {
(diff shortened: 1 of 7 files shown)


src/server/api/user.ts
```diff
@@ -105,6 +106,23 @@ function validateState(req: Express.Request, res: Express.Response, next: Expres
...
+function validatePresignedUrlRequest(
+  req: Express.Request,
+  res: Express.Response,
+  next: Express.NextFunction,
+) {
+  if (!req.body.fileType || !req.body.key) {
+    res.badRequest(
+      jsonMessage('Some or all required arguments are missing: fileType, key.')
+    )
+    return
```
liangyuanruo, line 121: nitpicking here, but it's good practice write it out like this
```suggestion
    res.badRequest(
      return jsonMessage('Some or all required arguments are missing: fileType, key.')
    )
```

src/server/util/aws.ts
```diff
@@ -0,0 +1,36 @@
...
+import { S3 } from 'aws-sdk'
+import { s3Bucket } from '../config'
+
+export const s3 = new S3()
+
+/**
+ * Reformat the pre-signed url to to one that will be accepted by S3.
+ *
+ * This is necessary because we used a gov.sg CNAME and configured S3
+ * to only accept requests that have that gov.sg CNAME as the path.
```
liangyuanruo, line 11: could you remind me again on which configuration resulted in this restriction?

src/server/util/aws.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+const reformatPresignedUrl = (url: string, fileName: string) => {
+  const urlObj = parse(url)
+  const {
+    host,
+    pathname,
+    protocol,
+    search,
+  } = urlObj
+  return `${protocol}//${pathname?.split('/')[1]}.${host}/${fileName}${search}`
```
liangyuanruo, line 24: appreciate the need to write this without hardcoding the S3 URL, but i think it would be more readable to use getters/setters in `url.hostname` and `url.pathname` to construct the new URL.

src/server/api/user.ts
```diff
@@ -105,6 +106,23 @@ function validateState(req: Express.Request, res: Express.Response, next: Expres
   next()
 }
 
+/**
+ * Make sure all parameters needed for the pre-signed url request are present.
```
liangyuanruo, line 110: appreciate some extra explanation for `key` and `fileType`

src/server/util/aws.ts
yong-jie: This happened after we did both the CNAME, SSL cert, and cloudfront configs, so I'm not sure exactly which one caused this behaviour. If I were to guess, it's likely related to [this](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-access-to-amazon-s3/)

liangyuanruo:
> This feature works with user-level IAM credentials, but the pre-signed URL generated on elastic beanstalk staging does not work, pending some help regarding permission settings for the EB's role.

Permissions have been configured

yong-jie:
Thanks for the help. Tested to be working on elastic beanstalk, ready to be merged after review

src/server/util/aws.ts
liangyuanruo: I don’t remember setting that, will investigate

src/server/util/aws.ts
```diff
@@ -0,0 +1,41 @@
...
+  const urlObj = parse(url)
+  const {
+    host,
+    pathname,
+    protocol,
+    search,
+  } = urlObj
+  const newUrl = new URL('https://lorem-ipsum.com')
+  newUrl.protocol = protocol as string
+  newUrl.host = `${pathname?.split('/')[1]}.${host}`
```
liangyuanruo, line 26: Any way to avoid doing a string split?

src/server/util/aws.ts
```diff
@@ -0,0 +1,41 @@
...
+  newUrl.search = search as string
+  return newUrl.href
+}
+
+export const generatePresignedUrl = async (fileName: string, fileType: string) => {
+  const params = {
+    Bucket: s3Bucket,
+    Key: fileName,
+    ContentType: fileType,
+    Expires: 60, // 60 seconds.
```
liangyuanruo, line 37: Does this restrict the duration of uploads to 60s, or is it only the beginning of the upload?

yong-jie: The latter; as long as uploading begins before the expiration time, the whole upload process will succeed. [source](https://docs.aws.amazon.com/en_us/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html#private-content-check-expiration)

src/server/util/aws.ts
yong-jie: frankly speaking i'm not too sure if there are better ways to do this, it's just what i came up with. the goal is to extract the `file[-staging].go.gov.sg` from `/file[-staging].go.gov.sg/<optional directories>/filename`

PR #49, opened 2020-04-23
Avoid race condition in LoginPage component

### LINKED ISSUE #48: Login snackbar persists through the User page
**Describe the bug**
A race condition happens between a server request and a ReactRouter redirection.

When a logged in user refreshes the /user route, LoginPage requests from the backend for a message to be supplied to the snackbar:
```
    get('/api/login/message').then((response) => {
      if (response.ok) {
        response.text().then((text) => {
          if (text) setLoginInfoMessage(text)
        })
      }
    })
```
But once the function resolves, ReactRouter has finished re-routing to the /user route -- displaying a misleading/unnecessary message.

**To Reproduce**
1. Login with email, enter OTP
2. While in `/#/user`, refresh the page

**Expected behavior**
1. `Your OTP might take awhile to get to you.` should not be displayed during a successful login.

**Screenshots**
snackbar-persists

 - OS: Ubuntu 18.04
 - Browser: Chrome

**Additional context**
Digging through the codebase tells me you're reading the environment variables server-side, even the strings intended for snackbar use. Why not just store those in `translation.js` to prevent this race condition?

reiallenramos:
Closes #48 

Changes:
- set a flag to decide whether to display the Login snackbar

Screenshots:
When refreshing the User page:
- if the reroute from /login to /user finishes sooner than the request to get LOGIN_MESSAGE resolves, the snackbar is never shown.
race_condition_fixed


reiallenramos:
@liangyuanruo makes sense. I'm just trying to wrap my head around your [comment](https://github.com/opengovsg/GoGovSG/issues/48#issuecomment-618491034). Does this mean if you change an environment variable in production, you wont have to re-deploy to reflect the changes? Express picks it up instantly? :thinking:

liangyuanruo:
> @liangyuanruo makes sense. I'm just trying to wrap my head around your [comment](https://github.com/opengovsg/GoGovSG/issues/48#issuecomment-618491034). Does this mean if you change an environment variable in production, you wont have to re-deploy to reflect the changes? Express picks it up instantly? 🤔

No, it means that changing the message is considered a system configuration, which does not require a code commit.

PR #50, opened 2020-04-26
Improve invalid email domain toaster message (#31)

### LINKED ISSUE #31: Improve invalid domain toaster messages
**Describe the enhancement**
Currently, our domain expression is fetched from the backend when user is routed to the login page. But since we are using client side routing, there may cases where users fetch the app but not the domain expression (i.e. bad or no connectivity when clicking on the login button). When that happens, the app does not prompt the user when he/she is using an email with an invalid domain.

While our backend does our good job in picking invalid domains, the message shown in the error toaster might not be intuitive enough for the user to think that the domain they are using is invalid. Relevant screenshot below.

_Edit for more context: Some relevant files to look at may be the [verify email](https://github.com/opengovsg/GoGovSG/blob/4077ff7f85a8dbb6ac770a72b502280d2458d237/src/server/api/login.ts#L95-L188) api endpoint (which returns the verification toaster text) and the [environment variables](https://github.com/opengovsg/GoGovSG/blob/master/README.md#setting-up-the-infrastructure) (to find the valid email domain used) in our app._

**Screenshot**
image

fterh:
## Problem

Closes #31.

## Solution

I use a new server-side env var to specify the human-readable email domain and sent that as part of the bad request response on the POST /otp route.

## Before & After Screenshots

**BEFORE**:
Before screenshot is available on issue #31.

**AFTER**:
image

## Tests
-

## Deploy Notes

`VALID_EMAIL_DOMAIN` which should be consistent with line 3 of the client-side translations file.


fterh:
@liangyuanruo I've discussed with Alwyn about this approach and I've force-pushed my changes to the same commit. Is this okay?

LoneRifle:
@fterh I just noticed that you're PR'ing against `master`. Could you switch your branches so that you use `develop` for both compare and base branches? This is an oversight on our part, sorry about that

fterh:
I'm closing this; I created a new PR (#58) instead to do what you suggested @liangyuanruo .

PR #53, opened 2020-04-27
Add `isFile` column in url table

yong-jie:
## Problem

We need to introduce a column in the url table to indicate whether a shortlink is pointing to a file.go.gov.sg object.

## Solution
Create two scripts.  The first would be to add the isFile column in a backwards-compatible way. This is done by making the column nullable(already true by default). After future PRs that update the codebase to start writing into the isFile column, run the second script to backfill and set the not-null constraint.

## Follow up
I'll PR the next step which updates the url model after the migration strategy has been finalised and merged.

### DIFF
--- a/src/server/db_migrations/20200428_backfill_isFile_column.sql
+++ b/src/server/db_migrations/20200428_backfill_isFile_column.sql
@@ -1,0 +1,15 @@
+-- This migration script is to be run only after backend code
+-- has been updated to write into the isFile column. This is because
+-- the column will be made NOT NULL here.
+
+BEGIN TRANSACTION;
+
+UPDATE urls SET "isFile" = false WHERE "isFile" IS NULL;
+
+UPDATE url_histories SET "isFile" = false WHERE "isFile" IS NULL;
+
+ALTER TABLE urls ALTER COLUMN "isFile" SET NOT NULL;
+
+ALTER TABLE url_histories ALTER COLUMN "isFile" SET NOT NULL;
+
+COMMIT;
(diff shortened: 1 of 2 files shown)


yong-jie:
@liangyuanruo I've changed the approach back to the one you were suggesting, involving two steps. The first being to set a nullable column in both tables, and then the second script to backfill + add not-null constraint.

src/server/db_migrations/20200428_backfill_isFile_column.sql
```diff
@@ -0,0 +1,15 @@
+-- This migration script is to be run only after backend code
+-- has been updated to write into the isFile column. This is because
+-- the column will be made NOT NULL here.
+
+BEGIN TRANSACTION;
+
+UPDATE urls SET "isFile" = false;
```
liangyuanruo, line 7:
```suggestion
UPDATE urls SET "isFile" = false WHERE "isFile" IS NULL;
```

src/server/db_migrations/20200428_backfill_isFile_column.sql
```diff
@@ -0,0 +1,15 @@
...
+
+UPDATE url_histories SET "isFile" = false;
```
liangyuanruo, line 9:
```suggestion
UPDATE url_histories SET "isFile" = false WHERE "isFile" IS NULL;
```

src/server/db_migrations/20200428_backfill_isFile_column.sql
liangyuanruo: This is in case someone uploaded a file between launching the new application version and running the backfill

yong-jie:
target branch has been changed to `develop`, requesting review before merge so i can get the next steps up

PR #60, opened 2020-05-04
Introduce transition page while preserving meta tags for crawlers

### LINKED ISSUE #10: Transition page
**Describe feature request**
A few reasons why a transition page would be helpful:
1. **Anti phishing:** Citizens should land on a page to verify go.gov.sg before being led off to long link. Phishers can easily place go.gov.sg in front of their QR codes pointing to malicious domains. 
2. **User experience**: Users can anticipate if the long link is something they want to visit before actually visiting

1. Serve transition page once a user visits any go.gov.sg link.
2. Button click to move past the page
3. If button not clicked, transition page lasts for 5 seconds (To be confirmed by designers if this should be removed because it can be destabilising)
4. Transition page should still serve meta tag information on destination page on social media

In the future, there can be another issue to customise message that appears on the transition page.

Sample screenshot:
image

liangyuanruo:
## Problem

Provide a transition page on redirect for browsers, whilst preserving meta tags for social media crawlers.

Closes #10 

### Detecting a crawler

Crawlers tend to have fairly unique user-agent (UA) strings, which are either terse or have the word "bot" embedded (see [Telegram](https://developers.whatismybrowser.com/useragents/explore/software_name/telegrambot/), [Facebook](https://developers.facebook.com/docs/sharing/webmasters/crawler/) and [Google](https://support.google.com/webmasters/answer/1061943?hl=en)).

To discern between bots/crawlers and human users, we rely on the popular `ua-parser-js` library to parse the user-agent string and determine additional metadata such as

- Operating system
- Browser
- Browser engine

which would be readily available for common browsers but not for crawler UAs. The server only serves the transition page if all three pieces of metadata can be determined with confidence. This method leans towards safety - the server should prefer immediately redirecting when operating under uncertainty, rather than risk serving a transition page to a crawler.

## TODOs

- [x] CSS tweaks
- [x] Mobile view
- [ ] Set a cookie to determine if a browser has been served a transition page recently e.g. last 48h.
- [x] Adjust design based on user tests (currently 3s automatic transition)

## Tests
- [x] [LinkedIn](https://www.linkedin.com/post-inspector/inspect/https:%2F%2Fstaging.go.gov.sg%2Ftransition-page)
- [ ] [Twitter](https://cards-dev.twitter.com/validator)
- [x] [Facebook](https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fstaging.go.gov.sg%2Ftransition-page)
- [x] Slack

Also see [metatags.io](https://metatags.io/).

**Features**:

- Transition page for non-crawler user agents

**Improvements**:

- Server-side views moved into `src` which can take advantage of hot-reloading

## Screenshots

Desktop

image

## Deploy Notes

**New dependencies**:

- `ua-parser-js` : User-agent string parsing library

**New dev dependencies**:

- `copyfiles` : Used to move server-side EJS template files into `/build` after Typescript compilation


src/server/views/transition-page.ejs
```diff
@@ -0,0 +1,139 @@
...
+            </div>
+
+        </div>
+        <div id="bottom-container">
+            <img src="https://gosg-public.s3-ap-southeast-1.amazonaws.com/go-logo.svg"/>
+        </div>
+        <script>
+            setTimeout(function() {
+                window.location = '<%- longUrl %>'
+            }, 3000)
```
kylerwsm, line 135: Should we also template out the duration before redirecting to `redirect.ts`?

src/server/index.ts
```diff
@@ -68,6 +70,7 @@ initDb()
       app.set('trust proxy', trustProxy)
     }
 
+    app.set('views', path.resolve(__dirname, './views'))
```
kylerwsm, line 73: I am curious to find out before this line of code and the `postbuild` line in `package.json`, how was 404 pages served in previous builds?

src/server/views/transition-page.ejs
liangyuanruo: potentially, yeah it could be an optional environment variable

src/server/index.ts
liangyuanruo: it previously relied on the default view folder being located in the top level directory where the node process was launched.

PR #63, opened 2020-05-05
Allow S3 object ACL to be toggled when a shortlink is activated/deactivated

yong-jie:
## Problem

We need a way to toggle the visibility of the s3 object. This is because our bucket defaults to `private`. After a short link has been reserved and the file has been uploaded to S3, the client will need to call the edit-url-state endpoint once to trigger this hook, setting the s3 object to `visible`.

## Solution

Use a hook in the active/inactive edit API endpoint to also toggle the visibility of the S3 object, if the url has `isFile` set to `true`

### DIFF
--- a/src/server/api/user.ts
+++ b/src/server/api/user.ts
@@ -11,7 +11,8 @@ import { logger } from '../config'
 import { redirectClient } from '../redis'
 import blacklist from '../resources/blacklist'
 import { isHttps, isValidShortUrl } from '../../shared/util/validation'
-import { PRIVATE, PUBLIC, generatePresignedUrl, setS3ObjectACL } from '../util/aws'
+import { FileVisibility, generatePresignedUrl, setS3ObjectACL } from '../util/aws'
+const { Public, Private } = FileVisibility
 
 const router = Express.Router()
 
@@ -313,7 +314,7 @@ router.patch('/url', validateState, async (req, res) => {
 
     if (url.isFile) {
       // Toggle the ACL of the S3 object
-      await setS3ObjectACL(url.shortUrl, state === ACTIVE ? PUBLIC : PRIVATE)
+      await setS3ObjectACL(url.shortUrl, state === ACTIVE ? Public : Private)
     }
   } catch (error) {
     logger.error(`Error rendering URL active/inactive:\t${error}`)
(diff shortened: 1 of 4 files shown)


src/server/api/user.ts
```diff
@@ -11,7 +11,7 @@ import { logger } from '../config'
 import { redirectClient } from '../redis'
 import blacklist from '../resources/blacklist'
 import { isHttps, isValidShortUrl } from '../../shared/util/validation'
-import { generatePresignedUrl } from '../util/aws'
+import { PRIVATE, PUBLIC, generatePresignedUrl, setS3ObjectACL } from '../util/aws'
```
liangyuanruo, line 14: consider exporting an `enum` object containing `PUBLIC`/`PRIVATE` instead?

src/server/util/aws.ts
```diff
@@ -1,6 +1,11 @@
 import { URL, parse } from 'url'
 import { S3 } from 'aws-sdk'
-import { s3Bucket } from '../config'
+import { logger, s3Bucket } from '../config'
+
+// Types for S3 object ACL toggling. Do not change string representations.
```
liangyuanruo, line 5: check out typescript string enums
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

src/server/util/aws.ts
```diff
@@ -39,3 +44,20 @@ export const generatePresignedUrl = async (fileName: string, fileType: string) =
...
 }
+
+export const setS3ObjectACL = (key: string, acl: VisibilityType): Promise<void> => {
+  const params = {
+    Bucket: s3Bucket,
+    Key: key,
+    ACL: acl,
+  }
+  return new Promise((res, rej) => {
+    s3.putObjectAcl(params, (err) => {
```
liangyuanruo, line 55: i believe the AWS SDK already supports async/await, could you check please?

src/server/api/user.ts
```diff
@@ -310,6 +310,11 @@ router.patch('/url', validateState, async (req, res) => {
         }
       })
     }
+
+    if (url.isFile) {
```
liangyuanruo, line 314: this operation needs to be wrapped in a transaction because of two-phase commit issues.

src/server/api/user.ts
yong-jie: yes that was braindead of me to forget the existence of enums

yong-jie:
sry this PR doesnt work properly on staging, will convert to draft first to fix

yong-jie:
Changes requested have been made. Issues that surfaced during a test on staging have also been fixed, making this PR ready for review.

PR #68, opened 2020-05-07
Use session to store shortUrl visits, avoiding repeated transition page redirects

yong-jie:
## Problem

We need a way to not show the transition page multiple times for the same shortUrl.

## Solution

Use session to store the shortUrl, and read it before deciding to redirect directly or show transition page. Sessions are much easier to work with compared to cookies, as we can see how difficult parsing the cookie was with GA.

### DIFF
--- a/src/server/api/redirect.ts
+++ b/src/server/api/redirect.ts
@@ -192,25 +192,20 @@ export default async function redirect(
       return
     }
 
-    // Reassure typescript that the session object exists.
-    if (!req.session) {
-      throw new Error('Session object does not exist')
-    }
-
-    if (!req.session.visits || !req.session.visits[shortUrl]) {
+    if (!req.session!.visits || !req.session!.visits[shortUrl]) {
       // This is the first time visiting a/the shortlink.
-      req.session.visits = {
-        ...req.session.visits,
+      req.session!.visits = {
+        ...req.session!.visits,
         [shortUrl]: Math.floor(Date.now()),
       }
 
       // Avoid exceeding cookie size limit by pruning
       // old entries.
-      if (_.size(req.session.visits) > MAX_SHORTURL_COUNT) {
+      if (_.size(req.session!.visits) > MAX_SHORTURL_COUNT) {
         // Build inverse dictionary
         const epochList: number[] = []
         const lookupTable: EpochToShortUrlMapping = {}
-        Object.keys(req.session.visits).forEach((url) => {
+        Object.keys(req.session!.visits).forEach((url) => {
           const epoch = req.session!.visits[url]
           lookupTable[epoch] = url
           epochList.push(epoch)
@@ -218,7 +213,7 @@ export default async function redirect(
 
         const earliestEpoch = epochList.sort()[0]
         const earliestShortUrl = lookupTable[earliestEpoch]
-        delete req.session.visits[earliestShortUrl]
+        delete req.session!.visits[earliestShortUrl]
       }
 
       // Extract root domain from long url.
@@ -234,7 +229,7 @@ export default async function redirect(
 
     // User has visited this shortlink before.
     // Update LRU cache and redirect.
-    req.session.visits[shortUrl] = Math.floor(Date.now())
+    req.session!.visits[shortUrl] = Math.floor(Date.now())
     res.status(302).redirect(longUrl)
   } catch (error) {
     if (!(error instanceof NotFoundError)) {
--- a/src/server/api/redirect.ts
+++ b/src/server/api/redirect.ts
@@ -17,6 +17,10 @@ const TRANSITION_PATH = 'transition-page.ejs'
 // limited by max cookie size of 4096 bytes
 const MAX_SHORTURL_COUNT = 100
 
+interface epochToShortUrlMapping {
+  [epoch: number]: string
+}
+
 /**
  *
  * @param {Object} req Express request object.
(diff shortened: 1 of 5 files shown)


yong-jie:
PR has been updated to use cookie-session. A sliding window is also used to prune old shortUrl visits.

src/server/api/redirect.ts
```diff
@@ -171,18 +180,52 @@ export default async function redirect(req: Express.Request, res: Express.Respon
     if (gaTrackingId) gaLogging(req, res, shortUrl, longUrl)
 
     // Redirect immediately if a crawler is visiting the site
-    if (isCrawler(req.headers["user-agent"] || '')) {
+    if (isCrawler(req.headers['user-agent'] || '')) {
```
liangyuanruo, line 183: Apparently this is the canonical way to obtain the header according to [Express docs](https://expressjs.com/en/api.html)
```suggestion
    if (isCrawler(req.get('user-agent') || '')) {
```

src/server/api/redirect.ts
```diff
@@ -171,18 +180,52 @@ export default async function redirect(req: Express.Request, res: Express.Respon
...
       res.status(302).redirect(longUrl)
       return
-    } else {
-      // Serve transition page
-      // TODO: Get/set a browser cookie so that this isn't served too frequently
+    }
+
+    // Reassure typescript that the session object exists.
+    if (!req.session) {
```
liangyuanruo, line 189: we should simply use `req.session!.property`, instead of modifying our program's behavior and writing code for the compiler.

src/server/api/redirect.ts
```diff
@@ -171,18 +180,52 @@ export default async function redirect(req: Express.Request, res: Express.Respon
...
+
+    if (!req.session.visits || !req.session.visits[shortUrl]) {
+      // This is the first time visiting a/the shortlink.
+      req.session.visits = {
+        ...req.session.visits,
+        [shortUrl]: Math.floor(Date.now()),
+      }
+
+      // Avoid exceeding cookie size limit by pruning
+      // old entries.
```
liangyuanruo, line 201: I have a concern with limiting by an arbitrary number of short links, which is susceptible to variability in the length of each short link, when we are able to compute exactly the number of bytes that will be taken up.

As the characters that make up the short links and epoch are a subset of ASCII, the size in bytes is also the length of the string. Can we use that instead to guarantee keeping under the 4k limit?

We might also leave some buffer and use just e.g. 3k.

FInally, could this be extracted into a separate function?

src/server/api/redirect.ts
```diff
@@ -171,18 +180,52 @@ export default async function redirect(req: Express.Request, res: Express.Respon
...
       res.status(200)
         .render(TRANSITION_PATH, {
           shortUrl,
           longUrl,
         })
+      return
     }
+
+    // User has visited this shortlink before.
+    // Update LRU cache and redirect.
```
liangyuanruo, line 226: how about writing an LRU cache class with a `maxSize` parameter in bytes? This would help to manage complexity in the codebase. Apart from regular get/set functions, you could easily extend to add a `toObject()` function that allows reassignment back to `req.session`.

yong-jie:
Rebased the branch on top of kyler's recently-merged PR. Implemented LRU cache class that abstracts away all the storage, pruning, get/set logic. The LRU cache now tracks storage size based on the length of the json-string, with a conservative max size of 2000 bytes.

src/server/util/lrucache.ts
```diff
@@ -0,0 +1,98 @@
+import _ from 'lodash'
+
+const MAX_SIZE_BYTES = 2000
+
+interface CookieData {
```
liangyuanruo, line 5: let's strive to make this LRUCache class standalone as an independent data structure - it currently violates the _Single Responsibility Principle_ because it knows a lot about the format of the cookies, mapping epoch to short links etc.

we could also use _types_ such as [Record<K, T>](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt) instead of interfaces here.

src/server/util/lrucache.ts
```diff
@@ -0,0 +1,98 @@
...
+interface EpochToShortUrlMapping {
+  [epoch: number]: string
+}
+
+export default class LRUCache {
+  data: CookieData
+
+  lookupTable: EpochToShortUrlMapping = {}
+
+  timestamps: number[] = []
```
liangyuanruo, line 18: checkout [this implementation](https://medium.com/sparkles-blog/a-simple-lru-cache-in-typescript-cba0d9807c40) of an LRUCache. it relies on the fact that Javascript `Map`s are actually ordered, so the implementation can be simplified significantly.

liangyuanruo: actually the correct data structure here should be a Priority Queue.

when a redirect request arrives, the server simply iterates through the existing key-value pairs and insert all records, which will be automatically sorted, then upserts the existing short link/epoch (upsert is achieved by removing and reinserting).

to build up the new cookie, simply take from the PQ until iterator is exhausted or the size limit is hit.

does this make sense? if it does perhaps it will be easier to look for an existing implementation instead of rolling our own.

liangyuanruo: Can look at Priority Queues from [google closure](https://google.github.io/closure-library/api/goog.structs.PriorityQueue.html) or [BucketJS](http://mauriciosantos.github.io/Buckets-JS/symbols/buckets.PriorityQueue.html)

liangyuanruo: thinking yet again - maybe we're over-engineering here. looking at this pattern:

```javascript
const cache = new LRUCache(req.session!.visits)
cache.updateEntry(shortUrl)
req.session!.visits = cache.getData()
```

we simply need a reducer function

```javascript
req.session!.visits = updateCookies(req.session!.visits || {}, shortUrl)
```

where

```javascript
type updateCookies = (visits: Record<string, number>, shortUrl: string) => Record<string, number>
```

PR #69, opened 2020-05-10
User page toolbar redesign and code improvements

kylerwsm:
## Problem

User page toolbar redesign, and code de-coupling and improvements.

## Solution

- Re-designing the toolbar and its components while reusing most logic from before.
- Following the v2 design, the *Download* icon button will not be shown in mobile view. Also in mobile view, the *Create link* button will be minimised into an icon button.
- Convert huge components become single-directory components.

**Improvements**:

- (For components which uses app margins but are not able to make use of the `ApplyAppMargins` container component) App margin values are now exposed using the `useAppMargins` hook rather than the `fetchAppMargins` function from before. Doing this allows the consumer components to better manipulate the app margins.
- Table rows now follow app margins.
- Removed some redundant eslint-disable statements.
- Changed some components from using absolute path to relative path, to get better IDE refactor support.
- Combined `tableTheme.js` into `theme.js`. As a result the link row slider now follows the app theme bluish color.

## Before & After Screenshots
**BEFORE**:
image
[Before: Mobile view](https://user-images.githubusercontent.com/28863045/81542049-9f9de200-93a6-11ea-9be7-667786db51e5.png)

**AFTER**:
image
[After: Mobile view](https://user-images.githubusercontent.com/28863045/81541974-85640400-93a6-11ea-9d67-4f8d4fcdcf3e.png)

## Tests
- [x] Search bar should search links.
- [x] Add link button should open the dialog to create a new short link.
- [x] Add link button should be minimise to an icon button on mobile displays.
- [x] Download button (only displayed on wider displays) should download all links in shown table.
- [x] When browser width changes, toolbar should collapse/expand gracefully.

## Deploy Notes

**New dependencies**:

- `@types/lodash` : Type file for lodash


yong-jie:
Quick query: i see that for imports there are some changes that remove paths like `'~/a/b/c'` to `'../../'` etc, is this something recommended by prettier etc? i'm seeing deeply nested components importing a really long string of ../../../../ etc when it could be ~/something etc

kylerwsm:
> Quick query: i see that for imports there are some changes that remove paths like `'~/a/b/c'` to `'../../'` etc, is this something recommended by prettier etc? i'm seeing deeply nested components importing a really long string of ../../../../ etc when it could be ~/something etc

It is not a recommendation by prettier but a preference because of better refactor support from VSCode. I also understand the benefits in using absolute paths. I can change it back to absolute paths if that is more desirable.

yong-jie:
i've yet to test refactor for the `~` variant, but `../../` one definitely works on intelliJ, so i'm personally fine with this change

src/client/components/UserPage/UserLinkTable/ToolBar/SearchInput/index.jsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react'
+import { connect, useSelector } from 'react-redux'
+import {
+  InputAdornment,
+  TextField,
+  createStyles,
+  makeStyles,
+} from '@material-ui/core'
+import debounce from 'lodash/debounce'
```
yong-jie, line 9: I think the convention is to `import _ from 'lodash'` then use `_.debounce()`. Might need to find other usages of this in your branch and amend them. paging @liangyuanruo for confirmation on this convention.

src/client/components/UserPage/UserLinkTable/ToolBar/SearchInput/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+      height: '100%',
+      minHeight: (props) => props.textFieldHeight,
+      padding: theme.spacing(0),
+      lineHeight: 1.5,
+    },
+  }),
+)
+
+// Height of the text field in the search input.
+const textFieldHeight = 48
```
yong-jie, line 45: full caps like `TEXT_FIELD_HEIGHT` for constants like this

src/client/components/UserPage/UserLinkTable/UrlTable/MemoTablePagination/index.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { TablePagination } from '@material-ui/core'
+import isMatch from 'lodash/isMatch'
```
yong-jie, line 3: same thing with this lodash use

src/client/components/UserPage/UserLinkTable/ToolBar/SearchInput/index.jsx
kylerwsm: Solved in commit 7843904.

src/client/components/UserPage/UserLinkTable/ToolBar/SearchInput/index.jsx
kylerwsm: I followed this convention from our legacy code. It seems like both ways are featured on [lodash website](https://lodash.com/). We might also benefit keeping cherry-picked methods, as it minimises bundle size without the need of webpack tree-shaking.

kylerwsm: Waiting for confirmation as to whether to change as well!

liangyuanruo: to benefit from tree-shaking, you'll have to use one of two syntax.

```javascript
import { debounce } from 'lodash'
import debounce from 'lodash/debounce'
```

@yong-jie we can integrate this as a [rule](https://github.com/wix/eslint-plugin-lodash/blob/HEAD/docs/rules/import-scope.md) in ESLint

kylerwsm:
Force pushing to rebase to develop branch

src/server/api/login/handlers.ts
```diff
@@ -83,7 +83,6 @@ export async function generateOtp(req: Express.Request, res: Express.Response) {
   } catch (error) {
     logger.error(`OTP generation failed unexpectedly:\t${error}`)
     res.serverError(jsonMessage('OTP generation failed unexpectedly.'))
-    return
```
yong-jie, line 86: This was the only change since my last review.

PR #72, opened 2020-05-12
Transition Page Redesign

kylerwsm:
## Problem

Completed several TODOs from the [transition-page pull request](https://github.com/opengovsg/GoGovSG/pull/60), includes:
* CSS tweaks
* Mobile view
* Adjust design based on user tests (currently 3s automatic transition)

## Solution

**Features**:

- <s>Created a `parseDomain` function that uses regex to extract root domain (domain and tld) from an input url. Used to show the user the root domain of the destination page on the transition page. If the url cannot be parsed, the function will return the input long url as a fallback for robustness.</s> Made use of the `URL` class to parse hostname from long urls. As a fallback, if the long url is not a valid url, the long url will be used. The extracted hostname will be displayed on the transition page.

**Improvements**:

- Converted page to adopt a mobile-first approach.
- Shifted styles into a dedicated `.css` file.
- Styles and icons are now served from the `public` directory.

## Before & After Screenshots

**AFTER**:
[iPhone 8](https://user-images.githubusercontent.com/28863045/81675808-e0b6f480-9481-11ea-8017-5525d583a988.png), [iPhone 11](https://user-images.githubusercontent.com/28863045/81675447-6b4b2400-9481-11ea-9ec3-0a6f0bc72ab8.png), [Desktop view](https://user-images.githubusercontent.com/28863045/81675465-70a86e80-9481-11ea-8ed9-d552fcec743f.png).

## Tests

- [x] Clicking on the button on the transition page brings users to the correct destination page.
- [x] The root domain shown in the transition page is correct.
- [x] The transition page should not redirect users unless the button is clicked.


src/server/views/transition-page.ejs
```diff
@@ -1,139 +1,50 @@
...
+            </button>
         </div>
-        <div id="bottom-container">
-            <img src="https://gosg-public.s3-ap-southeast-1.amazonaws.com/go-logo.svg"/>
+    </main>
+    <div id="page-banner">
+        <div id="page-banner-content">
+            <!-- banner info icon -->
+            <img src="./assets/transition-page/icons/icon-info.svg" draggable="false" />
+            <p>You will only be shown this page the first time you access this short link.</p>
```
liangyuanruo, line 40: We may have to reword this sentence because not tracking the full history of the links visited. FYI  @yong-jie

public/assets/transition-page/styles/transition-page.css
```diff
@@ -0,0 +1,153 @@
...
+  #main-message h3 {
+    font-size: 22px;
+  }
+
+  #main-container {
+    margin-top: 174px;
+  }
+
+  #page-banner {
+    top: 0;
```
liangyuanruo, line 147: that's a nifty trick to move the banner to the top without CSS Grid

src/server/util/domain.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Extracts the root domain from an input url.
+ * @param url The url for its root domain to be extracted.
+ * @returns The root domain is url is valid, else the unchanged url.
+ */
+export default function parseDomain(url: string): string {
+  const regex: RegExp = /^((http|https):\/\/)([^/]*)/
```
liangyuanruo, line 7: Can we not use the standard `url` library from NodeJS?

kylerwsm: Sounds good to me, will work on it

kylerwsm: Converted to using the `url` library in commit c6c9a28. Added fallback for invalid url in commit f375729.

PR #77, opened 2020-05-13
Re-designed NotFoundPage

kylerwsm:
## Problem

Re-designed the not-found-page with a mobile-first approach.

**Improvements**:

- Some Eslint changes.

## Before & After Screenshots

**AFTER**:
[Desktop view](https://user-images.githubusercontent.com/28863045/81852413-8efb9080-958d-11ea-93cf-c26f101756ef.png), [iPhone 8](https://user-images.githubusercontent.com/28863045/81852423-91f68100-958d-11ea-9a50-6d4513d6de2d.png), [iPhone 11](https://user-images.githubusercontent.com/28863045/81852483-ad618c00-958d-11ea-83a2-195a19218d10.png).

## Edit

- Improved how transition page looks on shorter device (like mobile landscape view) by adding a bottom margin.


public/assets/transition-page/styles/transition-page.css
```diff
@@ -129,7 +130,7 @@ img {
   margin-left: 5px;
 }
 
-@media only screen and (min-width: 600px) {
+@media only screen and (min-width: 960px) {
```
liangyuanruo, line 133: can we use more canonical breakpoints such as one of [these](https://ricostacruz.com/til/css-media-query-breakpoints)?

public/assets/transition-page/styles/transition-page.css
```diff
@@ -71,6 +71,7 @@ button {
 
 #main-container {
   margin-top: min(106px, 12vh);
+  margin-bottom: min(106px, calc(12vh + 61px));
```
liangyuanruo, line 74: not sure if such complex calculations are entirely necessary.

public/assets/transition-page/styles/transition-page.css
kylerwsm: Sound great, but doing that might affect consistency a little as the 600px / 960px values are referenced from default MUI breakpoints that we are using in our frontend React app.

liangyuanruo: Ok. I see that these are the [defaults](https://material-ui.com/customization/breakpoints/) in MUI

public/assets/transition-page/styles/transition-page.css
kylerwsm: Okay, I am assuming that the complex part is `calc(12vh + 61px`. The `61px` (which is the fixed height for the bottom banner) is used to make sure at super short device height, the bottom banner does not cover the main content on the page (as the bottom banner is fixed at the bottom using `position: fixed`, and height is "floating"). That is the best I could think of so far, and will improve it if something better comes to mind!

Issue #75, opened 2020-05-13
Pre-push hook allows improper formatting to be pushed


JasonChong96:
**Describe the bug**
The current pre-push hook runs `npm run lint` which runs `eslint` with the `--fix` flag. This allows the linter to run successfully even though the pushed code contains formatting issues that are fixed by the `--fix` flag. Since the commit has already been done, the flag does nothing to truly resolve the issue.

**To Reproduce**
1. Change any `.ts`/`.js` file to have 4 space indentation
2. Create a commit
3. Push to upstream (Github)

**Expected behavior**
The pre-push hook should have exited with a non-zero exit code to signify that there are formatting issues.

**Additional context**
One way to fix this would be removing the `--fix` flag from the `lint` script and creating another script with the flag for developers.

PR #78, opened 2020-05-14
Fix Pre-push hook to disallow improper formatting

JasonChong96:
## Problem
The current pre-push hook runs npm run lint which runs eslint with the --fix flag. This allows the linter to run successfully even though the pushed code contains formatting issues that are fixed by the --fix flag. Since the commit has already been done, the flag does nothing to truly resolve the issue.

Closes [#75]

## Solution

- Changed the `lint`/`lint-ts`/`lint-js` scripts to exclude the `--fix` flag.
- Added `lint:fix`/`lint-ts:fix`/`lint-js:fix` which include the `--fix` flag.

This way, the pre-push hook will be running the linter without the `--fix` flag and disallow improper formatting from getting pushed.

## Tests

1. Change any .ts/.js file to have 4 space indentation
2. Create a commit
3. Attempt to push to upstream (Github)
4. The pre-push commit should exit with a non-zero exit code and the code should not be pushed.


package.json
```diff
@@ -4,9 +4,12 @@
...
   "main": "src/server/index.js",
   "scripts": {
-    "lint-js": "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.json --ignore-path .gitignore --fix .",
-    "lint-ts": "eslint --ext .ts --no-eslintrc -c .eslintrc.ts.json --ignore-path .gitignore --fix .",
+    "lint-js": "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.json --ignore-path .gitignore .",
+    "lint-ts": "eslint --ext .ts --no-eslintrc -c .eslintrc.ts.json --ignore-path .gitignore .",
+    "lint-js:fix": "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.json --ignore-path .gitignore --fix .",
+    "lint-ts:fix": "eslint --ext .ts --no-eslintrc -c .eslintrc.ts.json --ignore-path .gitignore --fix .",
     "lint": "npm run lint-js && npm run lint-ts",
+    "lint:fix": "npm run lint-js:fix && npm run lint-ts:fix",
```
kylerwsm, line 12: Noticed the naming used for linting with auto-fix. I believe the naming convention used for these operations is `format`.

LoneRifle: `lint` is fine

liangyuanruo: how about `lint` for fixing and `lint:ci` for CI?

JasonChong96: I'm concerned it might not be immediately obvious what `lint:ci` does. I think that having `lint` and `lint:fix` would make it more obvious.

liangyuanruo: OK, just trying to save some keystrokes here.

PR #79, opened 2020-05-14
Use array for transition page cookie

yong-jie:
## Problem

Usage of an object coupled with LRU cache is might not be the best way to track the a user's visit history.

## Solution

As discussed, this is the approach using an ordered array. This PR uses the newly-introduced dependency injection IoC mechanism to inject the functions. Need @JasonChong96 help to check if I'm doing this correctly.

**Important testing note**:

If you have tested the transition page using the old code (json object), you might need to clear your cookies since this one relies on it being an array instead.


JasonChong96:
IoC LGTM

src/server/util/transitionPage.ts
```diff
@@ -0,0 +1,59 @@
+import _ from 'lodash'
+import { injectable } from 'inversify'
+
+/**
+ * Herein contains utility functions pertaining to
+ * the functionality of the transition page.
+ */
+
+const MAX_SIZE_BYTES = 2000
```
JasonChong96, line 9: Would it be better to have this in config.ts?

src/server/util/transitionPage.ts
```diff
@@ -0,0 +1,59 @@
...
+    shortUrl: string,
+  ): string[] {
+    if (!cookie) return [shortUrl]
+    if (cookie.includes(shortUrl)) {
+      return _.without(cookie, shortUrl).concat(shortUrl)
+    }
+    let newCookie = _.concat(cookie, shortUrl)
+    while (newCookie.toString().length > MAX_SIZE_BYTES) {
+      newCookie = _.drop(newCookie)
+    }
```
kylerwsm: should we take into account of the overhead when converting to base64 later on over here?

src/server/util/transitionPage.ts
```diff
@@ -0,0 +1,58 @@
...
+    shortUrl: string,
+  ): string[] {
+    if (!cookie) return [shortUrl]
+    if (cookie.includes(shortUrl)) {
+      return _.without(cookie, shortUrl).concat(shortUrl)
+    }
+    let newCookie = _.concat(cookie, shortUrl)
+    while (newCookie.toString().length > cookieSessionMaxSizeBytes) {
+      newCookie = _.drop(newCookie)
+    }
```
kylerwsm, line 55: should we take into account of the base64 encoding overhead here?

src/server/util/transitionPage.ts
```diff
@@ -47,7 +47,7 @@ export class CookieArrayReducer implements CookieReducer {
     shortUrl: string,
   ): string[] {
     if (!cookie) return [shortUrl]
-    if (this.userHasVisitedShortlink(cookie, shortUrl)) {
+    if (cookie.includes(shortUrl)) {
```
liangyuanruo, line 49: what's the weird inversify bug?

yong-jie: even though it's declared as a member of this class, when it is `bind`ed by inversify, `this` resolves to `undefined`

JasonChong96: Using an arrow function avoids javascript's weird 'this' binding behavior

PR #90, opened 2020-05-18
Format typescript code with prettier

yong-jie:
## Problem

Our eslint config for typescript files does not contain prettier extensions. This leads to conflicting errors between raw eslint and prettier.

## Solution

Edit the eslint config for typescript to contain prettier extensions.

### DIFF
--- a/.eslintrc.ts.json
+++ b/.eslintrc.ts.json
@@ -1,6 +1,11 @@
 {
   "parser": "@typescript-eslint/parser",
-  "extends": ["airbnb", "plugin:import/typescript"],
+  "extends": [
+    "airbnb",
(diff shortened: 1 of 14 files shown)


.eslintrc.ts.json
```diff
@@ -1,6 +1,11 @@
 {
   "parser": "@typescript-eslint/parser",
-  "extends": ["airbnb", "plugin:import/typescript"],
+  "extends": [
+    "airbnb",
+    "plugin:import/typescript",
+    "prettier/@typescript-eslint",
+    "plugin:prettier/recommended"
```
kylerwsm, line 7: Looks good!

PR #94, opened 2020-05-19
Add additional tests on redirect middleware

JasonChong96:
## Problem

With the addition of coveralls, we're able to easily identify untested branches. There were untested branches in the redirect middleware arising from the addition of transition pages.

## Solution

Add tests with differing results from cookie reducer and user agent string.

**Improvements**:

- Backend test coverage +1% 🎉
- Resolve uncaught promise rejections in `redirect.ts`

### DIFF
--- a/src/server/api/redirect.ts
+++ b/src/server/api/redirect.ts
@@ -35,11 +35,9 @@ async function getLongUrlFromStore(shortUrl: string): Promise<string> {
   } catch {
     // Cache failed, look in database
     const longUrl = await getLongUrlFromDatabase(shortUrl)
-    try {
-      await cacheShortUrl(shortUrl, longUrl)
-    } catch (err) {
-      logger.error(err)
-    }
+    cacheShortUrl(shortUrl, longUrl).catch((error) =>
+      logger.error(`Unable to cache short URL: ${error}`),
+    )
     return longUrl
   }
 }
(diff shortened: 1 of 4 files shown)


src/server/api/redirect.ts
```diff
@@ -36,7 +36,7 @@ async function getLongUrlFromStore(shortUrl: string): Promise<string> {
     // Cache failed, look in database
     const longUrl = await getLongUrlFromDatabase(shortUrl)
     try {
-      cacheShortUrl(shortUrl, longUrl)
+      await cacheShortUrl(shortUrl, longUrl)
```
liangyuanruo, line 39: we actually want to redirect ASAP, and not await on this operation.

liangyuanruo: how about `cacheShortUrl(shortUrl, longUrl).catch(...)` instead?

good catch.

test/server/api/redirect.test.ts
```diff
@@ -33,24 +36,97 @@ describe('redirect API tests', () => {
...
+      .to(CookieArrayReducerMockVisited)
   })
 
   afterEach(() => {
     container.unbindAll()
     loggerErrorSpy.mockClear()
   })
 
-  test('url exists in cache and db', async () => {
+  test('url exists in cache and db, real user unvisited', async () => {
```
liangyuanruo, line 47: i notice a lot of bind/unbind going on. would snapshots help?

https://github.com/inversify/InversifyJS/blob/master/wiki/container_snapshots.md

src/server/api/redirect.ts
JasonChong96: Changed to the async catch

test/server/api/redirect.test.ts
JasonChong96: I opted not to use it as almost all tests in this file have a different combination of dependencies.

PR #96, opened 2020-05-19
Server side implementation of file uploading feature

yong-jie:
## Problem

Under the premise that url objects cannot be deleted, the `creation of a link + upload file to s3` step cannot be atomic when it is done client side. 

## Solution

This PR modifies the shortUrl creation endpoint to accept files as an optional parameter that enables the file-uploading process to S3.

## Additional Changes
This PR also introduces Joi validation to protect the endpoint handlers.  

## Environment Variables
These new env vars will be introduced in docker-compose.yml to allow aws-sdk to interface with S3. They are necessary for development only when working with file links. These env vars will become obsolete once #117 is complete.
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY


src/server/index.ts
```diff
@@ -101,6 +102,8 @@ initDb()
       bodyParser.urlencoded({ extended: false }),
       // application/json
       bodyParser.json(),
+      // multipart/form-data
+      fileUpload(),
```
liangyuanruo, line 106: could you analyse the available options for Busboy here? we definitely need a maximum file size, otherwise a user can cause the server to go OOM.

liangyuanruo: also to clarify - `bodyParser.json()` parses fields out of the `multipart/form-data` request, and `fileUpload()` parses the files?

src/server/api/user.ts
```diff
@@ -155,17 +136,21 @@ router.post('/url', validateUrls, async (req, res) => {
...
           longUrl,
           shortUrl,
           isFile: !!isFile,
         },
         { transaction: t },
-      ),
-    )
+      )
+      if (isFile) {
+        await uploadFileToS3(file.data, shortUrl, file.mimetype)
```
liangyuanruo, line 150: is the file extension being saved to the S3 bucket?

yong-jie: nope. The S3 bucket only has the shortUrl as the key. the only hint that S3 gets is the ContentType which is obtained from the mimetype argument in this function call. This will let S3 behave accordingly depending on the file type. For example, if it received image/jpeg, it will display the image in the browser tab instead of initiating a download, etc.

src/server/index.ts
yong-jie: nope. even without `bodyParser`, `fileUpload()` will parse the fields from the `multipart/form-data` and insert them into the `req.body` object. I've verified this to be true by commenting out the `bodyParser.json()` and `bodyParser.urlencoded()` middlewares; `req.body` is still populated by the form data fields.

src/server/api/user.ts
liangyuanruo: Interesting. What file types are available?

yong-jie: my bad, after digging through docs and web resources, it appears that S3 does not inherently behave differently to different Content-Type headers. Instead this is stored and forwarded to all browsers that attempt to retrieve the object. It is then up to the browsers to behave accordingly, such as displaying the content directly if it's a pdf or image, and initiating a download if it's an executable or a video file, etc.

Quoting from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), _Important: Browsers use the MIME type, not the file extension, to determine how to process a URL, so it's important that web servers send the correct MIME type in the response's Content-Type header. If this is not correctly configured, browsers are likely to misinterpret the contents of files and sites will not work correctly, and downloaded files may be mishandled._

The list of MIME types are also available in the MDN link above.

src/server/index.ts
liangyuanruo: I see. we should consider carefully where to place this middleware. only endpoints that ought to accept file uploads should have this middleware present i.e. the POST and PATCH methods of the API.

I was under the impression our frontend has only been sending JSON at this point - which API endpoint requires _application/x-www-form-urlencoded_?

yong-jie: not too sure; after removing the urlencoded middleware, it looks like the site still works and I'm able to do basic operations like creating/editing links. will commit the removal

src/server/api/user.ts
```diff
@@ -80,6 +86,19 @@ function validateUrls(
     return
   }
 
+  // An upload request must contain a file
+  if (isFile) {
+    if (!req.files?.file) {
+      res.badRequest(jsonMessage('Missing file to upload.'))
+      return
+    }
+    const { isValidS3Shortlink } = container.get<S3Interface>(DependencyIds.s3)
```
liangyuanruo, line 95: resolve dependencies during module initialisation at the top of the file, instead of during runtime.

src/server/api/user.ts
```diff
@@ -112,33 +131,19 @@ function validateState(
...
-}
-
-/**
- * Endpoint for a user to create a short URL.
- */
-router.post('/url', validateUrls, async (req, res) => {
+router.post('/url', fileUploadMiddleware, validateUrls, async (req, res) => {
   const { isFile, userId, longUrl, shortUrl } = req.body
+  // @ts-ignore Type definition does not know about the compulsory file field.
+  const file: UploadedFile | undefined = req.files?.file
```
liangyuanruo, line 146: I see that the type of `req.files?.file` is  `fileUpload.UploadedFile | fileUpload.UploadedFile[] | undefined`. under what circumstances will it be an array? should we be addressing the potential types using type guards instead of overriding the compiler?

src/server/api/user.ts
```diff
@@ -155,17 +160,22 @@ router.post('/url', validateUrls, async (req, res) => {
...
           longUrl,
           shortUrl,
           isFile: !!isFile,
         },
         { transaction: t },
-      ),
-    )
+      )
+      if (isFile && file) {
+        const { uploadFileToS3 } = container.get<S3Interface>(DependencyIds.s3)
```
liangyuanruo, line 174: please resolve required functions during module initialisation

src/server/api/user.ts
```diff
@@ -238,42 +234,64 @@ router.patch('/url/ownership', async (req, res) => {
...
-      where: { id: userId },
-    })
+router.patch(
+  '/url/edit',
+  fileUploadMiddleware,
+  validateUrls,
+  async (req, res) => {
+    const { userId, longUrl, shortUrl } = req.body
+    // @ts-ignore Type definition does not know about the compulsory file field.
+    const file: UploadedFile | undefined = req.files?.file
```
liangyuanruo, line 250: same question to be asked here.

src/server/api/user.ts
```diff
@@ -238,42 +234,64 @@ router.patch('/url/ownership', async (req, res) => {
...
+        )
+      }
 
-    await transaction((t) => url.update({ longUrl }, { transaction: t }))
-    res.ok(jsonMessage(`Short link "${shortUrl}" has been updated`))
+      await transaction(async (t) => {
+        if (!url.isFile) {
+          await url.update({ longUrl }, { transaction: t })
+        } else if (file) {
+          const { uploadFileToS3 } = container.get<S3Interface>(
```
liangyuanruo, line 275: same here

src/server/api/user.ts
```diff
@@ -303,6 +321,7 @@ router.patch('/url', validateState, async (req, res) => {
       await url.update({ state }, { transaction: t })
       if (url.isFile) {
         // Toggle the ACL of the S3 object
+        const { setS3ObjectACL } = container.get<S3Interface>(DependencyIds.s3)
```
liangyuanruo, line 324: same here

src/server/util/aws.ts
```diff
@@ -10,48 +11,39 @@ export enum FileVisibility {
...
-  const newUrl = new URL('https://lorem-ipsum.com')
-  newUrl.protocol = protocol as string
-  newUrl.host = `${pathname?.split('/')[1]}.${host}`
-  newUrl.pathname = fileName
-  newUrl.search = search as string
-  return newUrl.href
+export interface S3Interface {
+  setS3ObjectACL: (key: string, acl: FileVisibility) => Promise<any>
+  uploadFileToS3: (file: Buffer, key: string, fileType: string) => Promise<any>
+  isValidS3Shortlink: (url: string, shortUrl: string) => boolean
```
liangyuanruo, line 17: can we adopt AWS types here instead of `Promise<any>`?

src/server/util/aws.ts
```diff
@@ -10,48 +11,39 @@ export enum FileVisibility {
...
-  fileType: string,
-) => {
-  const params = {
-    Bucket: s3Bucket,
-    Key: fileName,
-    ContentType: fileType,
-    Expires: 60, // 60 seconds.
+@injectable()
+/* eslint class-methods-use-this: ["error", { "exceptMethods":
+  ["setS3ObjectACL", "uploadFileToS3", "isValidS3Shortlink"] }] */
```
liangyuanruo, line 22: The methods in the class can be declared _static_, then there would be no need to circumvent ESLint

yong-jie: Looks like using the `static` keyword would cause the typescript error where the method is not recognised as one that would fulfill the `implements` keyword. [This is because when a class implements an interface, only the instance side of the class is checked.](https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes)

src/server/api/user.ts
yong-jie: Okay. Even though I haven't been able to reproduce the scenario where the field manifests as an array, adding type guards are ultimately the safer way to go.

src/server/api/user.ts
liangyuanruo: `container.get()` should be called with the imports at the top of file as resolving dependencies can be an expensive op

src/server/api/user.ts
```diff
@@ -1,192 +1,144 @@
...
+      }
 
-    if (!user) {
-      res.notFound(jsonMessage('User not found'))
-      return
-    }
+      // Success
+      const { buildFileLongUrl, uploadFileToS3 } = container.get<S3Interface>(
+        DependencyIds.s3,
+      )
```
liangyuanruo, line 118: Move this resolve outside of the runtime middleware too.

src/server/api/user.ts
```diff
@@ -1,192 +1,144 @@
...
+        const url = Url.create(
+          {
+            userId: user.id,
+            longUrl: file ? buildFileLongUrl(shortUrl) : longUrl,
+            shortUrl,
+            isFile: !!file,
+          },
+          { transaction: t },
+        )
+        if (file && !Array.isArray(file)) {
```
liangyuanruo, line 129: why do we need to check for an array here?

liangyuanruo: if so, should the check be performed at the start of the middleware with potential 4xx response, instead of here?

src/server/api/user.ts
yong-jie: When i tried to do dependency resolution at the top outside the functions, i got errors which mention something along the lines of container not having any corresponding functions binded yet. I'll see if that problem is solvable, otherwise we might be forced to stick with this way of resolving dependencies

liangyuanruo: is it a problem with the order of instantiation of dependencies in the application? the container needs to be injected first before the middleware imports them

yong-jie: indeed, i've pushed a commit that fixes this by bringing the binding process to the very top of the `index.ts` file so that it happens before other imports that depend on inversify's container

src/server/api/redirect.ts
```diff
@@ -8,7 +8,7 @@ import { UrlCache } from './cache/url'
 import { UrlRepository } from './repositories/url'
 import { DependencyIds } from '../constants'
 import { AnalyticsLogger } from './analytics/analyticsLogger'
-import { CookieReducer } from '../util/transitionPage'
+import { CookieReducer } from '../util/transition-page'
```
kylerwsm, line 11: Can I check why is there a change in naming convention here?

src/server/api/login/validators.ts
```diff
@@ -0,0 +1,36 @@
...
+  email: Joi.string()
+    .custom((email: string, helpers) => {
+      if (!isValidGovEmail(email)) {
+        return helpers.message({
+          message: 'Invalid email provided. Email domain is not whitelisted.',
+        })
+      }
+      return email
+    })
+    .required(),
```
kylerwsm, line 35: Okay I like joi now

src/types/server/api/user.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type NewUserEmailProperty = {
+  newUserEmail: string
+}
+
+type StateProperty = {
+  state: 'ACTIVE' | 'INACTIVE'
+}
+
+export type UrlCreationRequest = ShortUrlOperationProperty &
```
JasonChong96, line 23: Nitpicking but would it be more appropriate to name the types as `RequestBody` rather than just `Request` so that it is less ambiguous?

docker-compose.yml
```diff
@@ -30,6 +30,8 @@ services:
       - LOGIN_MESSAGE=Your OTP might take awhile to get to you.
       - AWS_S3_BUCKET=file-staging.go.gov.sg
       - ROTATED_LINKS=whatsapp,passport,spsc,sppr
+      - AWS_ACCESS_KEY_ID
+      - AWS_SECRET_ACCESS_KEY
```
kylerwsm, line 34: Since these docker env vars do not map to any value, is there any purpose for them being here?

yong-jie: They serve two purposes. First is to tell docker-compose that these two variables can be read from the terminal's environment and passed into the containers. Source [here](https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers)

The second purpose is more indirect, which is to let developers know that we are expecting these two environment variables to be provided one way or another.

src/server/api/redirect.ts
yong-jie: this is to maintain consistency with other filenames in our codebase. Check out sequelize-mock.d.ts and box-icon.d.ts

src/types/server/api/user.d.ts
yong-jie: Makes sense. I'm following the naming convention set in [handlers.d.ts](https://github.com/opengovsg/GoGovSG/blob/73b9ae86b80df2befd4c7a13bd95162c84697eee/src/types/server/api/login/handlers.d.ts#L9) and hence just using `Request` here. Perhaps it's worth checking this code in first and then do another separate PR to refactor everything all at once?

src/server/api/login/validators.ts
yong-jie: Indeed, declarative validation > imperative validation

PR #100, opened 2020-05-19
Implement Desktop View Drawer for UserPage re-design

kylerwsm:
## Problem

Implemented the desktop view side drawer.

**Improvements**:

- For simplification, the drawer manages its own state. Redux store is used for data that are globally used, like user links.
- Converted some files to `.tsx`.
- Using a `MuiDrawer` rather than `MuiDialog` for the main UserPage modal. Kudos to @JasonChong96 for recommending that.

## Before & After Screenshots

**AFTER**:
image


src/client/components/UserPage/Drawer/ControlPanel/helpers/reducers.ts
```diff
@@ -0,0 +1,48 @@
...
+  const newState: Partial<State> = {}
+
+  switch (action.type) {
+    case DrawerActions.openControlPanel:
+      newState.controlPanelIsOpen = true
+      newState.relevantShortLink = action.payload
+      break
+    case DrawerActions.closeControlPanel:
+      newState.controlPanelIsOpen = false
+      newState.relevantShortLink = undefined
```
JasonChong96, line 34: Nitpicking but should this be null instead? Since conceptually null means that there's no relevant link

kylerwsm: Converted optional string to nullable string in commit de093b3

kylerwsm:
Force pushed to rebase and resolve conflicts

PR #104, opened 2020-05-20
Sorting and Filtering for Mobile and Desktop

JasonChong96:
## Problem

On the current `user-page-table` branch, there is no way for users to sort or filter their links.

## Solution

- Added filter and sorting for both desktop and mobile
- useAppMargins was not used for this component as it is rather sparse and works fine even on the skinniest screens. Reducing the margin from 32px would make it look weird on phones
- The backend can now handle requests that filter `state` and `isFile`
- All new files are `.tsx` except for the main file for the new panel. This is because it accesses the redux store and conversion should be deferred till the actions are converted to `ts`

## Before & After Screenshots

**AFTER**:
[Desktop View](https://user-images.githubusercontent.com/35889982/82471041-bce65500-9af8-11ea-9314-b62e4e59a27f.png)
[Longer Desktop View](https://user-images.githubusercontent.com/35889982/82471494-59105c00-9af9-11ea-8b7e-7527284efd22.png)
[iPhone 6/7/8 Plus view](https://user-images.githubusercontent.com/35889982/82471439-45fd8c00-9af9-11ea-98f9-ea4d9be079dc.png)


src/client/components/UserPage/UserLinkTable/ToolBar/FilterSortPanel/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+        >
+          <SortPanel onChoose={setOrderBy} currentlyChosen={orderBy} />
+          <FilterPanel {...filterConfig} />
+          <FilterSortPanelFooter
+            onApply={changeSortAndFilterHandler}
+            onReset={reset}
+          />
+        </Grid>
+      </Paper>
+    </Collapse>
```
kylerwsm, line 112: Can I check why are we not using the MuiMenu for this?

src/client/components/UserPage/UserLinkTable/ToolBar/FilterSortPanel/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+        orderBy: title,
+        sortDirection: direction,
+        filter: {
+          state,
+          isFile,
+        },
+      }),
+    )
+    dispatch(userActions.getUrlsForUser())
+  },
```
kylerwsm, line 25: Why do we need to re-fetch user links on filter?

src/server/api/user.ts
```diff
@@ -338,6 +338,8 @@ router.get('/url', validateUrlRetrieval, async (req, res) => {
     offset = 0,
     orderBy = 'updatedAt',
     sortDirection = 'desc',
+    isFile,
+    state,
```
kylerwsm, line 342: I am pretty sure filtering will not need to touch on the backend right?

src/client/components/UserPage/UserLinkTable/ToolBar/FilterSortPanel/index.jsx
kylerwsm: Do you think we also split styles for widgets within component directories to have their own dedicated styles file too?

kylerwsm: And rather than using box-icons, I believe we can download the assets and put them into our project directly.

JasonChong96: Mui menu doesn't allow us to keep the dimensions equal to the parent. 

We could do that but I noticed that quite a few styles are shared so I opted to share it instead.

JasonChong96: Should we keep to using box-icons since we can easily manipulate the colour without duplication the asset?

kylerwsm: For the newer components, I have been using local assets in place of box-icons. It is easily better to do it progressively

kylerwsm: > Mui menu doesn't allow us to keep the dimensions equal to the parent.
> 
> We could do that but I noticed that quite a few styles are shared so I opted to share it instead.

I am worried about the trade-off, since we are using an absolute offset for the menu now

JasonChong96: The offset is in relation to the top of the search input so its simple to manage. It's way more complex to ensure that the two widths are equal without it as menu does not provide this form of functionality, especially if the width changes when resizing the page

kylerwsm: I believe the width of the search input does not change with varying desktop view width. Even if it is so, we should be able to override the paper property in Menu to fix the menu width. And to better connect the components, we can also abstract the width of the search input into a hook.

PR #106, opened 2020-05-21
Introduce template for technical specs

liangyuanruo:
## Problem

Writing a technical spec increases the chances of having a successful project, service, or feature that all stakeholders involved are satisfied with. Not having it decreases the chances of something going horribly wrong during implementation and even after the product is launched.

## Solution

Adapt the technical specification layout from [Stack Overflow](https://stackoverflow.blog/2020/04/06/a-practical-guide-to-writing-technical-specs/).

### DIFF
--- a/.github/ISSUE_TEMPLATE/technical_spec.md
+++ b/.github/ISSUE_TEMPLATE/technical_spec.md
@@ -1,0 +1,230 @@
+# Technical Proposal
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


liangyuanruo:
> lgtm. the release checklist template is stored in the wiki. might be a good idea to standardise where we store such documents

It’s better if it is able to show up as an issue template IMO.

PR #107, opened 2020-05-21
Refactor front end redux store to use TypeScript

JasonChong96:
## Problem

The team has been trying to progressively migrate front-end components from using javascript to typescript. Typescript provides many modern compile-time checks that would be useful to find and fix possible bugs before run-time. Many components cannot be migrated meaningfully yet as the redux store which they access are still not typed.

Typescript is magic

## TODO:

- [x] Wait for user page table revamp to be merged into develop

## Solution

Refactor actions and reducers.

**Improvements**:

- Each file of action/reducers have been split into two files each, `index.ts` and `types.ts` and put into their own folders.

**New dependencies**:
- `history`: Converted from react's peer dependency to an explicit one for typescript recognition

**New dev dependencies**:

- `@types/moment-timezone` : Types for `moment-timezone`


src/client/actions/user/index.ts
```diff
@@ -266,21 +319,26 @@ const createUrlOrRedirect = (history) => (dispatch, getState) => {
       })
   }
 
+  // @ts-ignore
```
yong-jie, line 322: same question here

src/client/reducers/user/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  filter: {
+    isFile?: boolean
+    orderBy?: string
+    state?: UrlState
+  }
+}
+
+export type UrlState = 'ACTIVE' | 'INACTIVE'
+
+export type SortDirection = 'desc' | 'asc'
```
yong-jie, line 32: both `UrlState` and `SortDirection` (and also all other types following this pattern) are better represented in a typescript enum

src/client/actions/user/index.ts
```diff
@@ -184,7 +186,7 @@ const createUrl = (dispatch: AllThunkDispatch, user: UserState) => {
     }
     return response.json().then((json) => {
       if (response.status === 200) {
-        dispatch<ResetUserStateAction>(resetUserState())
+        dispatch(resetUserState())
```
yong-jie: what happens when the converse approach is done btw? like instead, we keep `dispatch<ResetUserStateActions>` here, but the argument of the function instead of `dispatch<AllActions>` we use a more specific etc

JasonChong96: Would it be too troublesome to have to change it if more actions are added? Since it is valid to dispatch any allowable action

yong-jie: this is an interesting point. 

my interpretation of typescript integration in `dispatch` would be that we restrict ourselves to the smallest subset of types needed and work with that. This way we know that our function was designed to work with those specific types only, and if the `dispatch` calls inside the function body uses extra types it would throw, telling us this was not the intended use of that function.

paging @liangyuanruo for any inputs in this discussion

liangyuanruo: wholly agree that we should narrow down the types and lean on the compiler for downstream type safety. dispatch doesn't know the type beforehand, so it's important to template the specific type in.

src/client/actions/user/index.ts
```diff
@@ -1,34 +1,64 @@
 import moment from 'moment-timezone'
 
+import querystring from 'querystring'
+/* eslint-disable-next-line import/no-extraneous-dependencies */
+import { History } from 'history'
```
liangyuanruo, line 5: how does this import get resolved?

src/client/actions/user/index.ts
```diff
@@ -83,47 +134,53 @@ const getUrlsForUser = () => async (dispatch, getState) => {
     isFile,
   }
 
-  dispatch(isFetchingUrls(true))
+  dispatch<IsFetchingUrlsAction>(isFetchingUrls(true))
   const { json, isOk } = await getUrls(queryObj)
 
   if (isOk) {
-    json.urls.forEach((url) => {
+    json.urls.forEach((url: any) => {
```
liangyuanruo, line 141: we should probably annotate the type here at some point.

src/client/actions/user/index.ts
JasonChong96: It's used so that typescript recognises history.push in this file

yong-jie: i was looking at MDN, the history api doesn't have the `push` method that we are using actually... this is mysterious https://developer.mozilla.org/en-US/docs/Web/API/History_API

yong-jie: maybe the type we want is actually derived from react-router? [link](https://reacttraining.com/react-router/web/api/history)

yong-jie: @liangyuanruo i've discussed with jason regarding this, we've added 'history' into package.json as an explicit dependency so we should be good in terms of module resolution

PR #108, opened 2020-05-21
User page table footer

JasonChong96:
## Problem

The footer of the user page table has not been updated to the latest design

## Solution

Updated user page table.
- The position of elements provided by material's footer does not match the design. This was solved by hiding some of the elements and replacing them with our own in ActionComponents.
- To center the page selector, I made the container width 50% and pushed the selector to the right with `margin-left: auto` and then shifted it to the right by half its width to centralize it on the screen.

## Before & After Screenshots

**AFTER**:
[Desktop](https://user-images.githubusercontent.com/35889982/82580634-f5059a80-9bc1-11ea-96b0-2c7e239938be.png)
Mobile

### DIFF
--- a/src/client/components/UserPage/UserLinkTable/UrlTable/MemoTablePagination/PaginationActionComponent/index.jsx
+++ b/src/client/components/UserPage/UserLinkTable/UrlTable/MemoTablePagination/PaginationActionComponent/index.jsx
@@ -3,15 +3,26 @@ import { Grid, IconButton, createStyles, makeStyles } from '@material-ui/core'
 import arrowLeftIcon from '../../../assets/arrow-left-icon.svg'
 import arrowRightIcon from '../../../assets/arrow-right-icon.svg'
 
-const useStyles = makeStyles((theme) =>
+const useStyles = makeStyles(() =>
   createStyles({
     pageSelectGrid: {
       fontWeight: 500,
       color: '#767676',
-      width: 'auto',
-      marginRight: '-139px',
-      [theme.breakpoints.up('sm')]: {
-        marginLeft: 'auto',
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      display: 'flex',
+      justifyContent: 'center',
+      width: '100%',
+      height: '100%',
+      zIndex: 1,
+    },
+    gridItemHorizontalPadding: {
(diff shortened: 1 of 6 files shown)


src/client/components/UserPage/UserLinkTable/UrlTable/MemoTablePagination/PaginationActionComponent/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+      alignItems="center"
+      spacing={5}
+      className={classes.pageSelectGrid}
+    >
+      <Grid item>
+        <IconButton
+          onClick={(event) => onChangePage(event, page - 1)}
+          disabled={page <= 0}
+        >
+          <img src={arrowLeftIcon} alt="Previous page" />
```
kylerwsm, line 35: Small thing here, but can we make all the icons used here to be not draggable, i.e. add `draggable={false}` within the img tags?

src/client/components/UserPage/UserLinkTable/UrlTable/MemoTablePagination/PaginationActionComponent/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+import arrowLeftIcon from '../../../assets/arrow-left-icon.svg'
+import arrowRightIcon from '../../../assets/arrow-right-icon.svg'
+
+const useStyles = makeStyles((theme) =>
+  createStyles({
+    pageSelectGrid: {
+      fontWeight: 500,
+      color: '#767676',
+      width: 'auto',
+      marginRight: '-139px',
```
kylerwsm, line 12: I believe as much as possible we should avoid the use of negative margins

kylerwsm:
@JasonChong96 I have created a pull request into this branch that attempts to center the footer without the use of negative margins

JasonChong96:
> @JasonChong96 I have created a pull request into this branch that attempts to center the footer without the use of negative margins

merged it in with this

kylerwsm:
Anyways, before I merge this, will the border thickness fix be part of this PR, or a separate one?

JasonChong96:
@kylerwsm seperate

PR #122, opened 2020-05-27
Server side QR code rendering

### LINKED ISSUE #56: Bundle QR code with link in same image
Today QR codes are in their own image. However this is a phishing risk if link is not displayed below the QR code so user has the option to check that QR code indeed directs to the link. It is also a usability issue if citizens prefer to type links than scan QR codes.

Include short link inside image of QR code so both are shared together. Short link can be below the QR code.

### LINKED ISSUE #57: Email QR code to self
Today after generating the QR code agency typically transfers it back to their government email to circulate. As Go.gov.sg is on the Internet, officers might have to use their personal emails to email themselves. No sensitivity concerns as links are public anyway, but is quite troublesome.

1. Shift QR code generator to the server
2. Hook into SES to email QR code to logged in user
3. "Email self" button on the QR code modal

### LINKED ISSUE #124: Problem with PNG downloads on IE11
**Describe the bug**
In the midst of our v2 redesign, our changes have caused PNG QR code downloads on IE11 to not work. Svg downloads still works as intended.

I re-tested in on other browsers, and PNG and SVG downloads are still working well on Chrome Firefox, and Safari.

Moving forward, I have added new tests to our manual release checklist to ensure that QR codes are also working well also on IE11.

**To Reproduce**
1. Go to go.gov.sg, and sign in with our gov.sg credentials.
2. Click on any created links or create one.
3. Try downloading the QR code as a PNG image.
4. Observe that nothing is happening.

**Expected behavior**
The QR code should be downloaded to the user's download folder.

 - OS: Windows 10
 - Browser: IE
 - Version: 11

kylerwsm:
## Problem

- To enable QR codes to be generated on the backend server side.
- Upgrade the QR code to that of v2 designs.
- Use SVG rather than PNG for Go logo for better resolution.
- Part of #57.
- Closes #56.
- Closes #124.

## Solution

I am using jsdom as an alternative dom for node, and d3 to draw our GoGovSg logo on the generated QR code generated from the qrcode library.

## TODO
- [x] Generate SVG QR code
- [x] Generate PNG QR code
- [x] Generate JPEG QR code
- [x] Write test for QR code
- [x] Add validation using Joi
- [x] Add short url into QR code
- [x] QR codes can only be generated if user is signed in
- [x] Only QR codes of existing links can be generated

## After
image

## Tests
- [x] Download QR code on modern browsers locally
- [x] Download QR code on IE11 locally
- [x] Download QR code on modern browsers on staging
- [x] Download QR code on IE11 on staging

## Deploy Notes

**New dependencies**:

- `d3` : To add our Go QR code to the generated QR code.
- `jsdom` : To render our QR code in a node environment.
- `sharp`: A convenient tool to convert our source svg QR code to png and jpeg formats.

**New dev dependencies**:

- `jsqr`: To read rendered QR code for testing.
- `upng-js`: To parse rendered .png QR codes for testing.


src/server/api/qrcode.ts
```diff
@@ -0,0 +1,24 @@
+import Express from 'express'
+
+import createGoQrCode from '../util/qrcode'
+import ImageFormat from '../../shared/util/imageFormat'
+
+const router = Express.Router()
+
+router.post('/', (req, res) => {
+  const url = req.body.url as string
+  const format = req.body.format as ImageFormat
```
yong-jie, line 10: What happens when `format` is a string that is outside of the possible enum values?

src/server/api/qrcode.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (!url || !format) {
+    res
+      .status(400)
+      .send('Please include the required "url" and "format" headers')
+    return
```
yong-jie, line 15: Perhaps some validators can be useful too since this endpoint requires a very specific type of data to be passed in. If you branch off the file-upload-serverside PR branch, you'll get access to Joi's validators which might make it easy

src/server/api/qrcode.ts
kylerwsm: Thanks for picking this up. This is now fixed with an object value check

src/server/api/qrcode.ts
kylerwsm: Added Joi validators, it should be working pretty fine now

src/client/components/UserPage/Drawer/ControlPanel/widgets/DownloadButton.tsx
```diff
@@ -6,12 +6,49 @@ import {
...
 } from '@material-ui/core'
+import FileSaver from 'file-saver'
 
 import TrailingButton from './TrailingButton'
 import downloadIcon from '../assets/download-icon.svg'
-import { useDrawerState, useDrawerDispatch } from '../..'
-import QRCodeModal from '../QRCodeModal'
-import DrawerActions from '../helpers/reducers'
+import { useDrawerState } from '../..'
+import ImageFormat from '../../../../../../shared/util/imageFormat'
```
liangyuanruo, line 14:
```suggestion
import ImageFormat from '../../../../../../shared/util/image-format'
```
nitpick filename, should not be camelCase

src/client/components/UserPage/Drawer/ControlPanel/widgets/DownloadButton.tsx
```diff
@@ -6,12 +6,49 @@ import {
...
+}
+
+// Downloads QR code from server.
+async function downloadServerQrCode(
+  shortLink: string,
+  format: ImageFormat,
+): Promise<void> {
+  if (typeof window === 'undefined') {
+    throw Error('This code should only run in a browser environment')
+  }
```
liangyuanruo, line 38: I don't think this check is necessary given that the code clearly lives in the client-side of the codebase

src/server/util/qrcode.ts
```diff
@@ -0,0 +1,144 @@
...
+      return qrSvgString
+    }
+    case ImageFormat.PNG: {
+      const buffer = Buffer.from(qrSvgString)
+      return sharp(buffer).resize(IMAGE_WIDTH, imageHeight).png().toBuffer()
+    }
+    case ImageFormat.JPEG: {
+      const buffer = Buffer.from(qrSvgString)
+      return sharp(buffer).resize(IMAGE_WIDTH, imageHeight).jpeg().toBuffer()
+    }
```
liangyuanruo, line 140: Nice, we have JPEG now?

src/server/api/qrcode.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+      return format
+    })
+    .required(),
+})
+
+const router = Express.Router()
+const validator = createValidator()
+
+router.post('/', validator.body(qrCodeRequestSchema), (req, res) => {
```
liangyuanruo, line 36: this should be a GET endpoint because the client is requesting for a resource from the server, and not creating a new resource to be stored on the server.

test/server/util/qrcode.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    test('png string', async () => {
+      const buffer = (await createGoQrCode(testUrl, ImageFormat.PNG)) as Buffer
+      const data = png.decode(buffer)
+      const out = {
+        data: new Uint8ClampedArray(png.toRGBA8(data)[0]),
+        height: data.height,
+        width: data.width,
+      }
+      const code = jsQR(out.data, out.width, out.height)
+      expect(code).toBe(code)
```
liangyuanruo, line 20: isn't this line of code a given?

src/server/api/qrcode.ts
```diff
@@ -0,0 +1,48 @@
...
+import Joi from '@hapi/joi'
+import { createValidator } from 'express-joi-validation'
+
+import { ogUrl } from '../config'
+import createGoQrCode from '../util/qrcode'
+import ImageFormat from '../../shared/util/imageFormat'
+import { isValidShortUrl } from '../../shared/util/validation'
+
+function isValidFormat(format: string): boolean {
+  return (Object as any).values(ImageFormat).includes(format)
```
yong-jie, line 11: sounds a bit hacky, was researching and saw [this link](https://stackoverflow.com/questions/42966362/how-to-use-object-values-with-typescript/42967397) which might be able to help circumvent this.

src/client/components/UserPage/Drawer/ControlPanel/widgets/DownloadButton.tsx
```diff
@@ -6,12 +6,49 @@ import {
...
+  const url = `https://${document.location.host}/${shortLink}`
```
yong-jie, line 39: Curious to know your reasoning behind letting the client decide the whole url rather than just specifying shortlink and having the server generate the rest

src/client/components/UserPage/Drawer/ControlPanel/widgets/DownloadButton.tsx
kylerwsm: That makes sense, I will remove it

src/server/util/qrcode.ts
kylerwsm: Yes!

src/client/components/UserPage/Drawer/ControlPanel/widgets/DownloadButton.tsx
kylerwsm: Will make the change!

test/server/util/qrcode.test.ts
kylerwsm: I missed out removing this, thanks for picking it up 😅

src/server/api/qrcode.ts
kylerwsm: That makes sense. I can make that change. Subsequently, I will also shift the query parameters from the body to the query string

src/server/api/qrcode.ts
kylerwsm: I have changed it to this, which I think may be cleaner.
```
function isValidFormat(format: string): boolean {
  const validFormats = Object.values(ImageFormat) as string[]
  return validFormats.includes(format)
}
```

src/client/components/UserPage/Drawer/ControlPanel/widgets/DownloadButton.tsx
kylerwsm: The server now returns a proposed filename in the headers (since body and content-type will be reserved for the image file), which the client will use when saving the image. If the filename header does not exist, the client fallbacks to the short url generated.

PR #123, opened 2020-05-27
Integration with gitpod

yong-jie:
This PR integrates our codebase with Gitpod, allowing users to start developing/testing code without the need to set up any infrastructure on their computer. This is also useful for open-source contributors who would like to briefly check out the codebase to develop and fix bugs, etc.

The integration with gitpod is done primarily through two files: 1) the gitpod config file which states the ports and processes to run, and also 2) the custom gitpod dockerfile that contains all the dependencies installed on it.


yong-jie:
Verified that I am able to carry out core functionality such as creating/editing links, ownership transfers, and even file uploads in gitpod.

PR #132, opened 2020-05-31
Minor code cleanup for Joi validation

liangyuanruo:
**Improvements**:

- Introduce convention of naming the types of API requests
- Compose types where possible to emphasise on code reuse
- Rename `transitionPage.ts` to kebab-case


src/types/server/api/login/handlers.d.ts
```diff
@@ -0,0 +1,9 @@
+export type EmailProperty = {
+  email: string
+}
+
+type OtpProperty = {
+  otp: string
+}
+
+export type VerifyOtpRequest = EmailProperty & OtpProperty
```
yong-jie, line 9: I've recently read about the use-case of `.d.ts` files  to provide typescript with type information for a [library written in javascript](https://stackoverflow.com/questions/21247278/about-d-ts-in-typescript). Based on that comment it didnt feel right to use them in our pure-typescript server-side codebase and have been discouraging Jason from using it. Should we stick with the `.ts` extension for this file?

liangyuanruo: Fair point, we can migrate them in another PR - will merge first to save the need for another approval.

PR #141, opened 2020-06-02
IE11 Banner

kylerwsm:
## Problem

Display to users via a banner that IE browsers may not work optimally on Go.gov.sg. This banner will only be shown to users on IE browsers.

## Solution

To check whether the browser is IE11, I am checking whether the browser supports `documentMode`, which is supported only in IE.

For consistency, the banner is implemented mobile-first even though there should not be any mobile phones running IE.

Reference to `documentMode`: https://www.w3schools.com/jsref/prop_doc_documentmode.asp

## Before & After Screenshots

**AFTER**:
image


src/client/components/BaseLayout/index.jsx
```diff
@@ -40,10 +43,12 @@ const BaseLayout = ({
   children,
 }) => {
   const classes = useStyles()
+  const isIE = useIsIE()
   return (
     <>
       <CssBaseline />
       <Masthead />
+      {isIE && <BannerForIE />}
```
liangyuanruo, line 51: We really only need to show this on the UserPage, can we move it there?

kylerwsm:
Updated with proposed banner text. Banner is now only shown on IE users when they are on UserPage.

image

src/client/components/BaseLayout/index.jsx
kylerwsm: I have added an additional check to ensure that the banner only shows in UserPage for IE users.

kylerwsm:
Resolved merge conflicts and cleared CI

PR #149, opened 2020-06-03
LocalStack

### LINKED ISSUE #117: S3 service for local development
Development today requires the use of an actual S3 bucket to be set up on AWS.

Add an S3 service such as _localstack_ to the local development environment.

An alternative service to consider is _Minio_.

kylerwsm:
## Problem

So, we need a S3-like service for local development. We want it to be accessed without credentials. And we want something that is contained only within our local development environment. Enter LocalStack, an easy-to-use test/mock framework for developing cloud applications.

Closes #117.

## Solution

- Introduce dockerised version of LocalStack in our docker compose
- Created an init script (heavily inspired from [formsg's](https://github.com/datagovsg/formsg/pull/2213) version)
- In development, `LocalStack` is injected in-lieu of `S3ServerSide`.
- Updated `.gitignore` to ignore local files created by LocalStack

## Notable changes
- Since LocalStack runs locally on http, https is disabled in development.
- Since file links on localhost is not a proper url, url validation on Sequelize is disabled in development.

## Notes
- This PR should be merged after `file-upload-frontend`
- This is a working prototype, and subsequently, I will be refining it further.

### DIFF
--- a/init-localstack.sh
+++ b/init-localstack.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -x
+pip install awscli-local
+awslocal s3 mb s3://$AWS_BUCKET_NAME
+awslocal s3api put-bucket-acl --bucket $AWS_BUCKET_NAME --acl public-read
+set
--- a/init-localstack.sh
+++ b/init-localstack.sh
@@ -2,5 +2,5 @@
 set -x
 pip install awscli-local
 awslocal s3 mb s3://$AWS_BUCKET_NAME
-awslocal s3api put-bucket-acl --bucket $AWS_BUCKET_NAME --acl public-read
+awslocal s3api put-bucket-acl --bucket $AWS_BUCKET_NAME --acl private
 set
--- a/src/server/util/localstack.ts
+++ b/src/server/util/localstack.ts
@@ -5,9 +5,9 @@ import AWS from 'aws-sdk'
 import { FileVisibility, S3Interface } from './aws'
 import { logger } from '../config'
 
-const BUCKET_NAME = process.env.AWS_S3_BUCKET || 'local-bucket'
-const BUCKET_ENDPOINT = 'http://localstack:4572'
-const ACCESS_ENPOINT = 'http://localhost:4572'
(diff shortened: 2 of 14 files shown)


init-localstack.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -x
+pip install awscli-local
+awslocal s3 mb s3://$AWS_BUCKET_NAME
+awslocal s3api put-bucket-acl --bucket $AWS_BUCKET_NAME --acl public-read
```
yong-jie, line 5: If we want to emulate what it's like production, we use private as the bucket policy, then individual objects override it via public-read when necessary

src/server/util/localstack.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable import/prefer-default-export */
+import { injectable } from 'inversify'
+import AWS from 'aws-sdk'
+
+import { FileVisibility, S3Interface } from './aws'
+import { logger } from '../config'
+
+const BUCKET_NAME = process.env.AWS_S3_BUCKET || 'local-bucket'
+const BUCKET_ENDPOINT = 'http://localstack:4572'
+const ACCESS_ENPOINT = 'http://localhost:4572'
```
yong-jie, line 10: Few points: 

1) Will there be any instances where bucket_endpoint deviates in value from access_endpoint? if not, they can be in the same variable maybe

2) Feels like some of these can go into our config.ts, such as the BUCKET_NAME. Actually, will there be issues if the bucket_name just uses what's defined in the S3_BUCKET env var (i.e. file-staging.go.gov.sg)?

init-localstack.sh
kylerwsm: Thanks for the pointer, I have made the change

src/server/util/localstack.ts
kylerwsm: 1. The bucket endpoint is different from the access endpoint. The bucket endpoint is the endpoint for our containerised app to communicate with our localstack.

2. Good point there. But I felt it will be good to differentiate the actual staging bucket name from localstack bucket name. Point taken for the part on config.ts, I will be shifting relevant localstack variables there.

PR #164, opened 2020-06-08
Improve URL validation and centralise client and server validators

kylerwsm:
## Problem

This PR is a follow-up to #149. Previously, the check for https was disabled in development, as localstack endpoints are not valid urls.

And as suggested by @liangyuanruo, I have centralised the common validators used by our client and server sides.

## Solution

- Shift from the built-in Sequelize `isUrl` check to our customised checks using the `validator` library. After which, I whitelist the endpoint used by localstack (http://localhost:4572).
- Added a new optional parameter to `isValidUrl` and `isHttps`, to indicate whether to take into account whitelisted hostnames when running the validation.
- Created tests for all url validators.

**Improvements**:

- Validation methods used by the client and server are now centralised at `src/shared/util/validation.ts`.
- Migrate the above-mentioned validation file to TypeScript.


src/shared/util/validation.ts
```diff
@@ -0,0 +1,60 @@
+import validator from 'validator'
+import { parse } from 'url'
+
+import blacklist from '../../server/resources/blacklist'
+
+export const WHITELIST = [new RegExp('^http://localhost:4572')]
```
liangyuanruo, line 6:
```suggestion
export const WHITELIST = [new RegExp('^http://localhost:4572$')]
```

src/shared/util/validation.ts
```diff
@@ -0,0 +1,60 @@
...
+// Tests if a URL string begins with https://.
+export function isHttps(url: string, useWhitelist = false): boolean {
+  if (useWhitelist && isWhitelisted(url)) return true
+  return /^https:\/\//.test(url)
+}
+
+// Check if a URL string is valid.
+export function isValidUrl(url: string, useWhitelist = false): boolean {
+  if (useWhitelist && isWhitelisted(url)) return true
+  return validator.isURL(url, URL_OPTS)
```
liangyuanruo, line 39: does this _guarantee_ that there is no change in behavior from the previous implementation?

src/server/models/url.ts
```diff
@@ -36,12 +40,14 @@ export const Url = <UrlTypeStatic>sequelize.define(
     longUrl: {
       type: Sequelize.TEXT, // Support >255 chars
       validate: {
-        // Disable long url checks for localstack in development.
-        ...(!DEV_ENV ? { isUrl: true } : {}),
+        urlCheck(url: string) {
+          if (!isValidUrl(url, true)) {
```
liangyuanruo, line 44: does this mean production will be able to save the whitelisted localhost URL?

src/shared/util/validation.ts
kylerwsm: The only change in this function is the addition of line 38, which will only be activated if `useWhitelist` is set as true. This change is backwards compatible as the new parameter is optional. And hence, there should not be a change of behaviour.

src/server/models/url.ts
kylerwsm: Technically, users will not be able to using as we are enforcing https during link creation and edits. But I get the gist here. I will convert the whitelist to a dev-only whitelist. Subsequently, if there is a need for a global whitelist in the future, it should be quite easy to add in.

kylerwsm:
I have updated my code such that whitelist is only used when `DEV_ENV` is true. I was considering the use of dependency injection, which I have decided to drop for now, as it somewhat decentralises the validators again (as the frontend does not have access to environment variables, and have no dependency injection).

Issue #156, opened 2020-06-04
Re-organise and refactor backend structure


liangyuanruo:
**Describe the problem**
The backend directory structure beginning to look incoherent and is showing signs of unbridled growth. It is currently a good time for refactoring given the recent expansion of file upload feature.

**Expected behavior**
Re-organise the files into separate folders . In increasing levels of hierarchy:

Models
- Model files define the objects that will be saved in the database e.g. `User`, `Url`

Repositories
- Repositories are singletons that provide an abstraction over the storage layer, such as retrieving a link (whether from cache or database), uploading files over S3 etc.

Services
- Services are singletons that provide an abstraction over business logic. Services rely on repositories and other services to process requests handed from the controllers. The application can also rely on Dependency Injection to obtain the correct service depending on the operating environment e.g. `S3LocalDev` vs `S3ServerSide`

Controllers / Middleware
- The Controller layer is responsible for handling requests/responses, such as returning the appropriate HTTP status code. They do not do any actual work, but delegate it to the service layers. They can also pass on the request to subsequent controllers as part of a middleware pattern.

Utilities
- A loose collection of functions and classes that help with miscellaneous tasks e.g. custom Error types

TODO:
- [x] Login API
- [x] Logout API
- [x] links API
- [x] qrcode API
- [x] Sentry API
- [x] Statistics API
- [x] User API
- [x] ga.ts util
- [x] qr code util

liangyuanruo:
"A Repository is essentially a facade for persistence that uses Collection style semantics (Add, Update, Remove) to supply access to data/objects. It is a way of decoupling the way you store data/objects from the rest of the application.

A service supplies coordination or other "services" that are required to operate your application. They are very different in that Services don't typically know how to access data from persistence, and repositories typically only access data/objects for any services you may have."

PR #165, opened 2020-06-08
Introduce UrlRepository abstraction layer

JasonChong96:
## Problem

This process will start with refactoring and introducing an abstraction barrier between the data store and other parts of the code.

## Solution

Refactor API endpoints to use UrlRepository when accessing data store.

**Improvements**:

- All endpoints that edit URLs now access the same UrlRepository method, which can handle all cases. e.g File vs Not file, Changing state, Changing ownership
- Removed now obselete old UrlCache and old UrlRepository

**Alternatives Considered**:

I have considered sharing more methods in the BaseRepository interface but found that we have few repositories and they have unique requirements. e.g UrlRepository needs to handle files/cache/db, Otp needs cache only, User needs db only.
For the same reason, I also decided against having a BaseCache/SequelizeRepository and it is difficult to share code and method signatures.

As transactions in URL involve both S3 and Sequelize, it is difficult to abstract them into different files while keeping code readable, hence the relatively large size of UrlRepository.

## Tests

I have run through the following on staging:
1. Creating URL works
2. Transferring ownership of URL works
3. Editing of URL works
4. Redirect works (The automated tests should cover this but just to be sure)
5. Toggling state to active/inactive works and is reflected accordingly
6. Repeat all previous steps for files


src/server/repositories/UrlRepository.ts
```diff
@@ -0,0 +1,240 @@
...
+export class UrlRepository extends AbstractUrlRepository {
+  private fileBucket: S3Interface
+
+  public constructor(@inject(DependencyIds.s3) fileBucket: S3Interface) {
+    super()
+    this.fileBucket = fileBucket
+  }
+
+  findOne: (where: object) => Promise<StorableUrl | null> = async (where) => {
+    return (await Url.findOne({ where: { ...where } })) as StorableUrl | null
```
liangyuanruo, line 75:
```suggestion
    return (await Url.findOne({ where })) as StorableUrl | null
```

src/server/repositories/UrlRepository.ts
```diff
@@ -0,0 +1,240 @@
...
+import { FileVisibility, S3Interface } from '../util/aws'
+import { ACTIVE } from '../models/types'
+
+const { Public, Private } = FileVisibility
+
+/**
+ * A url repository that handles access to the data store of Urls.
+ */
+@injectable()
+export abstract class AbstractUrlRepository
```
liangyuanruo: may i ask what drove your thinking to use an abstract class instead of an interface?

JasonChong96: I just realised this and committed the change. Was initially experimenting with sharing methods between repositories but abandoned it for the reason given in the pr

src/server/repositories/UrlRepository.ts
JasonChong96: Typescript complains that it does not match Sequelize's types when it isn't spread for some reason.

JasonChong96: Found a way to typecast it to the sequelize type

src/server/repositories/interfaces/UrlRepositoryInterface.ts
```diff
@@ -0,0 +1,48 @@
+import { BaseRepository } from './BaseRepository'
+
+/**
+ * A url repository that handles access to the data store of Urls.
+ */
+export interface UrlRepositoryInterface extends BaseRepository<StorableUrl> {
+  findOne(where: object): Promise<StorableUrl | null>
```
liangyuanruo, line 7: can we change the implementation to `findByShortUrl` that accepts a single string? since that is pretty much the only valid scenario in which you can `findOne` reliably.

JasonChong96: Then should we get rid of BaseRepository? Since its going to be different for each Repo

src/server/repositories/UrlRepository.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+@injectable()
+export class UrlRepository implements UrlRepositoryInterface {
+  private fileBucket: S3Interface
+
+  public constructor(@inject(DependencyIds.s3) fileBucket: S3Interface) {
+    this.fileBucket = fileBucket
+  }
+
+  findOne: (where: object) => Promise<StorableUrl | null> = async (where) => {
```
liangyuanruo, line 26: can we add the `public` keyword for consistency and clarity?

src/server/repositories/types.ts
```diff
@@ -0,0 +1,27 @@
...
+ * A type that represents Urls stored in the data store.
+ */
+type StorableUrl = {
+  shortUrl: string
+  longUrl: string
+  state: StorableUrlState
+  clicks: number
+  isFile: boolean
+}
+
```
liangyuanruo, line 11: can we take a look at combining this with the existing `UrlType` interface specified in `src/server/models/url.ts`? it looks like we are violating the DRY principle here.

JasonChong96: UrlType exposes Sequelize methods and making the interface use StorableUrl's fields does not work either since state in `UrlType` is a `Sequelize.EnumDataType<string>`

liangyuanruo: i see. my primary motivation is to avoid the constant typecasting to _unknown_ and then again to _StorableUrl_.

JasonChong96: Seems like the issue was that the original UrlType had the wrong typings, i.e the runtime type was actually just a number and string but they were labelled differently. Fixing this has allowed converting to the dto type without any typecasting.

liangyuanruo: brilliant

PR #169, opened 2020-06-09
Introduce UserRepository abstraction layer and Introduce Mappers

JasonChong96:
## Problem

This PR continues the process of addressing this by creating a UrlRepository which handles the data store of users.

With the introduction of UserRepository, multiple classes need to share usage of mapper functions which map persistence (sequelize) objects to data transfer objects.

## Solution

Introduce UserRepository to further abstract the data store from other services.

**Improvements**:

- Mapper classes and interfaces have been introduced to map between persistence and dto's.
- Code coverage increased somehow

## Tests

- [x] Creating URL works
- [x] Transferring ownership of URL works
- [x] Editing of URL works
- [x] Redirect works 
- [x] Toggling state to active/inactive works and is reflected accordingly
- [x] Repeat all previous steps for files
- [x] Logging in works
- [x] Verifying OTP works

**Additional notes**
There are some consecutive lines such as these 
`persistenceToDto(urlType: UrlType): StorableUrl`
`persistenceToDto(urlType: UrlType | null): StorableUrl | null {`
in the mappers. These are there to tell the typescript compiler to that the method returns a nullable value if and only if the input is nullable.


src/server/api/login/handlers.ts
```diff
@@ -73,7 +73,7 @@ export async function verifyOtp(req: Express.Request, res: Express.Response) {
   const { getOtpForEmail, setOtpForEmail, deleteOtpByEmail } = container.get<
     OtpCache
   >(DependencyIds.otpCache)
-  const { findOrCreateWithEmail } = container.get<UserRepository>(
+  const { findOrCreateWithEmail } = container.get<UserRepositoryInterface>(
```
liangyuanruo, line 76: can we move all `container.get` calls to the top-level imports?

JasonChong96: They are currently in the function to allow test suites to inject dependencies before every function call. If it is injected outside the function, then they are bound during module import and can't be changed.

It can be solved once we refactor controllers/services to be objects and have them injected on instantiation.

liangyuanruo: sounds good

PR #179, opened 2020-06-10
Unify URL edit patch endpoints

### LINKED ISSUE #121: Unify PATCH endpoints for link edit operations
At present, two endpoints exist for editing links in GoGovSG. The first allows one to modify the longUrl, and another to toggle the `ACTIVE` or `INACTIVE` state of the link.

Perhaps a better approach would be to unify these two endpoints into a single `PATCH` endpoint. To counter the potential complication in logic, we could make use of dependency injection provided by `inversify` to route different operations to different 'resolvers' in the same endpoint handler.

yong-jie:
## Problem

At present, two endpoints exist for editing links in GoGovSG. The first allows one to modify the longUrl, and another to toggle the `ACTIVE` or `INACTIVE` state of the link. This can be achieved more efficiently with just one PATCH endpoint.

Closes #121 

## Solution

- [x] Unify the validators for both state-edit and longUrl-edit endpoints
- [x] Merge logic of the longUrl-edit endpoint into the state-edit one. The reason is because we want to take the `/url` endpoint rather than the `/url/edit` one.
- [x] Modify frontend API call code to make longUrl/file edit requests to the /url endpoint rather than /url/edit

## Backwards compatibility

To cater to the possibility of old client code being run by some browsers, currently opting to keep the old /url/edit endpoint there for this release, and then remove it in the subsequent one.

### DIFF
--- a/src/server/api/user/index.ts
+++ b/src/server/api/user/index.ts
@@ -20,9 +20,9 @@ import { UserRepositoryInterface } from '../../repositories/interfaces/UserRepos
 import { NotFoundError } from '../../util/error'
 import {
   ownershipTransferSchema,
-  stateEditSchema,
   urlRetrievalSchema,
   urlSchema,
+  urlEditSchema,
 } from './validators'
 
 const router = Express.Router()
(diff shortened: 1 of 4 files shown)


src/server/api/user/index.ts
```diff
@@ -215,7 +214,6 @@ router.patch(
           }
         : undefined
 
-      // TODO: Remove force cast once UserRepository has been made
```
yong-jie, line 258: @JasonChong96 wanted to confirm with you whether the comments deleted in this commit is okay

JasonChong96: They can be removed, I forgot to remove them in the previous pr

PR #185, opened 2020-06-12
Refactor StatisticsController and its dependencies

JasonChong96:
## Problem

This PR continues the process of addressing this by refactoring StatisticsController and its dependencies to conform to the Controller Service Repository pattern

## Solution

- Refactor StatisticsController/Service/Repository
- Statistics makes use of its own repository as statistics for analytics might come from other tables in the future.

### DIFF
--- a/src/server/services/StatisticsService.ts
+++ b/src/server/services/StatisticsService.ts
@@ -1,0 +1,23 @@
+import { inject, injectable } from 'inversify'
+import { StatisticsServiceInterface } from './interfaces/StatisticsServiceInterface'
+import { DependencyIds } from '../constants'
+import { StatisticsRepositoryInterface } from '../repositories/interfaces/StatisticsRepositoryInterface'
+import { GlobalStatistics } from '../repositories/types'
+
+@injectable()
+export class StatisticsService implements StatisticsServiceInterface {
+  private statisticsRepository: StatisticsRepositoryInterface
+
+  public constructor(
+    @inject(DependencyIds.statisticsRepository)
+    statisticsRepository: StatisticsRepositoryInterface,
+  ) {
+    this.statisticsRepository = statisticsRepository
+  }
+
+  getGlobalStatistics: () => Promise<GlobalStatistics> = async () => {
+    return this.statisticsRepository.getGlobalStatistics()
+  }
+}
+
+export default StatisticsService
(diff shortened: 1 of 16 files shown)


src/server/services/StatisticsService.ts
```diff
@@ -0,0 +1,27 @@
...
+    @inject(DependencyIds.statisticsRepository)
+    statisticsRepository: IStatisticsRepository,
+  ) {
+    this.statisticsRepository = statisticsRepository
+  }
+
+  getGlobalStatistics: () => Promise<GlobalStatistics> = async () => {
+    const clickCount = await this.statisticsRepository.getGlobalClickCount()
+    const linkCount = await this.statisticsRepository.getGlobalLinkCount()
+    const userCount = await this.statisticsRepository.getGlobalUserCount()
```
liangyuanruo, line 21: Previously we could retrieve all the values at once from Redis via `mget`. Can we preserve that behavior so that we don't triple the request load to Redis?

LoneRifle:
I would dispense with .NET naming conventions when it comes to interfaces, ie don't prefix with `I`

test/server/mocks/repositories/StatisticsRepository.ts
```diff
@@ -2,19 +2,12 @@
 
 import { injectable } from 'inversify'
 import { StatisticsRepositoryInterface } from '../../../../src/server/repositories/interfaces/StatisticsRepositoryInterface'
+import { GlobalStatistics } from '../../../../src/server/repositories/types'
 
 @injectable()
 export class MockStatisticsRepository implements StatisticsRepositoryInterface {
```
LoneRifle, line 8: Surely you could use `jest.mock` for this..?

JasonChong96: As in to mock to repository?

liangyuanruo: @LoneRifle i think this is fine though. one significant benefit of using DI is to easily swap out implementation when necessary, including tests.

LoneRifle: Not disputing that - I'm saying that instead of creating a specific class for testing, you just supply one dynamically created using mocks

PR #187, opened 2020-06-15
chore(husky): enhance with lint-staged

### LINKED ISSUE #183: Apply lint to pre-commit hooks
While it is good that our code push now fails on lint errors, it is still possible to commit and push badly formatted code (formatting fixes need not be committed before the code can be pushed). While badly formatted code will be caught by our CI, it would be better if we have in-development measures as well.

A pre-commit hook with lint-staged may be the ideal solution for this problem. This ensures that all of our commits are properly formatted and structured, and makes dedicated commits to fix stylings redundant. However, an obvious drawback will be additional delays when commit code changes.

I am open to hear alternative solutions!

LoneRifle:
## Problem
Files can currently be committed without linting, which presents inconvenience to the developer at pre-push or CI since the developer has to add or modify a commit to incorporate the linting. lint-staged is designed to avoid precisely this, `git add`ing any modifications to files (including linting) before commit.

Closes #183

## Solution

Allow husky to invoke lint-staged so that files that are about to
be committed get automatically linted.

- Update husky, add lint-staged
- Add lint-staged commands to lint-fix files before they are committed

**New dev dependencies**:

- `lint-staged` : as detailed above

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,12 @@
     "winston": "^3.2.1"
   },
   "devDependencies": {
+    "@babel/core": "^7.4.3",
+    "@babel/plugin-proposal-class-properties": "^7.4.0",
+    "@babel/preset-env": "^7.4.3",
+    "@babel/preset-react": "^7.0.0",
+    "@babel/preset-typescript": "^7.9.0",
+    "@sentry/webpack-plugin": "^1.10.0",
     "@types/bcrypt": "^3.0.0",
     "@types/bluebird": "^3.5.26",
     "@types/classnames": "^2.2.10",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -158,7 +159,15 @@
   "husky": {
     "hooks": {
       "pre-push": "npm run lint",
-      "pre-commit": "sh secrets-check.sh"
+      "pre-commit": "sh secrets-check.sh && lint-staged"
     }
+  },
+  "lint-staged": {
+    "**/*.(js|jsx)": [
+      "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.json --fix"
```
karrui, line 167: Don't actually need to pass in .js or .jsx anymore, lint staged automatically does it. Same for the other globs

PR #189, opened 2020-06-15
build(lint): add commitlint, commitizen

LoneRifle:
## Problem
Going forward, commits to this repository should follow conventional commits

## Solution
Enforce commit linting on push, adhering to conventional commits
with exceptions as defined in the commitlint config file

**New scripts**:

- `cz` : invokes commitizen via `git-cz`

**New dev dependencies**:

- `@commitlint/{cli, config-conventional, travis-cli}` : lints commits on push and on Travis
- `commitizen`, `cz-conventional-changelog`: provides a human-friendly interface to compose commit messages via `npm run cz`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ script:
   - npm run lint
   - npm run build
   - npm run test:ci
+  - npx commitlint-travis
 
 before_deploy:
   # Workaround to run before_deploy only once
--- a/package.json
+++ b/package.json
@@ -18,10 +18,15 @@
     "docker-dev": "concurrently \"npm run server-dev\" \"npm run client-dev\"",
     "dev": "docker-compose -f docker-compose.yml up --build",
     "test": "jest --collectCoverage",
-    "test:ci": "jest --coverage && coveralls < coverage/lcov.info"
+    "test:ci": "jest --coverage && coveralls < coverage/lcov.info",
+    "cz": "git-cz"
   },
   "author": "Yuanruo Liang",
-  "license": "",
+  "license": "MIT",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -158,7 +168,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "npm run lint",
+      "pre-push": "commitlint --from origin/develop --to HEAD --verbose",
```
liangyuanruo, line 171: has code linting been removed?

.travis.yml
```diff
@@ -17,6 +17,7 @@ script:
   - npm run lint
   - npm run build
   - npm run test:ci
+  - npx commitlint-travis
```
liangyuanruo, line 20: this line only lints the new commit messages?

package.json
LoneRifle: Code linting has been moved to pre-commit in an earlier PR. Linting errors should be caught then and during CI, so that we don't have to wait for the pre-push hook to lint the entire codebase before pushing

.travis.yml
LoneRifle: Correct - as intended

package.json
karrui: Sorry to be nosy, but propose using `commit-msg` hook instead of `pre-push`.

Also suggest using `HUSKY_GIT_PARAMS`, which automatically retrieves the git params that invoked the husky hook.
```
"husky": {
  "hooks": {
    "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
  }
},
```

PR #192, opened 2020-06-15
feat: google analytics for drop-off statistics

### LINKED ISSUE #95: Add Google Analytics to measure transition page drop off
We launched transition page, and there have been no agency or citizen complaints, but we are not quantitatively certain transition page is not locking users out of their long links.

Theoretically can track a) each page's view, and b) emit Proceed button click (by passing in short link too so can map back to page view). If users are not locked out, b) should not be that much lower than a).

This is an important P2. Good to do this sooner rather than later. We just launched transition page and want to make sure users are not locked out.

kylerwsm:
## Problem

Make use of Google Analytics to measure drop-offs that might occur with the introduction of the transition page.

Closes #95.

## Solution

I have made use of event logging in Google Analytics.

I named this event which is focused on TransitionPage as `transition-page`. There are three possible event actions in this Event:
1. `served` - Logged by the server when serving a transition page.
2. `loaded` - Logged by the user after rendering the transition page.
3. `proceeded` - Logged by the user when the proceed button is clicked.

GA logging on the TransitionPage is done using ~~`analytics.js`~~ `gtag.js`.

In this configuration, realtime statistics can be view under Reports > Realtime > Events.

Tested on:
- [x] Local development
- [x] Staging, on links
- [x] Staging, on files

**Improvements**:

- The googleAnalytics directory is refactored into Services.


src/server/views/transition-page.ejs
```diff
@@ -41,8 +41,23 @@
...
+
+        ga('create', '<%- gaTrackingId %>', 'auto');
+
+        // Logs transition-page loaded event to GA.
+        ga('send', 'event', '<%- gaEventType %>', '<%- gaOnLoad %>');
+
         function proceedToDestination() {
-            window.location = "<%- escapedLongUrl %>" // Use double-quotes to avoid templating issue as per RFC3986 (https://tools.ietf.org/html/rfc3986#appendix-C)
+            // Logs transition-page proceeded event to GA.
+            ga('send', 'event', '<%- gaEventType %>', '<%- gaOnProceed %>');
```
liangyuanruo, line 58: I think we should be using the latest `gtag.js` instead - the code itself is [quite similar](https://developers.google.com/analytics/devguides/collection/gtagjs/migration).

JasonChong96:
Would having the client browser send the event to GA directly cause it to be blocked by ad blockers?

kylerwsm:
> Would having the client browser send the event to GA directly cause it to be blocked by ad blockers?

That is a good question. Through my manual tests, sending events to GA with my adblocker turned on is still working as intended. If you have tried it, is it also working as intended for you?

To slightly alleviate things, we do have an intermediate `loaded` event, which might give us insights as to whether the transition page is loading properly, or if events are not sending properly from the users side.

The alternative I can think of right now will be for the transition page to ping our server, so that our server can send the analytics instead. But they could double calls to our servers on every redirects.

@liangyuanruo @yong-jie do you have any inputs on this?

yong-jie:
Although the approach of routing our tracking requests through our server before sending to GA could solve this problem, I feel like it might be a bit too much development effort for the benefits it provides. Using this bypass might also be deemed unethical(?).

If we're concerned that firefox users could skew our data, would ignoring redirect events coming from firefox help?

kylerwsm:
I agree on the unethical part. I believe we can keep my current implementation, to observe the proportion of users blocking GA first, from the difference in `served` and `loaded` events. We might only need a solution if the proportion of untracked users is huge. What do you think?

PR #196, opened 2020-06-17
chore: enable debugging

LoneRifle:
Debugging in Gitpod

## Problem
To allow developers to quickly troubleshoot the codebase, configure
GoGovSG to run with debugging enabled, even when in a docker container.

## Solution
[npm, docker]
- replace nodemon and ts-node with ts-node-dev
- expose port 9229 (Node.js debug) for app container in docker-compose
- `npm run server-dev` listen for port 9229 on all interfaces, not just
  localhost, to facilitate debugging within a Docker container [1]

[docs]
- rewrite the README to document ts-node-dev instead of nodemon

[gitpod]
- ignore port 9229
- add Theia launch config to attach a debugger to `npm run docker-dev`

[misc]
- remove nodemon.json, no longer used
- allow favicons to be served directly by webpack-dev-server

**Notes**
[1] - This is needed as Docker would treat its host as a separate
machine. This also means that debugging on bare-metal would allow
others to connect to the port remotely. Caution is hence advised when
choosing to debug GoGovSG on bare-metal


LoneRifle:
@liangyuanruo it's basically the VSCode [fork](https://theia-ide.org/) maintained by the Eclipse Foundation that Gitpod builds on top of

Issue #181, opened 2020-06-11
Link Search


liangyuanruo:
# Technical Proposal

### Overview

Members of the public (MOPs) view go.gov.sg as a central hub for accessing government resources. By providing a link search feature, we will be able to better direct MOPs to access the resources that they require.

### Goals
- Admins must be able to _optionally_ provide search description and keywords for each link.
- Users must be able to search _performantly_ based on the description provided.
- Search functionality should be powered by [Postgres Full Text Search](https://www.postgresql.org/docs/9.5/textsearch-intro.html)
- Search results should be ranked by a relevance algorithm that balances
  - [Textual rank](https://www.postgresql.org/docs/9.5/textsearch-controls.html) as determined by Postgres
  - Total number of link clicks
  - Recency

### Non-Goals or Out of Scope

Due to the limited amount of text expected, we will not be considering a separate search database such as ElasticSearch.

We will not support searching by agency or user email.

### Design

- Extend the current `Url` and `UrlHistory` table to accept a new `description` and `keywords` column
- The column should be indexed by a GIN index configured to _english_. There may be a need to restrict saved text to english / ASCII
- Expose a public `/search` API for the frontend

### Test Plan

Explanations of how the tests will make sure user requirements are met
Unit tests
Integrations tests
QA

### Monitoring and Alerting Plan

To monitor alarms for expected load increase on the database (CPU, IOPS) after rollout.

#### Maintenance
Implementation should rely on GIN indexing to maintain the preprocessed _ts_vector_ instead of manually introducing database triggers.

#### Security
- Implementation should protect against SQL injection given that API is public and raw queries are expected to be used for this feature

### Future Work

- Fine-tuning search relevance algorithm
- Logging user search behavior

### References

https://www.postgresql.org/docs/9.5/textsearch-controls.html

LoneRifle:
Given search is now in soft-launch, I will have this issue closed. Follow-up work to be tracked through other issues.

PR #207, opened 2020-06-18
feat: data collection for search

JasonChong96:
## Problem

Before link search can be fully implemented, further collection of data from link owners is required.

#181

## Solution

This PR implements phase 0 of the implementation for link search. UI Fields and database columns are added for contact information and description for each link.

**Improvements**:

- Updated URL drawer margins to fit the design more
- Use of negative margins in `DrawerTextField` have been replaced by having helper text in absolute position. The negative margins made it hard to debug element positioning in browser debuggers and calculations of margins needed for subsequent elements were also confusing.

## Before & After Screenshots

**AFTER**:
[Desktop](https://user-images.githubusercontent.com/35889982/85267034-46e25e80-b4a7-11ea-8913-44cf13ae12ee.png)
[Mobile](https://user-images.githubusercontent.com/35889982/85267084-59f52e80-b4a7-11ea-9c39-aac3ceea1e76.png)

## Deploy Notes

The new migration script should be run before deployment. The new columns will be nullable and compatible with both the new and old code.

## TODO:
- [x] Create sql migration file
- [x] Update url retrieval endpoint to return new fields
- [x] Add new fields in edit url UI (Desktop)
- [x] Add new fields in edit url UI (Mobile)
- [x] Update call to url edit API
- [x] Update url edit endpoint
- [x] Add validation for contact email

### DIFF
--- a/src/client/actions/user/index.ts
+++ b/src/client/actions/user/index.ts
@@ -319,7 +319,7 @@ const updateLongUrl = (shortUrl: string, longUrl: string) => (
 }
 
 // API call to update description and contact email
-const applyLinkInformation = (shortUrl: string) => (
+const updateUrlInformation = (shortUrl: string) => (
   dispatch: ThunkDispatch<
     GoGovReduxState,
     void,
@@ -624,5 +624,5 @@ export default {
   replaceFile,
   setEditedContactEmail,
   setEditedDescription,
-  applyLinkInformation,
+  updateUrlInformation,
 }
--- a/src/client/actions/user/index.ts
+++ b/src/client/actions/user/index.ts
@@ -318,6 +318,45 @@ const updateLongUrl = (shortUrl: string, longUrl: string) => (
   })
 }
 
+// API call to update description and contact email
+const applyLinkInformation = (shortUrl: string) => (
+  dispatch: ThunkDispatch<
+    GoGovReduxState,
+    void,
+    SetErrorMessageAction | SetSuccessMessageAction
+  >,
+  getState: GetReduxState,
+) => {
+  const { user } = getState()
+  const url = user.urls.filter(
+    (urlToCheck) => urlToCheck.shortUrl === shortUrl,
+  )[0]
+  if (!url) {
+    dispatch<SetErrorMessageAction>(
+      rootActions.setErrorMessage('Url not found.'),
+    )
+    return null
+  }
+  return patch('/api/user/url', {
+    contactEmail: url.editedContactEmail,
+    description: url.editedDescription,
+    shortUrl,
+  }).then((response) => {
+    if (response.ok) {
+      dispatch<void>(getUrlsForUser())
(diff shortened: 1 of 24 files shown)


src/client/actions/user/index.ts
```diff
@@ -232,6 +258,10 @@ const getUrlsForUser = (): ThunkAction<
         .format('D MMM YYYY')
       // eslint-disable-next-line no-param-reassign
       url.editedLongUrl = removeHttpsProtocol(url.longUrl)
+      // eslint-disable-next-line no-param-reassign
```
liangyuanruo, line 261: might want to disable the lint for the entire file?

JasonChong96: I've changed it to disable this particular code block rather than having 4 disable-next-line's

src/server/models/url.ts
```diff
@@ -86,6 +88,19 @@ export const Url = <UrlTypeStatic>sequelize.define(
...
       allowNull: false,
     },
+    contactEmail: {
+      type: Sequelize.TEXT,
+      allowNull: false,
+      validate: {
+        isEmail: true,
+        isLowercase: true,
+        is: emailValidator.makeRe(),
+      },
```
liangyuanruo, line 98: isn't there an inconsistency, if the default value populated by the SQL migration is an empty string, but here we are insisting on `allowNull: false` and `isEmail: true`?

would it be easier to simply allow null values?

src/client/actions/user/index.ts
```diff
@@ -288,6 +315,45 @@ const updateLongUrl = (shortUrl: string, longUrl: string) => (
...
+    GoGovReduxState,
+    void,
+    SetErrorMessageAction | SetSuccessMessageAction
+  >,
+  getState: GetReduxState,
+) => {
+  const { user } = getState()
+  const url = user.urls.filter(
+    (urlToCheck) => urlToCheck.shortUrl === shortUrl,
+  )[0]
```
liangyuanruo, line 330: nitpick - prefer array destructuring over indexing
```suggestion
  const [ url ] = user.urls.filter(
    (urlToCheck) => urlToCheck.shortUrl === shortUrl,
  )
```

src/server/controllers/UserController.ts
```diff
@@ -60,7 +60,14 @@ export class UserController implements UserControllerInterface {
...
-    const { userId, longUrl, shortUrl, state }: UrlEditRequest = req.body
+    const {
+      userId,
+      longUrl,
+      shortUrl,
+      state,
+      description,
+      contactEmail,
+    }: UrlEditRequest = req.body
     const file = req.files?.file
```
liangyuanruo, line 71: can we refactor this function to introduce utility functions in the url management service for each of these items that may potentially be updated? i see a lot of inline logic in the controller.

JasonChong96: Do you mean moving the preprocessing to the service layer?

PR #209, opened 2020-06-19
feat: data collection for link click statistics

### LINKED ISSUE #168: [Phase 0] Analytics - Collection of data prior to the launch of analytics
_Related issue #54 and #55_ 
Before we roll out link click history and analytics, we should start collecting link click data.
Upon launch, users will be able to immediately see their link click history (up to 1 month before)

Start storing link clicks data in [insert location] 

**Additional context**
Go's Analytics will be done in multiple phases:
**Phase 0:** Start collecting data so that there is data upon launch
**Phase 1:** Link clicks for the past day, week and month
**Phase 2:** Link clicks history beyond 1 month

kylerwsm:
## Problem

This PR aims to start-off data collection for our upcoming link click statistics feature.

Closes #168.

## Solution

We do this by collecting and storing data directly to dedicated tables, each representing a statistic we will show to users. Additionally, in this implementation, badly formatted data will not be written to the statistics databases.


src/server/models/statistics/clicks.ts
```diff
@@ -0,0 +1,34 @@
...
+    type: Sequelize.STRING,
+    primaryKey: true,
+    validate: {
+      is: /^[a-z0-9-]+$/,
+    },
+  },
+  epochHours: {
+    type: Sequelize.INTEGER,
+    primaryKey: true,
+  },
```
liangyuanruo, line 29: isn't the time series chart shown by the day?

src/server/models/statistics/heatmap.ts
```diff
@@ -0,0 +1,45 @@
...
+    type: Sequelize.ENUM,
+    values: [
+      StorageDay.Monday,
+      StorageDay.Tuesday,
+      StorageDay.Wednesday,
+      StorageDay.Thursday,
+      StorageDay.Friday,
+      StorageDay.Saturday,
+      StorageDay.Sunday,
+    ],
```
liangyuanruo, line 37: i think using 0-6 would be more efficient for the database, and for lookup. can take reference from MomentJS.

also i believe the heat map stores the hours of the day as well, from 0-23?

kylerwsm: Thanks, I have took your recommendation!

src/server/models/statistics/clicks.ts
kylerwsm: I have converted the table to use date instead of hours

yong-jie:
have confirmed with @kylerwsm that data collection is presently working when tested on staging

src/server/controllers/LinkStatisticsController.ts
```diff
@@ -0,0 +1,37 @@
...
+  ) => Promise<void> = async (req, res) => {
+    const shortUrl = req.query.url
+    if (!shortUrl) res.json(null)
+    const user = req.session?.user as UserType
+    if (!user) res.json(null)
+    const linkStats = await this.linkStatisticsService.getLinkStatistics(
+      user.id,
+      shortUrl,
+    )
+    res.json(linkStats)
```
liangyuanruo, line 33: aren't you missing return statements after each call to `res.json`? the server will continue processing even if a response has been returned to the client

liangyuanruo: also get in the habit of using HTTP status codes to indicate the server response e.g. (`res.status(404).send()`), instead of `res.json` which returns an implicit HTTP 200 OK. the frontend should check the HTTP status code and react accordingly.

src/server/models/statistics/daily.ts
```diff
@@ -0,0 +1,45 @@
...
+      type: Sequelize.INTEGER,
+      defaultValue: 0,
+    },
+  },
+  {
+    indexes: [
+      {
+        unique: true,
+        fields: ['shortUrl', 'date'],
+      },
```
liangyuanruo, line 42: could you share the underlying `CREATE TABLE` and `CREATE INDEX` `SQL` statements for this table? it may not be necessary to specify the index, since a composite primary key is already defined.

src/server/models/statistics/devices.ts
```diff
@@ -0,0 +1,55 @@
...
+      defaultValue: 0,
+    },
+  },
+  {
+    indexes: [
+      {
+        unique: true,
+        fields: ['shortUrl'],
+      },
+    ],
```
liangyuanruo, line 53: similarly here

src/server/models/statistics/weekday.ts
```diff
@@ -0,0 +1,61 @@
...
+      defaultValue: 0,
+    },
+  },
+  {
+    indexes: [
+      {
+        unique: true,
+        fields: ['shortUrl', 'weekday', 'hours'],
+      },
+    ],
```
liangyuanruo, line 59: and here

src/server/util/device-type.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Infers the type of device from a user agent string.
+ *
+ * @param userAgent The user agent string to infer device type from.
+ */
+export function getDeviceType(userAgent: string) {
+  const parser = new UAParser(userAgent)
+  const deviceType = parser.getDevice().type
+
+  // Desktop browsers do not identify themselves in user agent.
+  if (!deviceType) return 'desktop'
```
liangyuanruo, line 13: can we consider whether the OS is able to be detected before determining whether desktop is acceptable, as per [this thread](https://github.com/faisalman/ua-parser-js/issues/182)?

src/server/util/time.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export function getLocalTime(): Time {
+  return {
+    weekday: getLocalWeekday(),
+    hours: getLocalDate().hours(),
+    date: getLocalDayGroup(),
+  }
+}
+
+export default { getLocalTime }
```
liangyuanruo, line 60: slightly confused by this statement vs the other `export` keywords used. is this line necessary?

src/server/services/RedirectService.ts
```diff
@@ -50,6 +50,13 @@ export class RedirectService implements RedirectServiceInterface {
         logger.error(`Unable to increment click count: ${error}`),
       )
 
+    // Update click statistics for link.
+    this.urlRepository
+      .updateLinkStatistics(shortUrl, userAgent)
```
liangyuanruo, line 55: can we merge `incrementClick` with `updateLinkStatistics` since they're really doing the same thing?

liangyuanruo: it might be good to merge/migrate this functionality to the statistics service - we should try and minimise cross-communication between services and repositories.

src/server/repositories/UrlRepository.ts
```diff
@@ -136,6 +141,57 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+      })
+      clickStats.increment(deviceType!)
+    }
+  }
+
+  public updateLinkStatistics: (
+    shortUrl: string,
+    userAgent: string,
+  ) => Promise<void> = async (shortUrl, userAgent) => {
+    const urlExists = Boolean(await Url.findOne({ where: { shortUrl } }))
```
liangyuanruo, line 181: is this check really necessary? it will be yet another network call to the DB

src/server/controllers/LinkStatisticsController.ts
kylerwsm: I have modified the code to use return statements and send explicit status codes

src/server/util/time.ts
kylerwsm: This line is not exactly necessary. But it allows for default exports without wildcard imports

src/server/repositories/UrlRepository.ts
kylerwsm: It was mostly for caution as I wanted to be very sure that nonexistent urls do not get added to the stats tables. In regards to cutting down on network calls, this check can definitely be removed, since the service layer already checks whether the user owns the short url.

kylerwsm: Alright, I have removed the check

src/server/util/device-type.ts
kylerwsm: I took some time to explore this, and found it very easy to extract the OS using the [useragent](https://www.npmjs.com/package/useragent) library. But doing this will mean that we have to keep our list of desktop OS updated. This can also be especially hard with linux distributions uniquely identifying themselves. Do you think we should really do this?

src/server/models/statistics/daily.ts
kylerwsm: I have removed the unnecessary index

src/server/models/statistics/weekday.ts
kylerwsm: I have removed the unnecessary index

src/server/models/statistics/devices.ts
kylerwsm: I have removed the unnecessary index

src/server/models/statistics/weekday.ts
```diff
@@ -0,0 +1,53 @@
...
+      validate: {
+        is: /^[a-z0-9-]+$/,
+      },
+    },
+    weekday: {
+      type: Sequelize.INTEGER,
+      primaryKey: true,
+      validate: {
+        rangeCheck(day: number) {
+          return day >= 0 && day < 7
```
liangyuanruo, line 34: can add another quick comment here to indicate which day does the integer represent?

src/server/repositories/LinkStatisticsRepository.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @param shortUrl The target short url to retrieve link statistics.
+   */
+  public findByShortUrl: (
+    shortUrl: string,
+  ) => Promise<LinkStatisticsInterface | null> = async (shortUrl) => {
+    const url = await Url.findOne({
+      where: { shortUrl },
+      include: [
+        { model: Devices, as: 'DeviceClicks' },
+        { model: Clicks, as: 'DailyClicks' },
```
liangyuanruo, line 35: i believe we also need to filter `DailyClicks` by date range too?

src/server/repositories/LinkStatisticsRepository.ts
```diff
@@ -0,0 +1,74 @@
...
+        } as DailyClicksInterface
+      })
+
+      const weekdayClicks = urlStats.WeekdayClicks.map((clicks) => {
+        return {
+          weekday: clicks.weekday,
+          hours: clicks.hours,
+          clicks: clicks.clicks,
+        } as WeekdayClicksInterface
+      })
```
liangyuanruo, line 62: is there a `toJSON()` method you can call?

src/server/util/device-type.ts
liangyuanruo: my concern is that all bot traffic will be classified as "desktop" since they are unlikely to have a device type. also UAParser can help via `getOS()`, right?

src/server/services/RedirectService.ts
liangyuanruo: finally - may i suggest that `updateLinkStatistics` be moved to the link statistics service, since we rely on it for data retrieval as well? the idea being that we want services to talk to each other to get stuff done, instead of each one reaching into the repository directly.

src/server/models/statistics/weekday.ts
kylerwsm: Added!

src/server/repositories/LinkStatisticsRepository.ts
kylerwsm: I have added that, and also ordering or daily clicks and weekday clicks. I hope to get some comments from you regarding my [implementation](https://github.com/opengovsg/GoGovSG/blob/stats/src/server/repositories/LinkStatisticsRepository.ts)

src/server/util/time.ts
kylerwsm: I have removed the export default

src/server/repositories/LinkStatisticsRepository.ts
kylerwsm: Yes there is, and it works really well. But it also includes unnecessary fields like updatedAt and createdAt which the frontend will not be consuming. I think specifying the field explicitly is beneficial to reduce over-sending data. What do you think?

src/server/util/device-type.ts
kylerwsm: As discussed, our check for bots will be: first to check for a JavaScript engine, followed by matching keywords found in common bots

src/server/services/RedirectService.ts
kylerwsm: I have merged `incrementClick` into `updateLinkStatistics` and moved `updateLinkStatistics` into `LinkStatisticsService`.

Additionally, I shifted `incrementClick`, `updateDailyStatistics`, `updateWeekdayStatistics`, and `updateDeviceStatistics` from `UrlRepository` to `LinkStatisticsRepository` to centralise link statistics updates in one repository.

src/server/repositories/LinkStatisticsRepository.ts
liangyuanruo: in that case you can use lodash's `pick` function which will be clearer and less verbose

PR #215, opened 2020-06-23
build(deps-dev): update webpack-related dependencies

LoneRifle:
## Problem and Solution

As part of ongoing efforts to bring dependencies up-to-date, update webpack-related packages and eliminate the ones we no longer use

- Remove url-loader (not used) and style-loader (no client CSS,
  styling now done via React themes)
- Update README to reflect this
- clean-webpack-plugin - Remove explicit argument to clean output directory, as v2+ defaults
to output.path
- Update everything else without further changes

Tested through `npm run dev`


liangyuanruo:
i wonder how hard is it to retrospectively introduce `create-react-app` here.

LoneRifle:
Why would you? Your app is already created

liangyuanruo:
Chunking, tree shaking and minification etc

LoneRifle:
Wouldn't it make more sense to introduce those features into webpack instead?

PR #217, opened 2020-06-24
feat: disallow non-ascii characters in desc field

JasonChong96:
## Problem

When users input descriptions, they're allowed to enter non-ascii characters such as chinese characters. Characters from other languages are not effectively indexed for search. Furthermore, GoGovSg is monolingual at the moment and visitors are unlikely to not know English.

## Solution

Disallow non-ascii characters on both the input field and backend API validation.

## Before & After Screenshots

**AFTER**:

image

## Tests

There should be an error message and the save button greyed out when a non-english character (e.g chinese) is inputted in the description field.

### DIFF
--- a/src/shared/util/validation.ts
+++ b/src/shared/util/validation.ts
@@ -60,5 +60,6 @@ export function isCircularRedirects(url: string, hostname?: string): boolean {
 }
 
 export function isPrintableAscii(string: string): boolean {
+  // Only accepts characters from 0x20 to 0x7F
   return /^[\x20-\x7F]*$/.test(string)
 }
--- a/src/shared/util/validation.ts
+++ b/src/shared/util/validation.ts
@@ -58,3 +58,7 @@ export function isCircularRedirects(url: string, hostname?: string): boolean {
   if (!hostname) return false
   return parse(url).hostname === hostname
 }
+
+export function isPrintableAscii(string: string): boolean {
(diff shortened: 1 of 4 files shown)


LoneRifle:
The error message may be confusing to non-technical users (ie, "what is ASCII?"). Could you rephrase this?

JasonChong96:
> The error message may be confusing to non-technical users (ie, "what is ASCII?"). Could you rephrase this?

Would saying that only english characters are allowed be less confusing?

LoneRifle:
We might need to have the designers involved. @pearlyong - thoughts?

pearlyong:
Suggested:
Description cannot contain special characters

JasonChong96:
Changed to "Description should only contain alphanumeric characters and symbols" as per discussion.

src/shared/util/validation.ts
```diff
@@ -58,3 +58,7 @@ export function isCircularRedirects(url: string, hostname?: string): boolean {
   if (!hostname) return false
   return parse(url).hostname === hostname
 }
+
+export function isPrintableAscii(string: string): boolean {
+  return /^[\x20-\x7F]*$/.test(string)
+}
```
liangyuanruo, line 64: can we document this function to say that this is merely the hexadecimal representation of the ascii charset

PR #218, opened 2020-06-24
refactor: append return statements after response callback

kylerwsm:
## Problem

Some of the response statements are not followed by a return statement.

## Solution

- Append return statements to the next line of response statements.
- Also added explicit status code to the existing `res.json` statement.
- Disabled an Eslint rule that removes return statements which it deems unnecessary.

### DIFF
--- a/src/server/controllers/LoginController.ts
+++ b/src/server/controllers/LoginController.ts
@@ -26,6 +26,7 @@ export class LoginController implements LoginControllerInterface {
     res: Express.Response,
   ) => void = (_, res) => {
     res.send(loginMessage)
+    return
   }
 
   public getEmailDomains: (
@@ -33,6 +34,7 @@ export class LoginController implements LoginControllerInterface {
     res: Express.Response,
   ) => void = (_, res) => {
     res.send(validEmailDomainGlobExpression)
+    return
   }
 
   public generateOtp: (
@@ -49,6 +51,7 @@ export class LoginController implements LoginControllerInterface {
     }
 
     res.ok(jsonMessage('OTP generated and sent.'))
+    return
   }
 
   public verifyOtp: (
@@ -69,11 +72,14 @@ export class LoginController implements LoginControllerInterface {
             `OTP hash verification failed, ${error.retries} attempt(s) remaining.`,
           ),
         )
-      } else if (error instanceof NotFoundError) {
-        res.unauthorized(jsonMessage('OTP expired/not found.'))
-      } else {
-        res.serverError(jsonMessage(error.message))
+        return
       }
+      if (error instanceof NotFoundError) {
+        res.unauthorized(jsonMessage('OTP expired/not found.'))
+        return
+      }
+      res.serverError(jsonMessage(error.message))
+      return
     }
   }
 
@@ -86,9 +92,10 @@ export class LoginController implements LoginControllerInterface {
     if (user) {
       const response = { ...jsonMessage('Logged in'), user }
       res.ok(response)
-    } else {
-      res.notFound(jsonMessage('User session not found'))
+      return
     }
+    res.notFound(jsonMessage('User session not found'))
+    return
   }
 }
 
--- a/src/server/controllers/LoginController.ts
+++ b/src/server/controllers/LoginController.ts
@@ -45,6 +45,7 @@ export class LoginController implements LoginControllerInterface {
       await this.authService.generateOtp(email)
     } catch (error) {
       res.serverError(jsonMessage(error.message))
+      return
     }
 
     res.ok(jsonMessage('OTP generated and sent.'))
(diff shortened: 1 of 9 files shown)


JasonChong96:
Technically in those cases the function call ends after the response statements. But I think its an improvement to have the explicit return statements to prevent bugs in case anyone adds any lines after.

kylerwsm:
@JasonChong96 I agree. Additionally, this issue might had been exacerbated by eslint, which right now removes return statements it deems unnecessary. This creates problems as we need to make conscious efforts to remember to add the return statements back when it becomes less unnecessary. I talked to @liangyuanruo about it, and am working on removing that eslint check.

yong-jie:
fired gitpod and did a grep project-wide grep check. confirmed that all relevant `res.status and res.json, res.ok, res.badRequest` etc calls now have the `return` statement

src/server/controllers/LoginController.ts
```diff
@@ -60,18 +64,22 @@ export class LoginController implements LoginControllerInterface {
...
+        return
+      }
+      if (error instanceof NotFoundError) {
         res.unauthorized(jsonMessage('OTP expired/not found.'))
-      } else {
-        res.serverError(jsonMessage(error.message))
+        return
       }
+      res.serverError(jsonMessage(error.message))
+      return
```
liangyuanruo, line 82: I think we should retain the `else-if` and `else` blocks here for greater code clarity.

src/server/controllers/LoginController.ts
```diff
@@ -84,9 +92,10 @@ export class LoginController implements LoginControllerInterface {
     if (user) {
       const response = { ...jsonMessage('Logged in'), user }
       res.ok(response)
-    } else {
-      res.notFound(jsonMessage('User session not found'))
+      return
     }
+    res.notFound(jsonMessage('User session not found'))
+    return
```
liangyuanruo, line 98: similarly here

src/server/controllers/RedirectController.ts
```diff
@@ -81,9 +81,10 @@ export class RedirectController implements RedirectControllerInterface {
           gaOnLoad: EventAction.LOADED,
           gaOnProceed: EventAction.PROCEEDED,
         })
-      } else {
-        res.status(302).redirect(longUrl)
+        return
       }
+      res.status(302).redirect(longUrl)
+      return
```
liangyuanruo, line 87: and here

src/server/controllers/UserController.ts
```diff
@@ -44,15 +44,19 @@ export class UserController implements UserControllerInterface {
...
+      if (error instanceof AlreadyExistsError) {
         res.badRequest(jsonMessage(error.message, MessageType.ShortUrlError))
-      } else {
-        logger.error(`Error creating short URL:\t${error}`)
-        res.badRequest(jsonMessage('Server error.'))
+        return
       }
+      logger.error(`Error creating short URL:\t${error}`)
+      res.badRequest(jsonMessage('Server error.'))
+      return
```
liangyuanruo, line 59: and here

src/server/controllers/UserController.ts
```diff
@@ -80,13 +84,15 @@ export class UserController implements UserControllerInterface {
...
       if (error instanceof NotFoundError) {
         res.notFound(jsonMessage(error.message))
-      } else {
-        logger.error(`Error editing URL:\t${error}`)
-        res.badRequest(jsonMessage(`Unable to edit short link "${shortUrl}"`))
+        return
       }
+      logger.error(`Error editing URL:\t${error}`)
+      res.badRequest(jsonMessage(`Unable to edit short link "${shortUrl}"`))
+      return
```
liangyuanruo, line 95: and here

src/server/controllers/UserController.ts
```diff
@@ -107,15 +113,19 @@ export class UserController implements UserControllerInterface {
...
+      if (error instanceof AlreadyOwnLinkError) {
         res.badRequest(jsonMessage(error.message))
-      } else {
-        logger.error(`Error transferring ownership of short URL:\t${error}`)
-        res.badRequest(jsonMessage('An error has occured'))
+        return
       }
+      logger.error(`Error transferring ownership of short URL:\t${error}`)
+      res.badRequest(jsonMessage('An error has occured'))
+      return
```
liangyuanruo, line 128: and here

src/server/controllers/UserController.ts
```diff
@@ -151,12 +161,14 @@ export class UserController implements UserControllerInterface {
...
     } catch (error) {
       if (error instanceof NotFoundError) {
         res.notFound(error.message)
-      } else {
-        res.serverError(jsonMessage('Error retrieving URLs for user'))
+        return
       }
+      res.serverError(jsonMessage('Error retrieving URLs for user'))
+      return
```
liangyuanruo, line 171: and here

kylerwsm:
Discussed with @liangyuanruo that the removal of else and else-if statements is because of an [ESLint](https://eslint.org/docs/rules/no-else-return) rule, and we will be keeping the rule for now.

PR #220, opened 2020-06-25
feat: include helpful description placeholder and tooltip

JasonChong96:
## Problem

Users might not understand what description field is for and write information which is not as useful for search.

## Solution

Include helpful placeholders and tooltips to nudge users to write descriptions which are useful for members of the public.

image

### DIFF
--- a/src/client/components/UserPage/Drawer/ControlPanel/index.tsx
+++ b/src/client/components/UserPage/Drawer/ControlPanel/index.tsx
@@ -258,7 +258,7 @@ export default function ControlPanel() {
     <>
       Link description{' '}
       <Tooltip
-        title="Link descriptions have a maximum character count of 200."
+        title="Write a description that will help users understand what your short link is for."
         arrow
         placement="top"
         classes={{ tooltip: classes.drawerTooltip }}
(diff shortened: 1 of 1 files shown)


src/client/components/UserPage/Drawer/ControlPanel/index.tsx
```diff
@@ -258,7 +258,7 @@ export default function ControlPanel() {
     <>
       Link description{' '}
       <Tooltip
-        title="Link descriptions have a maximum character count of 200."
+        title="Write a description that will help users understand what your short link is for."
```
liangyuanruo, line 261:
```suggestion
        title="Write a description that will help the public understand what your short link is for."
```

src/client/components/UserPage/Drawer/ControlPanel/index.tsx
```diff
@@ -516,7 +516,7 @@ export default function ControlPanel() {
                     )
                   }
                   error={!isDescriptionValid}
-                  placeholder=""
+                  placeholder="Tip: Include your agency name to tell users who this link belongs to."
```
liangyuanruo, line 519:
```suggestion
                  placeholder="Tip: Include your agency name to inform the public who this link belongs to."
```

PR #234, opened 2020-06-30
fix: new link highlighting

### LINKED ISSUE #231: Auto-highlighting of newly-created shortlinks is not working
**Describe the bug**
After creating a new link, the link is not highlighted on UserPage.
This bug was introduced in Release 1.19.0 code refactoring

**To Reproduce**
1. Log into go.gov.sg
2. Create a new link

**Expected behavior**
The new link should be highlighted on the UserPage after a link has been created.
Screenshot 2020-06-30 at 3 40 34 PM

**Screenshots**
Screenshot 2020-06-30 at 3 59 14 PM

JasonChong96:
## Problem

As of release 1.19, new links are not highlighted on users' url table. This is due to changes in backend response that were incompatible with the frontend implementation

## Solution

Revert change in behavior in backend response.

Fixes #231

### DIFF
--- a/src/server/controllers/UserController.ts
+++ b/src/server/controllers/UserController.ts
@@ -42,8 +42,13 @@ export class UserController implements UserControllerInterface {
     }
 
     try {
-      await this.urlManagementService.createUrl(userId, shortUrl, longUrl, file)
-      res.ok(jsonMessage(`Short link "${shortUrl}" has been updated`))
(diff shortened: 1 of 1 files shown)


LoneRifle:
> this technically breaks backwards compatibility right - should we detect the HTTP 200 OK on the frontend instead?

The revert strikes me as more useful than the current returning of `jsonMessage`. Would be helpful perhaps to keep.

PR #242, opened 2020-07-01
feat: user message banner

JasonChong96:
## Problem

We need an additional way to notify link owners of any service degradation or any other important messages.

## Solution

A banner for the user page that shows important messages to users.

The message is read from the USER_MESSAGE env_var.

## Before & After Screenshots

**AFTER**:
image
image
image

## Deploy Notes

- `USER_MESSAGE` : user message to be shown. It is optional and the banner will not be shown if this is empty or not supplied.


yong-jie:
I have updated staging env to contain the USER_MESSAGE env var

JasonChong96:
Added screenshot for IE

PR #243, opened 2020-07-01
fix: remove link statistics and ga served events

kylerwsm:
## Problem

- Remove link statistics logging code.
- Improvements from @LoneRifle earlier PR [here](https://github.com/opengovsg/GoGovSG/pull/238)

## Solution

- Move GA page view logging to after page redirect
- Remove link statistics logging, which as of right now is problematic
- Remove GA served events, which can lighten the network load on our servers

### DIFF
--- a/src/server/config.ts
+++ b/src/server/config.ts
@@ -168,4 +168,6 @@ export const bucketEndpoint =
 export const accessEndpoint =
   process.env.ACCESS_ENDPOINT || 'http://localhost:4566'
 
+export const dbPoolSize = Number(process.env.DB_POOL_SIZE) || 20
+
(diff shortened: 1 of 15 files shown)


liangyuanruo:
hold off from merging first.

src/server/config.ts
```diff
@@ -168,4 +168,6 @@ export const bucketEndpoint =
 export const accessEndpoint =
   process.env.ACCESS_ENDPOINT || 'http://localhost:4566'
 
+export const dbPoolSize = Number(process.env.DB_POOL_SIZE) || 20
```
liangyuanruo, line 171:
```suggestion
export const dbPoolSize = Number(process.env.DB_POOL_SIZE) || 40
```

PR #244, opened 2020-07-02
Update transition page UI

yong-jie:
## Problem

The transition page has a high drop-off rate. Our current hypothesis is that the current transition page looks like an error page and is causing people to close the page without reading the contents.

## Solution

A new design has been implemented in hopes of deviating from a generic error page and more like an informational "before you proceed" site.

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


package.json
```diff
@@ -14,7 +14,7 @@
     "postbuild": "copyfiles -u 1 src/server/views/**/* build && copyfiles -u 4 src/server/services/QrCodeService/assets/**/* build/server/services/QrCodeService",
     "start": "node build/server/index.js",
     "client-dev": "webpack-dev-server --mode development --host 0.0.0.0 --devtool inline-source-map --hot",
-    "server-dev": "ts-node-dev --respawn --transpileOnly --inspect=0.0.0.0 -- src/server/index.ts",
+    "server-dev": "ts-node-dev --poll --respawn --transpileOnly --inspect=0.0.0.0 -- src/server/index.ts",
```
JasonChong96, line 17: what does this do

public/assets/transition-page/styles/transition-page.css
```diff
@@ -143,3 +145,171 @@ img {
...
+.proceed-button img {
+  margin-left: 6px;
+}
+
+.bottom-half {
+  text-align: center;
+  position: absolute;
+  bottom: 70px;
+  left: 50%;
+  transform: translateX(-50%);
```
JasonChong96, line 247: Does this work on IE11?

package.json
kylerwsm: This might be something I left behind while I was working on it. It helps with the problem of ts-node-dev not restarting in networked environments. More information here: https://github.com/remy/nodemon#application-isnt-restarting

src/server/views/transition-page.ejs
```diff
@@ -57,6 +63,17 @@
         // Ensures that proceeded event sends at most once.
         var proceedIsLogged = false
 
+        // For rotating links on graphic.
+        var rotatingLinks = ['whatsapp', 'sppr']
```
kylerwsm, line 67: I think we can pull this from environment variables 🤔

package.json
yong-jie: to add, without the --poll flag, my local development environment also struggles to hot-reload, which is why i thought we could also just merge this change into develop at the same time

kylerwsm: the legacy watch mode counterpart in ts-node-dev is --poll

public/assets/transition-page/styles/transition-page.css
yong-jie: pushing to stage to test now, update in 15 min

src/server/controllers/RedirectController.ts
```diff
@@ -74,6 +74,7 @@ export class RedirectController implements RedirectControllerInterface {
 
         res.status(200).render(TRANSITION_PATH, {
           escapedLongUrl: RedirectController.encodeLongUrl(longUrl),
+          rotatingLinks: JSON.stringify(linksToRotate),
```
liangyuanruo, line 77:
```suggestion
          rotatingLinks: JSON.stringify(linksToRotate) || '',
```

liangyuanruo:
@pearlyong will be proposing an automatic transition page today at sync up, so further changes will be necessary.

src/server/views/transition-page.ejs
```diff
@@ -31,7 +31,7 @@
         </div>
     </div>
     <div class="bottom-half">
-        <p>You will be redirected in <span id="countdown-seconds"></span> second<span id="s">s</span></p>
+        <p id="url" data-href="<%- escapedLongUrl %>">You will be redirected in <span id="countdown-seconds">8</span> second<span id="s">s</span></p>
```
yong-jie, line 34: setting `data-href` in an arbitrary html element was the best hack I could think of to pass data into gtag.ejs. I borrowed this idea from @LoneRifle, but all suggestions for improvements are welcome

PR #253, opened 2020-07-03
Fix bug/code smell where GA cookie is set after a server redirect

### LINKED ISSUE #250: Fix bug where GA cookie is set after a server redirect
**Describe the bug**
When a new visitor without a GA cookie visits a short link, the server attempts to set a cookie after a response has already taken place. This results in a HTTP 500 Internal Server Error being logged when in fact a HTTP 302 Redirect has already been returned to the client.

**To reproduce**
1. Clear cookies
2. Visit a short link
3. Verify that browser is redirected successfully
4. Observe stack trace and HTTP 500 in logs

yong-jie:
## Problem

When a new visitor without a GA cookie visits a short link, the server attempts to set a cookie after a response has already taken place. This results in a HTTP 500 Internal Server Error being logged when in fact a HTTP 302 Redirect has already been returned to the client.

Closes #250

## Solution

- [x] Move cookie generation and setting to another place before setting res
- [ ] Don't create a new `logRedirect` function in every redirect
- [ ] Don't pass res and req into other functions
- [x] Fix broken tests

### DIFF
--- a/src/server/controllers/RedirectController.ts
+++ b/src/server/controllers/RedirectController.ts
@@ -56,10 +56,16 @@ export class RedirectController implements RedirectControllerInterface {
       const logRedirect = () => {
         this.analyticsLogger.logRedirectAnalytics(
           req,
-          res,
           shortUrl.toLowerCase(),
           longUrl,
         )
+      }
+
+      const generatedCookie = this.analyticsLogger.generateCookie(
+        req.headers.cookie,
+      )
+      if (generatedCookie) {
+        res.cookie(...generatedCookie)
       }
 
       req.session!.visits = visitedUrls
--- a/src/server/services/analyticsLogger.ts
+++ b/src/server/services/analyticsLogger.ts
@@ -13,25 +13,30 @@ export interface AnalyticsLogger {
    */
   logRedirectAnalytics: (
     req: Express.Request,
-    res: Express.Response,
     shortUrl: string,
     longUrl: string,
   ) => void
+
+  generateCookie: (
+    cookie?: string,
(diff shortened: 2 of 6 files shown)


src/server/controllers/RedirectController.ts
```diff
@@ -56,12 +56,18 @@ export class RedirectController implements RedirectControllerInterface {
...
           longUrl,
         )
       }
 
+      const generatedCookie = this.analyticsLogger.generateCookie(
+        req.headers.cookie,
+      )
+      if (generatedCookie) {
+        res.cookie(...generatedCookie)
+      }
```
LoneRifle, line 69: The logic here is a bit roundabout - basically we are saying "speculatively generate the cookie, and if we happen to have one, set it". It really should say "check for the presence of a cookie, and if absent, generate one and set it"

src/server/services/analyticsLogger.ts
```diff
@@ -13,25 +13,30 @@ export interface AnalyticsLogger {
...
+
+  generateCookie: (
+    cookie?: string,
+  ) => [string, string, { maxAge: number }] | null
 }
 
 @injectable()
 export class GaLogger implements AnalyticsLogger {
   logRedirectAnalytics: (
     req: Express.Request,
```
liangyuanruo, line 28: can we refactor out the code smell with the need for `req` here?

LoneRifle: Given the turnaround time, let's hold off on that. I can refactor this after the release

yong-jie:
Acknowledged that code refactors are not the priority in this PR, and will be addressed in a separate PR after today's release.

PR #264, opened 2020-07-07
Link Statistics with SQL Server Functions

kylerwsm:
## Problem

The previous attempt to log link statistics is too computationally expensive.

We needed an alternative that is able to handle and thrive under high user traffic.

## Solution

Rather than logging using Sequelize model instances, we turned to SQL server functions. This function will be initialised by the server when it starts, and will be called to increment the relevant statistics on redirects.

Additionally, @liangyuanruo and I have load tested this implementation with the following results:
- 199,978/200,000 clicks were logged
- 22 network requests failed from HTTP 502 Bad Gateway and did not obtain a redirect
- Load was about ~200/s


src/server/models/functions/index.ts
```diff
@@ -0,0 +1,24 @@
...
+} from './updateLinkStatistics'
+
+/**
+ * Syncs database functions.
+ */
+export async function syncFunctions() {
+  // Removes any previous implementations of link statistics.
+  await sequelize.query(dropLinkStatistics, {
+    type: QueryTypes.RAW,
+  })
```
liangyuanruo, line 16: Remove this, otherwise there will be a brief period of time where the function does not exist. just rely on `CREATE OR REPLACE` instead.

src/server/models/functions/updateLinkStatistics.ts
```diff
@@ -0,0 +1,67 @@
+import { Url } from '../url'
+import { Clicks } from '../statistics/daily'
+import { Devices } from '../statistics/devices'
+import { WeekdayClicks } from '../statistics/weekday'
+
+/**
+ * Name of the link statistics update SQL function.
+ */
+export const rawFunctionName = 'update_link_statistics'
```
liangyuanruo, line 9: can we move this file into the service instead, so that everything is in the same place?

also i'm not sure if the `rawFunctionName` abstraction is really necessary - it doesn't appear to add value since there is only one function.

src/server/services/LinkStatisticsService.ts
```diff
@@ -7,28 +7,37 @@ import { LinkStatisticsInterface } from '../../shared/interfaces/link-statistics
 import { UserRepositoryInterface } from '../repositories/interfaces/UserRepositoryInterface'
 import { NotFoundError } from '../util/error'
 import { logger } from '../config'
+import { DeviceCheckService } from './DeviceCheckService'
```
liangyuanruo, line 10: we only need _DeviceCheckService_ for analytics, right? should we fold it into _LinkStatisticsService_?

src/server/models/functions/updateLinkStatistics.ts
```diff
@@ -0,0 +1,67 @@
...
+  ON CONFLICT ("shortUrl")
+  DO UPDATE SET "others" = "${devicesTable}"."others" + 1;
+END IF;
+-- Update daily clicks.
+INSERT INTO "${clicksTable}" ("shortUrl", "date", "clicks", "createdAt", "updatedAt")
+VALUES (inputShortUrl, current_timestamp::date, 1, current_timestamp, current_timestamp)
+ON CONFLICT ("shortUrl", "date")
+DO UPDATE SET "clicks" = "${clicksTable}"."clicks" + 1;
+-- Update weekday clicks.
+INSERT INTO "${weekdayTable}" ("shortUrl", "weekday", "hours", "clicks", "createdAt", "updatedAt")
```
liangyuanruo, line 60: what is your plan to convert and display data in GMT+8 to the user?

kylerwsm: From my previous testings, the timestamps the function is using from the database is already in GMT+8. And since the logged link statistics data strips all timezone information, there should not be a need to convert the timezone of the data to the user.

src/server/models/functions/index.ts
kylerwsm: Alright, I have removed the drop query

src/server/services/LinkStatisticsService.ts
kylerwsm: I am not so sure about that, as keeping `DeviceCheckService` separated does help with testing, and possibly code reuse in the future

src/server/models/functions/updateLinkStatistics.ts
kylerwsm: Initially, I abstracted out the function name because if there is any disparity between the initialised SQL function's name and the function name called when updating link statistics, we will only be able to catch it in runtime. But since there is only one function, I guess we can not abstract it. Additionally, now that I think of it, hardcoding it might also improve performance with lesser string interpolation.

kylerwsm: For now, I will also be shifting the file to the repository, which should help put everything in one place, and ensures better consistency between the function initialisation and calls.

src/server/models/functions/updateLinkStatistics.ts
liangyuanruo: i believe that's only because your localhost time is in GMT+8. the servers are in UTC.

```sql
select current_timestamp;

> 2020-07-14 04:38:38.918593+00
```

kylerwsm: I have rebased, and added explicit timezones to the function

PR #265, opened 2020-07-11
feat: display the link user is editing on edit link drawer

### LINKED ISSUE #180: Indicate which link the user is editing on edit link drawer
Currently when user select on a link to edit, the drawer does not indicate which link the user is editing. This may cause confusion for users especially those who has a large number of links in their account. 

Display on the drawer the link that the user is editing
Screenshot 2020-06-11 at 2 28 43 AM

xming13:
## Problem

Closes #180 

## Solution

Update according to the proposed design in https://github.com/opengovsg/GoGovSG/issues/180#issuecomment-645731155

**Improvements**:

- Change `buttonText` prop in `CopyButton` to be optional, defaults to the url to copy if omitted.
- Removed unused styles
- Destructure props for `CopyButton`

## Before & After Screenshots

**BEFORE (Desktop)**:
image

**AFTER (Desktop)**:
image

**BEFORE (Mobile)**:
image

**AFTER (Mobile)**:
image

### DIFF
--- a/src/client/components/UserPage/Drawer/ControlPanel/DrawerHeader.tsx
+++ b/src/client/components/UserPage/Drawer/ControlPanel/DrawerHeader.tsx
@@ -13,9 +13,6 @@ import CopyButton from '../../Widgets/CopyButton'
 const useStyles = makeStyles((theme) =>
   createStyles({
     drawerTitleDiv: {
-      display: 'flex',
-      alignItems: 'flex-end',
-      justifyContent: 'space-between',
       marginBottom: theme.spacing(6.5),
       marginTop: theme.spacing(1.5),
       [theme.breakpoints.up('md')]: {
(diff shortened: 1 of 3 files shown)


src/client/components/UserPage/Drawer/ControlPanel/DrawerHeader.tsx
```diff
@@ -13,21 +13,15 @@ import CopyButton from '../../Widgets/CopyButton'
...
         marginTop: 0,
       },
     },
-    copyLinkDiv: {
-      display: 'flex',
-    },
-    copyIcon: {
-      marginRight: 5,
+    copyButtonWrapper: {
+      marginLeft: -8,
```
liangyuanruo, line 24: Could we avoid the use of negative margins?

xming13: The `CopyButton` component (http://localhost:3000/test-google) has a 8px padding. The use of negative margin is to align the `CopyButton` with other elements as per design. 

If the elements do not need to be visually aligned, we can remove the negative margin. Otherwise, I think it is preferred to have negative margin rather than to restructure the layout to achieve the design.

Without negative margin
image

With negative margin,
image

With negative margin, hovered state
image

PR #277, opened 2020-07-20
feat: index tweaks and exclude blank description from search

JasonChong96:
## Problem

Links with no description should be excluded from search as some links should not be discoverable.

## Solution

Add a condition to search to exclude urls with blank descriptions

**Improvements**:

- Index will be managed by sequelize instead
- Explicit configuration of search weights allow easy reconfiguration either through `config.ts` or env vars
- Empty state graphic should no longer overlap the sort panel
- Nav buttons hidden on search page

**Deploy notes**:
- Run migration script before deploying on production. This will allow sequelize to create the new updated search index in place of the old one.

### DIFF
--- a/src/server/models/search.ts
+++ b/src/server/models/search.ts
@@ -1,33 +1,10 @@
-import { QueryTypes } from 'sequelize'
-import { Url } from './url'
-import { sequelize } from '../util/sequelize'
 import { StorableUrlState } from '../repositories/enums'
 
-const { tableName } = Url
-const INDEX_NAME = 'urls_weighted_search_idx'
-
-// Warning: This expression has to be EXACTLY the same as the one used in the index
+// Warning: This expression and conditions has to be EXACTLY the same as the one used in the index
 // or else the index will not be used leading to unnecessarily long query times.
 export const urlSearchVector = `
-  setweight(to_tsvector('english', ${tableName}."shortUrl"), 'A') ||
-  setweight(to_tsvector('english', ${tableName}."description"), 'B')
+  setweight(to_tsvector('english', urls."shortUrl"), 'A') ||
+  setweight(to_tsvector('english', urls."description"), 'B')
 `
 
 export const urlSearchConditions = `urls.state = '${StorableUrlState.Active}' AND urls.description != ''`
-
-const dropIndex = `DROP INDEX IF EXISTS ${INDEX_NAME};`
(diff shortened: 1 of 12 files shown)


src/server/models/search.ts
```diff
@@ -0,0 +1,26 @@
...
+// or else the index will not be used leading to unnecessarily long query times.
+export const urlSearchVector = `
+  setweight(to_tsvector('english', ${tableName}."shortUrl"), 'A') ||
+  setweight(to_tsvector('english', ${tableName}."description"), 'B')
+`
+
+export const urlSearchConditions = `urls.state = '${StorableUrlState.Active}' AND urls.description != ''`
+
+const dropIndex = `DROP INDEX IF EXISTS ${INDEX_NAME};`
+const createIndex = `CREATE INDEX urls_weighted_search_idx ON urls USING gin ((${urlSearchVector}))
```
liangyuanruo, line 19: what happens when we deploy to multiple servers and the queries are interleaved? the DROP statement appears to be safe but will the CREATE INDEX statement fail if an index already exists?

liangyuanruo: will using a BEGIN ... COMMIT help?

JasonChong96: Will add BEGIN, COMMIT. Will also be adding "IF NOT EXISTS" to the create statement. This way, at least one of the deployments should be able to commit with the new index.

liangyuanruo:
Code lgtm, but one nagging doubt I have is the effect of deploying this to the database during a period of high search usage. It's generally considered bad practice to create indexes on deployment, but the performance impact is probably acceptable give that it takes only ~1s.

Maybe @LoneRifle can advise as well.

LoneRifle:
You really really should be letting sequelize handle this for you:
https://sequelize.org/master/manual/indexes.html

JasonChong96:
> You really really should be letting sequelize handle this for you:
> https://sequelize.org/master/manual/indexes.html

Will check if this is possible for the search index

LoneRifle:
What's the intention actually? Do you wish to keep the index schema fresher than what `sequelize.sync` tries to do? Bear in mind that because `sequelize.sync` uses `CREATE X IF NOT EXISTS`, it tends to refrain from altering existing db objects

JasonChong96:
It was initially done with a sql migration script as sequelize doesn't have support for postgres full text search so I didn't think it would be possible to use sequelize's sync for a text search index but that might not be the case

JasonChong96:
Changed to use sequelize to manage index

PR #279, opened 2020-07-20
refactor(ga): decouple Express.Request from analytics

### LINKED ISSUE #254: Code refactor for RedirectController
A lack of separation of concerns between controller vs business logic contributed to bugs occurring on the redirect route.

Refactor the code such that `req` and `res` is not passed to services such as `analyticsLogger.logRedirectAnalytics`. The redirect controller should be the only function that handles server responses.

LoneRifle:
## Problem

Closes #254

## Solution

- Move biz logic for creating GA page view requests into its own func,
so that we can more easily decouple Express.Request from analytics
  - Declare and implement `createPageViewHit()`, call that from
  `sendPageViewHit()`
  - Augment tests to expect mockAnalyticsLogger to record and compare
  page view hits, to facilitate later refactoring which changes
  func arguments for `logRedirectAnalytics()`

- Change analytics logging such that responsibility of creating
view page requests to GA falls on the controller by invoking an
utility function, which is then passed to `analyticsLogger`

  - Call `createPageViewHit` from controller, pass result to
  `analyticsLogger`
  - Change args to `logRedirectAnalytics()` such that it receives
  an object of generic type T, usually the type of the page view hit
  payload. Change GaLogger so that it implements
  AnalyticsLogger<IGaViewPageForm>
  - Rework tests to account for changes in signature

- Re-org naming

## Tests

To be done in staging, exercising GA


src/server/services/interfaces/AnalyticsLoggerService.ts
```diff
@@ -0,0 +1,10 @@
+export interface AnalyticsLoggerService<T> {
+  /**
+   * @param {T} pageViewHit - A page view request payload.
+   */
+  logRedirectAnalytics: (pageViewHit: T) => void
+
+  generateCookie: (
+    cookie?: string,
+  ) => [string, string, { maxAge: number }] | null
```
liangyuanruo, line 9: Should we start to give a name to the type `[string, string, { maxAge: number }] | null`?

LoneRifle: It's an array - basically arguments to be spread onto `res.cookie`. We could refactor this later

LoneRifle:
> 2. remove the need to pass `req` to `createPageViewHit`? in fact, we could refactor to make `createPageViewHit` an internal function within `logRedirectAnalytics` and call that at the appropriate junction, instead of exposing it to `RedirectController`?
Wouldn't that defeat the purpose of this PR? Isn't the intent to ensure that only the controller deals with the handling of Express.Request _before_ it gets passed to the Service?

liangyuanruo:
> I would contend that this is what the PR is doing - it's delegating work to other modules and doesn't actually know how a req is transformed into a pageViewHit

True, but I still feel that the level of abstraction may be incomplete . If `pageViewHit` is only used by `logRedirectAnalytics`, then the logic could also be encapsulated as an internal function.

LoneRifle:
While true, shouldn't the need to prevent Express.Request leaking into analyticsLogger take precedence?

LoneRifle:
Given you've made the approval I will merge this in for now. We can discuss what further changes - if any - should be made at a later point

PR #290, opened 2020-07-21
feat: gosearch ui tweaks

JasonChong96:
## Problem

Certain icon sizes are not in line with the Figma design. Furthermore it is not easy to reach the contact emails on search.

## Solution

- Allow contact emails to be clickable
- Adjust icon sizes
- Add border to search input on hover

### DIFF
--- a/src/client/components/widgets/GoSearchInput/index.tsx
+++ b/src/client/components/widgets/GoSearchInput/index.tsx
@@ -134,7 +134,6 @@ const GoSearchInput: FunctionComponent<GoSearchInputProps> = ({
           variant="outlined"
           InputProps={{
             className: classes.searchInput,
-            // disableUnderline: true,
             startAdornment: (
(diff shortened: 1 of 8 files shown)


src/client/components/widgets/GoSearchInput/index.tsx
```diff
@@ -125,12 +125,13 @@ const GoSearchInput: FunctionComponent<GoSearchInputProps> = ({
           value={query}
           onChange={(e) => onQueryChange(e.target.value)}
           onKeyPress={onKeyPress}
+          variant="outlined"
           InputProps={{
             className: classes.searchInput,
-            disableUnderline: true,
+            // disableUnderline: true,
```
yong-jie, line 131: do we plan to eventually restore this config option?

JasonChong96: nope, will remove

PR #299, opened 2020-07-22
fix: explicit parsing config on query

JasonChong96:
## Problem

While the default configuration for parsing queries is set to english on most database configurations and on the local docker image, it might be different in production environments.

For example, in AWS RDS, the default is set to `simple` which does not stem query words.

## Solution

Explicitly state that the configuration to use in the query is `english`


JasonChong96:
> How did you find out?

On production there's a link description that includes "Introduction" but it did not appear when searching "Introduction". I found that by searching it only appears when i searched for its root word "Introduct" which means that the query words weren't being stemmed according the the english config

PR #306, opened 2020-07-24
feat(upload): scan files for viruses, filter for types

### LINKED ISSUE #274: Add scanning of files before upload [By Jul 24]
It is likely we need not over-protect file uploads, given it is only available to authenticated gov.sg users. However, to prevent escalation of access rights if a gov.sg email is compromised or ignorance if a public officer unknowingly uploads a malicious file, it is still useful to prevent bad files from uploaded.

Cloudflare WAF does not seem to scan files for vulnerabilities. Can we try using packages such as clamscan? This article provides a test file with innocuous malware: https://blog.didierstevens.com/2015/08/28/test-file-pdf-with-embedded-doc-dropping-eicar/

To implement this by Jul 24 so testers can do remediation testing.

LoneRifle:
## Problem

Closes #274 

## Solution

- Add a controller that will perform checks on uploaded files
- Implement both a filter for file type and a service that submits files to virus scanning APIs
- Provide VirusScanService implementations for VirusTotal and Cloudmersive

## Tests

- Both an innocuous PNG file as well as the EICAR test file should be uploaded. Go should accept the former but reject the latter.
- Go should also reject an attempt to upload an `.mp4` file, amongst others outside the allowed list of types

## Deploy Notes

The new environment variable detailed below should be added to production to enable scanning

- `CLOUDMERSIVE_KEY` : API key for access to Cloudmersive


src/server/services/VirusScanService.ts
```diff
@@ -0,0 +1,77 @@
...
+import { logger, virusTotalKey } from '../config'
+import { VirusScanServiceInterface } from './interfaces/VirusScanServiceInterface'
+
+const VIRUS_TOTAL_ENDPOINT = 'https://www.virustotal.com/api/v3'
+
+@injectable()
+export class VirusScanService implements VirusScanServiceInterface {
+  private static async checkResponseThenJSON(response: Response): Promise<any> {
+    const result = await response.json()
+    if (!response.ok) {
```
liangyuanruo, line 14: should the HTTP status be checked before deserialisation?

LoneRifle: yes, that's what the implementation does. At any rate VirusScanService is not actually used - look at the other impl

src/server/services/VirusScanService.ts
```diff
@@ -0,0 +1,77 @@
+import fetch from 'cross-fetch'
+import CrossFormData from 'form-data'
+
+import { injectable } from 'inversify'
+import { logger, virusTotalKey } from '../config'
+import { VirusScanServiceInterface } from './interfaces/VirusScanServiceInterface'
+
+const VIRUS_TOTAL_ENDPOINT = 'https://www.virustotal.com/api/v3'
```
liangyuanruo, line 8: actually, i believe we're going with cloudmersive - can we delete all this code?

LoneRifle: Sure, let me have that done

src/server/controllers/FileCheckController.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  public checkFile: (
+    req: Request,
+    res: Response,
+    next: NextFunction,
+  ) => Promise<void> = async (req, res, next) => {
+    const file = req.files?.file
+
+    if (Array.isArray(file)) {
+      res.badRequest(jsonMessage('Only single file uploads are supported.'))
```
liangyuanruo, line 34: consider using `422 UNPROCESSABLE ENTITY`

src/server/controllers/FileCheckController.ts
```diff
@@ -0,0 +1,60 @@
...
+      return
+    }
+
+    if (file) {
+      if (!(await this.fileTypeFilterService.hasAllowedType(file))) {
+        res.badRequest(jsonMessage('File type disallowed.'))
```
liangyuanruo, line 40: consider using `415 UNSUPPORTED MEDIA TYPE`

src/server/controllers/FileCheckController.ts
LoneRifle: This was a copypaste from elsewhere in the codebase. I'll update that part too

PR #314, opened 2020-07-27
feat(helmet): implement CSP policy

### LINKED ISSUE #291: Implement HTTP Security Headers
Beef up the security of the web application, to protect against possible attacks e.g. XSS and so on.

Implement a loader that configures the Express application to inject the following security headers with HelmetJS

CSP headers
- [ ] Content Security Policy
- [ ] Content Security Policy Report Only (can set this via an optional env var to Sentry.io)

Others
- [x] Strict Transport Security (HSTS)
- [ ] Upgrade Insecure Requests
- [x] X-Content-Type-Options
- [x] X-Frame-Options
- [x] X-XSS-Protection

LoneRifle:
## Problem

Closes #291 

## Solution
Tighten how the browser interacts with the rest of the world when
accessing go.gov.sg. Avoid having to inject a nonce into the transition
page by moving the sole inline script into its own templated file.

- specify a content security policy using helmet
- allow requests from Google, Sentry (if present)
- move inline script in transition-page.ejs to its own template, and
  map this to `/assets/transition-page/js/gtag.js`

## Screenshots
image

## Deploy Notes

- `CSP_ONLY_REPORT_VIOLATIONS` - when true, only reports CSP violations
- `CSP_REPORT_URI` - the URI that a browser can POST CSP violations to

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -18,7 +18,15 @@ bindInversifyDependencies()
 import api from './api'
 
 // Logger configuration
-import { cookieSettings, logger, sessionSettings, trustProxy } from './config'
+import {
+  cookieSettings,
+  cspOnlyReportViolations,
+  cspReportUri,
+  logger,
+  sentryDns,
+  sessionSettings,
+  trustProxy,
+} from './config'
 
 // Services
 const SessionStore = connectRedis(session)
@@ -35,6 +43,7 @@ import { container } from './util/inversify'
 import { DependencyIds } from './constants'
 import { Mailer } from './services/email'
 import { RedirectControllerInterface } from './controllers/interfaces/RedirectControllerInterface'
+import parseDomain from './util/domain'
 // Define our own token for client ip
 // req.headers['cf-connecting-ip'] : Cloudflare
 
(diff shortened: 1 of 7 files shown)


src/server/views/gtag.ejs
```diff
@@ -0,0 +1,23 @@
+window.dataLayer = window.dataLayer || [];
```
liangyuanruo, line 1: should this be renamed to `.js` / moved into a `js` folder?

src/server/index.ts
```diff
@@ -117,14 +156,19 @@ initDb()
     // Log http requests
     app.use(morgan(MORGAN_LOG_FORMAT))
 
+    const redirectController = container.get<RedirectControllerInterface>(
+      DependencyIds.redirectController,
+    )
     // API configuration
     app.use('/api', ...apiSpecificMiddleware, api) // Attach all API endpoints
+    app.get(
+      '/assets/transition-page/js/gtag.js',
```
liangyuanruo, line 165: would it be cleaner to serve the file from a static endpoint instead?

src/server/index.ts
```diff
@@ -60,8 +69,36 @@ const MORGAN_LOG_FORMAT =
...
+}
+if (sentryDns) {
+  connectSrc.push(sentryDns)
+}
+
 const app = express()
 app.use(helmet())
+app.use(
+  helmet.contentSecurityPolicy({
+    directives: {
```
liangyuanruo, line 84: could we also include `frameSrc`, `frameAncestors` to protect against frame injection XSS?

src/server/views/gtag.ejs
LoneRifle: No - this is a template that will be interpolated with values held by the backend

src/server/index.ts
LoneRifle: Not possible given that we do not want to have our GA tracking ids as part of the contents of the file. _Maybe_ we could think about spitting out a static file from GA as part of the build... but that is too much work for a very specific part of this codebase

src/server/index.ts
LoneRifle: Wouldn't `frameSrc` fallback on `defaultSrc` if not specified? I'll look into frameAncestors now

PR #330, opened 2020-07-29
feat(helmet): implement CSP policy

### LINKED ISSUE #291: Implement HTTP Security Headers

LoneRifle:
## Problem

Closes #291

## Solution

- Reapply #314 
- Loosen CSP policy to allow inline styles, data blob images
- Trim unused embedded JavaScript in 404 page
- Rework gtag.js and transition page to bind the click handler at
  runtime as opposed to inlining it, to comply with CSP

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,7 @@ services:
       - ROTATED_LINKS=whatsapp,passport,spsc,sppr
       - AWS_ACCESS_KEY_ID=foobar
       - AWS_SECRET_ACCESS_KEY=foobar
+      - CSP_ONLY_REPORT_VIOLATIONS=true
       - CLOUDMERSIVE_KEY=
     volumes:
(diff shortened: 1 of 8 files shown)


LoneRifle:
you can introduce a nonce, but you need to _somehow_ modify webpack's output so that Express can dynamically inject the nonce into the bundle. For reference, @isomerpages allows `unsafe-inline`

PR #346, opened 2020-08-03
Integrate with Github Actions

yong-jie:
## Problem

I think there is value in trying out github actions as our CI provider, with the main reason being support for concurrent builds without hoarding travis' concurrent builds limit. 

## Approach

Traditionally, our CI/CD pipeline consists of the following steps done in sequence:

Lint -> Test -> Build -> Push to ECR -> Deploy to EB

With Github Actions, we can do Lint, Test and Build+Push concurrently, before Deploying to EB after the previous 3 jobs have finished.  This an optimistic approach because it builds and pushes the docker image to ECR without checking if Lint and Test have passed. Here is the primary tradeoff:

**Pros**:
- Overall time to deployment drastically reduces because the build step happens concurrently with tests.

**Cons**:
- Potentially wasted computation if either Lint or Test builds fail.

**Rationale**:
- Tests and lint rarely fails in the deployment merge because we would have undergone both the release checklist and all other checks before merging into `release`.
- This approach is still safe because the final deployment step still waits for all of Lint, Test, and Build+Push to pass before executing.

## Things to note

**Syntax is lacking**
The Github Actions syntax is still in its early phases of development and is missing some quality-of-life configuration options. This means I had to rely on some workarounds to get some things done.

An example is the dedicated job called `gatekeep` which helps determine whether the build & deploy jobs need to be executed. The workaround is necessary because I can't utilise the `env` context inside the `if` statements for jobs, so something like `if: branch == env.STAGING || branch == env.PRODUCTION` is not possible.

https://github.com/opengovsg/GoGovSG/blob/b5228713aa1ac35896a0f28594120fb6176c1d49/.github/workflows/ci.yml#L56-L72

**Weak env var selection mechanism**
Our builds use different versions of env vars depending on the branch that the Github Action is being run on. At present, I could not find a nice way to select variants of env vars depending on the branch, so I have resorted to using scripts to set our necessary EB env vars.

https://github.com/opengovsg/GoGovSG/blob/b5228713aa1ac35896a0f28594120fb6176c1d49/.github/workflows/ci.yml#L129-L146

**Lack of inter-workflow dependency/message-passing**
It's a universally good practice to split blocks of logic/code into different files. This is not yet supported in Github Actions and I am lumping both CI and CD into a single workflow file for the time being.

## List of things done

- [x] Build (tsc)
- [x] Lint
- [x] Test
- [x] Integrate with Coveralls
- [x] Build (Docker)
- [x] Send to ECR
- [x] Set up conditional jobs
- [x] Deploy to EB


.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+          SHA: ${{ github.sha }}
+      - run: docker build --tag my-image .
+      - name: Push to ECR
+        uses: jwalton/gh-ecr-push@v1
+        with:
+          access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          region: ap-southeast-1
+          local-image: my-image
+          image: gogovsg:${{steps.extract_tag.outputs.tag}}
```
LoneRifle, line 102:
```suggestion
          image: $REPO:${{steps.extract_tag.outputs.tag}}
```
nitpick?

LoneRifle (approved):
lgtm otherwise; please:

- review comments
- figure out if we can immediately replace Travis

.github/workflows/ci.yml
yong-jie: fixed!

PR #351, opened 2020-08-04
feat(links): scan for threats with Cloudmersive

LoneRifle:
## Problem
Links submitted by users are not checked for threats, and may be malicious

## Solution
Use Cloudmersive's Virus Scan API to check URLs submitted by users through 
a new controller

- Rename CloudmersiveVirusScanService to CloudmersiveScanService
- Declare UrlThreatScanServiceInterface and provide
  CloudmersiveScanService as an implementation, delegating to the
  Virus Scan API client
- Declare and implement UrlCheckController and interface, along with
  scaffolding to drop into Express
- Rework inversify config so that a single CloudmersiveScanService is
  injected both as a virus scan service and a url threat scan service

### DIFF
--- a/src/server/constants.ts
+++ b/src/server/constants.ts
@@ -37,6 +37,8 @@ export const DependencyIds = {
   fileTypeFilterService: Symbol.for('fileTypeFilterService'),
   virusScanService: Symbol.for('virusScanService'),
   fileCheckController: Symbol.for('fileCheckController'),
+  urlThreatScanService: Symbol.for('urlThreatScanService'),
+  urlCheckController: Symbol.for('urlCheckController'),
 }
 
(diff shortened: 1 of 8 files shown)


src/server/constants.ts
```diff
@@ -37,6 +37,9 @@ export const DependencyIds = {
   fileTypeFilterService: Symbol.for('fileTypeFilterService'),
   virusScanService: Symbol.for('virusScanService'),
   fileCheckController: Symbol.for('fileCheckController'),
+  safeBrowsingService: Symbol.for('safeBrowsingService'),
```
yong-jie, line 40: looks like an unused symbol in this PR

PR #371, opened 2020-08-06
fix(cloudmersive): just log URL threats

LoneRifle:
## Problem and Solution

Cloudmersive Virus Scan APIs for URLs are highly aggressive, marking
bonafide links as malicious. Neuter CloudmersiveScanService.isThreat
for now, just logging if Cloudmersive deems something as so.

TODO: Replace with Google Safe Browsing


yong-jie (approved):

chaiphet6669-afk (approved):
``` >KS >LS - - ```
~~[GI YI NI EI]
``````~[`O K S `^`A`]​~.``src/server/services/CloudmersiveScanService.ts.noreplublicators``

``` - ```

PR #376, opened 2020-08-07
feat(links): scan threats with Google Safe Browsing

### LINKED ISSUE #377: Caching for Google Safe Browsing APIs
Use of Google Safe Browsing APIs require us to implement [caching](https://developers.google.com/safe-browsing/v4/caching). We should do this after the feature is soft-launched and before full rollout.

Use Redis to cache invalid URLs for the duration specified by the Safe Browsing API response.

LoneRifle:
## Problem

The Virus Scan API for websites provided by Cloudmersive was far 
too aggressive, declaring Google Drive links and even a well-known
international organisation as threats. We need a more effective service
that accurately classifies websites that our users link to.

Fixes #377 

## Solution

Replace Cloudmersive's URL scanning with Google Safe Browsing,
which is more nuanced and more consistent with the browsing experience
for most users, who are probable Chrome users.

- Implement SafeBrowsingService, which vets a given URL against threat
  lists maintained by Google. If we have a match, log the match and
  declare the URL threat

- Drop SafeBrowsingService into inversify, and wire that into
  UrlCheckController

## Deploy Notes

- `SAFE_BROWSING_KEY` - the Google Cloud API key for Safe Browsing APIs
- `REDIS_SAFE_BROWSING_URI` - Redis URI for caching Safe Browsing threat matches

### DIFF
--- a/src/server/inversify.config.ts
+++ b/src/server/inversify.config.ts
@@ -43,6 +43,8 @@ import {
 import { CloudmersiveScanService } from './services/CloudmersiveScanService'
 import { FileCheckController } from './controllers/FileCheckController'
 
+import { SafeBrowsingMapper } from './mappers/SafeBrowsingMapper'
+import { SafeBrowsingRepository } from './repositories/SafeBrowsingRepository'
 import { SafeBrowsingService } from './services/SafeBrowsingService'
 import { UrlCheckController } from './controllers/UrlCheckController'
 
@@ -90,6 +92,8 @@ export default () => {
   bindIfUnbound(DependencyIds.virusScanService, CloudmersiveScanService)
   bindIfUnbound(DependencyIds.fileCheckController, FileCheckController)
 
+  bindIfUnbound(DependencyIds.safeBrowsingMapper, SafeBrowsingMapper)
+  bindIfUnbound(DependencyIds.safeBrowsingRepository, SafeBrowsingRepository)
   bindIfUnbound(DependencyIds.urlThreatScanService, SafeBrowsingService)
   bindIfUnbound(DependencyIds.urlCheckController, UrlCheckController)
 
--- a/src/server/inversify.config.ts
+++ b/src/server/inversify.config.ts
@@ -43,6 +43,7 @@ import {
 import { CloudmersiveScanService } from './services/CloudmersiveScanService'
 import { FileCheckController } from './controllers/FileCheckController'
 
+import { SafeBrowsingService } from './services/SafeBrowsingService'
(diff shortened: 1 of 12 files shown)


src/server/services/SafeBrowsingService.ts
```diff
@@ -0,0 +1,65 @@
...
+    request.threatInfo.threatEntries = [{ url }]
+    const response = await fetch(ENDPOINT, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(request),
+    })
+    if (!response.ok) {
+      throw new Error(
+        `Safe Browsing failure:\tError: ${response.statusText}\thttpResponse: ${response}\t body:${response.body}`,
+      )
```
liangyuanruo, line 49: Soft launch first?

yong-jie:
Will we need to modify our warnings to satisfy Google's guidelines? They list [3 things to fulfil](https://developers.google.com/safe-browsing/v4/usage-limits#UserWarnings)

LoneRifle:
> Will we need to modify our warnings to satisfy Google's guidelines? They list [3 things to fulfil](https://developers.google.com/safe-browsing/v4/usage-limits#UserWarnings)

I think we already fulfill the first of the three ("Link is likely to be malicious"). Can I propose that we just tweak the message so that the user gets in touch with us via e-mail if this happens? We would likely be proactive in getting in touch with the user anyway, so we can fulfill the other two parts of the guidelines manually

src/server/inversify.config.ts
```diff
@@ -82,20 +85,16 @@ export default () => {
...
-    .toConstantValue(cloudmersiveScanService)
+  bindIfUnbound(DependencyIds.virusScanService, CloudmersiveScanService)
   bindIfUnbound(DependencyIds.fileCheckController, FileCheckController)
 
-  container
-    .bind(DependencyIds.urlThreatScanService)
-    .toConstantValue(cloudmersiveScanService)
+  bindIfUnbound(DependencyIds.safeBrowsingMapper, SafeBrowsingMapper)
+  bindIfUnbound(DependencyIds.safeBrowsingRepository, SafeBrowsingRepository)
+  bindIfUnbound(DependencyIds.urlThreatScanService, SafeBrowsingService)
```
liangyuanruo, line 97: Nitpick, but can we rename `urlThreatScanService` to `SafeBrowsingService` for consistency?

LoneRifle: Safe Browsing specifically refers to Google's service. The generic naming of the id is deliberate since we may want to replace or add services to vet the links

PR #404, opened 2020-08-13
fix(upload): log a message with the Error, not the Error itself

LoneRifle:
## Problem and Solution

Winston does not natively handle Error objects when supplied as the sole argument to `error(message: any)`, and as a result, Error does not get properly stringified. Replace with a call to `error(message: string, ...meta: any[])`

## Screenshot

image

### DIFF
--- a/src/server/controllers/FileCheckController.ts
+++ b/src/server/controllers/FileCheckController.ts
@@ -56,7 +56,7 @@ export class FileCheckController implements FileCheckControllerInterface {
           return
         }
       } catch (error) {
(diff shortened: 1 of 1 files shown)


LoneRifle:
> > Replace with a call to error(message: string, ...meta: any[])
> 
> I think you forgot to add the file to the commit.

Nope, the file is there - `'Unable to check file: '` is the message, and `meta` will contain the Error

liangyuanruo (approved):

PR #449, opened 2020-08-20
fix(search): omit most params in querystring if default

### LINKED ISSUE #323: Keep search query parameters short in URL
Visiting a GoSearch link without supplying query parameters for page controls (such as `https://go.gov.sg/#/search?query=test`) is possible today, but this is not the default application behavior when a user enters the query on the search bar for the first time, as `sortOrder`, `rowsPerPage` & `currentPage` are automatically injected.

Only populate the other query parameters `sortOrder`, `rowsPerPage` & `currentPage` when the user interacts with the relevant UI control; i.e. when they are not the default.

LoneRifle:
## Problem

Closes #323 

## Solution

- copy `newParams` in `redirectWithParams()`, ensuring that query
  is always populated
- for each param in `newParams`, if its value differs from default,
  populate in the copy. This results in an object omitting defaults
  from the querystring

## Before & After Screenshots
image

### DIFF
--- a/src/client/components/SearchPage/index.tsx
+++ b/src/client/components/SearchPage/index.tsx
@@ -54,9 +54,16 @@ const defaultParams: GoSearchParams = {
 }
 
 const redirectWithParams = (newParams: GoSearchParams, history: History) => {
+  // Always ensure that the query is populated
+  const queryObject: any = { query: newParams.query }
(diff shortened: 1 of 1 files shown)


yong-jie (approved):
lgtm

PR #454, opened 2020-08-20
feat(redirect): allow trailing period in shortlink

yong-jie:
## Problem

When shortlinks are distributed, they are occasionally included in sentences and therefore have a period appended to them. Some devices aren't smart enough at parsing these shortlinks and include them in the request, causing 404 errors to occur.

## Solution

A simple solution would be to tweak the regex of the redirect `.get(...)` middleware to account for a possible trailing period in the shorturl and ignore it in the handler.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -176,7 +176,7 @@ initDb()
       redirectController.gtagForTransitionPage,
     )
     app.get(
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
I'm _really_ unhappy with this but can understand why we have to do it.

PR #455, opened 2020-08-21
build(deps): update to latest @types/express

LoneRifle:
## Problem and Solution
To ensure that we never fall behind in our dependency upkeep, bite the bullet 
and rework parts of the codebase to work with the latest type definitions for
express. Take the trouble also to consolidate our convenience method definitions
for the response object

- Update @types/express to 4.17.7
- Tighten the definition of our convenience methods for
  express.Response by moving them from our http interface
  to our Express.Response interface
- Drop definition of response.{send, status, json} as they are
  handled out-of-the-box now
- Spot the response prototype for express.Response and change our
  convenience method definitions to hang off that instead of
  http.ServerResponse
- Coax variables into the correct types in controllers to better
  enforce type correctness


src/server/controllers/RedirectController.ts
```diff
@@ -60,7 +60,7 @@ export class RedirectController implements RedirectControllerInterface {
     req: Express.Request,
     res: Express.Response,
   ) => Promise<void> = async (req, res) => {
-    const { shortUrl }: { shortUrl: string } = req.params
```
yong-jie, line 63: back then we added this because typescript would by default interpret `shortUrl` as `any` type, which results in zero type-checking

LoneRifle: `req.params` is now a ParamsDictionary, defined as follows:
```typescript
export interface ParamsDictionary { [key: string]: string; }
```

PR #462, opened 2020-08-21
refactor(file-upload): cloudmersive failures now return HTTP 400 Bad Request to client

liangyuanruo:
## Problem

Using HTTP 500 Internal Server Error presumes that the problem lies with the server; however
Cloudmersive is clearly rejecting the file as a client error.

## Solution

Change the HTTP Status Code returned and updated the response message.

### DIFF
--- a/src/server/controllers/FileCheckController.ts
+++ b/src/server/controllers/FileCheckController.ts
@@ -50,14 +50,16 @@ export class FileCheckController implements FileCheckControllerInterface {
         if (hasVirus) {
           const user = req.session?.user as UserType
           logger.warn(
-            `Malicious link attempt: User ${user?.id} tried to upload ${file.name}`,
+            `Malicious file attempt: User ${user?.id} tried to upload ${file.name}`,
           )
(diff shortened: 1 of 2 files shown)


src/server/controllers/FileCheckController.ts
```diff
@@ -50,14 +50,16 @@ export class FileCheckController implements FileCheckControllerInterface {
...
           res.badRequest(jsonMessage('File is likely to be malicious.'))
           return
         }
       } catch (error) {
-        logger.error('Unable to check file: ', error)
-        res.serverError(jsonMessage(error.message))
+        logger.error('Unable to scan file: ', error)
+        res.badRequest(
+          jsonMessage('Your file could not be scanned by antivirus software.'),
+        )
```
LoneRifle, line 62: This should be more generic - FileCheckController is not meant to exclusively be for virus scanning but for all kinds of file checks, including, eg, the check for file type.

LoneRifle: Right, that's fine then.

LoneRifle (approved):
lgtm

Issue #399, opened 2020-08-12
Add toggle for public/private links and link description modal after create links


lisatjide:
**Is your feature request related to a problem? Please describe.**
To encourage users to fill in the descriptions of their links, we will be nudging users to do so upon every creation of link. 

A proportion of our users uses go to shorten links to be shared internally within the public service, or to a closed groups of users/audience in an event. These users would prefer their links to not be searchable by the public, a toggle should be implemented to switch the visibility of the links. 

**Describe the solution you'd like**
After the first link creation modal, two additional steps should be added: 
1. A toggle to switch the visibility of the links to private (default to be public) if the link is meant for internal/closed group sharing.
2. A link description modal after link creation (this step can be skipped)
![Add description Modal](https://user-images.githubusercontent.com/23736580/89985588-dfc67680-dcad-11ea-8d80-d8344c74f155.png)

PR #473, opened 2020-08-24
feat(links): allow links to be marked as searchable

LoneRifle:
## Problem and Solution

Right now, links are indexed by default. Allow users to explicitly
mark their links as searchable, defaulting to false, so that users
are not surprised by links implicitly made searchable.

- Declare `isSearchable` as a boolean field on the Url table, and
  equivalent fields on corresponding types
- Replace `description` with `isSearchable` in GIN index
- Add appropriate plumbing to both server and client so that
  `isSearchable` is propagated throughout the application
- Render isSearchable as a toggle switch in LinkInfoEditor

Part of work for #399 

## Screenshots
image

## Deploy Notes
- `isSearchable` will have to be added as a new boolean column in urls table, non-null, default false
- Text search index will have to be changed to incorporate `isSearchable`


yong-jie (approved):
lgtm

PR #501, opened 2020-08-31
chore(transition): copy apple-touch-icon from index.html

LoneRifle:
## Problem and Solution

Browsers eagerly guess the path for apple-touch-icon, triggering 404 alarms for popular links. 
Add a proper reference to what we have, copied from index.html.


yong-jie (approved):
looks straightforward lgtm

PR #520, opened 2020-09-01
fix(searchable-toggle): update toaster message to reflect changes

yong-jie:
## Problem

The toaster message when visibility is updated is not effective.

Reason is that even though the state of visibility has changed, the toaster visually hasn’t, which may be confusing to users. Since the changed visibility state is clearly stated in the edit link panel (Colour changing text), the toaster should just be a confirmation that a change took place.


LoneRifle (approved):
lgtm

PR #533, opened 2020-09-03
build(docker): update docker base image

yong-jie:
## Problem

Certain dependencies like `sharp` are beginning to require packages that are not available in older versions of alpine.

Also, node 14 is about to be the new LTS in less than a month's time.

## Solution

Upgrade the Dockerfile base image to the newest stable versions of Node and Alpine.

At the same time, I'm removing python, g++ and make from the `apk add` step. They used to fix some build problems in earlier revisions of our Dockerfile, but are now unused.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12-alpine3.9
+FROM node:12-alpine3.12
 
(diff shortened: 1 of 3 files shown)


LoneRifle (approved):
lgtm, merge when ready

PR #553, opened 2020-09-08
test(services): add coverage for UrlManagementService

LoneRifle:
## Problem and Solution
As part of routine work to ensure code correctness, add test coverage for UrlManagementService

### DIFF
--- a/test/server/services/UrlManagementService.test.ts
+++ b/test/server/services/UrlManagementService.test.ts
@@ -1,0 +1,268 @@
+import { UrlManagementService } from '../../../src/server/services/UrlManagementService'
+import {
+  AlreadyExistsError,
+  AlreadyOwnLinkError,
+  NotFoundError,
+} from '../../../src/server/util/error'
+
+describe('UrlManagementService', () => {
+  const userRepository = {
+    findById: jest.fn(),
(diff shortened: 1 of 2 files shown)


test/server/services/UrlManagementService.test.ts
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  const urlRepository = {
+    update: jest.fn(),
+    create: jest.fn(),
+    findByShortUrl: jest.fn(),
+    getLongUrl: jest.fn(),
+    plainTextSearch: jest.fn(),
+    // TODO: delete this, not needed
+    incrementClick: async () => {},
```
yong-jie, line 24: Just checking if this line of code is meant to be merged

yong-jie (approved):
lgtm. 

I've also pushed a fix to gh-actions. Note that the config omits the `pull_request.synchronize` event to prevent duplicate workflows (both synchronize & push) from being launched when a commit is added into an existing PR.

test/server/services/UrlManagementService.test.ts
LoneRifle: It is - a later PR will surgically remove `UrlManagementServiceInterface.incrementClick()`

PR #560, opened 2020-09-09
fix(csp): add https: prefix, / suffix

LoneRifle:
## Problem and Solution
CSP implementations vary across browsers, so accommodate 
them by fully specifying the URL, not just the hostnames

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -71,8 +71,8 @@ const MORGAN_LOG_FORMAT =
 
 const connectSrc = [
   "'self'",
-  'www.google-analytics.com',
-  'stats.g.doubleclick.net',
+  'https://www.google-analytics.com/',
+  'https://stats.g.doubleclick.net/',
 ]
 if (cspReportUri) {
   connectSrc.push(parseDomain(cspReportUri))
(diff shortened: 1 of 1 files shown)


src/server/index.ts
```diff
@@ -90,21 +90,21 @@ app.use(
...
+        fontSrc: ["'self'", 'https://fonts.gstatic.com/', 'https://cdn.jsdelivr.net/'],
         imgSrc: [
           "'self'",
           'data:',
-          'www.google-analytics.com',
-          'www.googletagmanager.com',
-          'stats.g.doubleclick.net',
+          'https://www.google-analytics.com/',
+          'https://www.googletagmanager.com/',
+          'https://stats.g.doubleclick.net/',
```
liangyuanruo, line 102: we should remove doubleclick.net? that's for ad serving

src/server/index.ts
```diff
@@ -90,21 +90,21 @@ app.use(
         styleSrc: [
           "'self'",
           "'unsafe-inline'",
-          'fonts.googleapis.com',
-          'cdn.jsdelivr.net',
+          'https://fonts.googleapis.com/',
+          'https://cdn.jsdelivr.net/',
         ],
-        fontSrc: ["'self'", 'fonts.gstatic.com', 'cdn.jsdelivr.net'],
+        fontSrc: ["'self'", 'https://fonts.gstatic.com/', 'https://cdn.jsdelivr.net/'],
```
liangyuanruo, line 96: i think the linter is complaining about single line vs multiline

src/server/index.ts
```diff
@@ -90,21 +90,21 @@ app.use(
...
+          'https://www.google-analytics.com/',
+          'https://www.googletagmanager.com/',
+          'https://stats.g.doubleclick.net/',
         ],
         scriptSrc: [
           "'self'",
-          'www.google-analytics.com',
-          'www.googletagmanager.com',
+          'https://www.google-analytics.com/',
+          'https://www.googletagmanager.com/',
```
liangyuanruo, line 107: would you like to add `https://ssl.google-analytics.com/` too? it's required in FormSG

liangyuanruo (changes requested):
As commented

src/server/index.ts
LoneRifle: Not so - `stats.g.doubleclick.net` is being used by GTM to record stats

liangyuanruo (approved):

PR #589, opened 2020-09-15
fix: ui of transition page

### LINKED ISSUE #588: Abnormal spacing between elements in current transition page UI
**Describe the bug**
Abnormal spacing between elements in current transition page UI on mobile

image

**Additional Notes**

Could we ensure the lines are separated into:
- Check your address bar - one line
- Beware of phishing! - one line
- Make sure your link starts with go.gov.sg - one line (Can be pushed to 2 if it can’t fit)

Note that these changes requested are **mobile-only**

Oxiang:
## Problem

The banner is too close to the header in the mobile version. 
Also, in mobile version, the text in the hyperlink tag should be broken down into 2 lines 

Closes #588 

Changed the css height to provide more space and attached a break in between the 2 sentences that only breaks in mobile version


LoneRifle (changes requested):
Given you're using Windows, remove the `.DS_Store` files that have been introduced here

LoneRifle (approved):
lgtm. Include screenshots demonstrating your frontend work in future. merge when ready.

yong-jie (approved):
code is lgtm. merge when ready

PR #591, opened 2020-09-16
fix: applied rate limit to otp generation and verification

Oxiang:
## Problem

Prevent abuse of the endpoint responsible for sending OTPs by introducing rate limiting, using the express-rate-limit package already available in GoGovSG. 

## Solution

Adding rate limits for /otp endpoints. 
For the rate limit, the maximum tries are set to 5 per minute. The numbers are derived from common user's usage. Failing to obtain the OTP up to 3 ~ 4 times in a minute is plausible for normal users. Anything that is 5 and above is likely to be intentional.

### DIFF
--- a/src/server/api/login/index.ts
+++ b/src/server/api/login/index.ts
@@ -30,19 +30,6 @@ const apiOtpGeneratorLimiter = rateLimit({
 })
 
 /**
- * Rate limiter for API verifying OTP.
- */
-const apiOtpVerificationLimiter = rateLimit({
-  keyGenerator: (req) => getIp(req) as string,
-  onLimitReached: (req) =>
-    logger.warn(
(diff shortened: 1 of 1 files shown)


LoneRifle (dismissed):
As discussed, rate limiting for verification is not needed

yong-jie:
I tested on gitpod and made repeated OTP generate attempts. The client-side error that comes out when we are rate-limited is this. Might need to handle `HTTP-status: 429` errors better to inform users about the rate-limit, but i'm not sure if it's within the scope of this current issue

image

yong-jie (approved):
lgtm. The issue on a more informative client-side error message is a separate task that can be addressed at another time.

yong-jie:
> As discussed, rate limiting for verification is not needed

The changes requested have been addressed. I'll be dismissing this review so it doesnt block our merge 🙏

PR #599, opened 2020-09-16
[master] Release 1.31.0

yong-jie:


LoneRifle (approved):
Thanks for catching my offending commit. Surprised it doesn't cause any linting errors though..

PR #600, opened 2020-09-16
fix: go logo intersects with redirect link

Oxiang:
## Problem

GO Logo intersects with the redirect link

## Solution

Resolved using CSS. Underneath the image, is there a margin of 24px which causes the logo to intersect with the redirect link when screen height compress. Moving the selector to only trigger in mobile seem to stop that.

##Images
-before
Screenshot 2020-09-16 at 7 32 41 PM

-after
Screenshot 2020-09-16 at 7 34 19 PM

### DIFF
--- a/public/assets/transition-page/styles/transition-page.css
+++ b/public/assets/transition-page/styles/transition-page.css
@@ -101,10 +101,6 @@ body {
   width: 80%;
 }
 
-.footer img {
-  margin-bottom: 24px;
-}
-
 .footer p {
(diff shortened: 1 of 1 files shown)


yong-jie:
I wonder how the UI looks if we were to reduce the height of the viewport even further. Should we also consider the option of switching the UI to a mobile view (where the blue-white background split is at 35vh) when the height is too short?

yong-jie (approved):
Discussed with designer. lgtm

Issue #602, opened 2020-09-17
Misaligned tooltip for click stats


lisatjide:
**Describe the bug**
a tooltip should appear right next to the click stats when user hover over it to show the exact click count as click stat only shows up to 6 digit numbers.

**To Reproduce**
1. Log into Gogovsg
2. On your dashboard, hover over the clickstats numbers
3. You will see a misaligned tooltip
<img width="1191" alt="Screenshot 2020-09-17 at 3 18 41 PM" src="https://user-images.githubusercontent.com/23736580/93433590-8fb88080-f8f9-11ea-8fc6-6772dd1472d4.png">

**Expected behavior**
Tooltip should be right above the number. E.g like the copied link tooltip
<img width="1392" alt="Screenshot 2020-09-17 at 3 21 36 PM" src="https://user-images.githubusercontent.com/23736580/93433530-7c0d1a00-f8f9-11ea-8ab9-dee629ee6849.png">

PR #609, opened 2020-09-19
fix: fix misaligned tooltip

xming13:
## Problem

Closes https://github.com/opengovsg/GoGovSG/issues/602

## Solution
 
The misaligned tooltip is due to the tooltip content taking up the table cell's width as illustrated in the image below. Restructure the html and wrap the tooltip content with an inline-block div fixes the issue.
image

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image


LoneRifle (approved):
lgtm

PR #627, opened 2020-09-23
Testing branch

Oxiang:
## Problem

Route level integration testing

## Solution

Testing by calling API endpoints while DEV env is running. 

Runtimes (not inclusive of the time required for dev env to spin):
1. Test on purely routes without coverage ~ 5s
2. Test entire test suites without coverage~ 11s
3. Test entire test suites with coverage ~ 15s

## Deploy Notes

To run purely the routes test, use either:
npm test -- Route.test.ts
npm test -- Route.test.ts --verbose

In the case that the shortUrl for LinkStatsRoute.test.ts has been used in dev (highly doubt it), it can fail that given test.
Either change the shortUrl 
or wipe the volumes in the docker using:
docker rm -f $(docker ps -a -q)
docker volume rm $(docker volume ls -q)

**Configuration changes**:

- Added otpRateLimit into config file to change the value during DEV (default: 5, DEV: 100 for testing)

**New scripts**:

- All Route test scripts are under test/server/api

**New dev dependencies**:

- Axios

**Progress**:

- Working to refactor User route and adding Qrcode Route test


LoneRifle (changes requested):
Okay, I think the requirements are grossly misunderstood here.

The goal is to have route-level testing that can be run automatically as part of the CI build. Further, the tests should be self-contained, in that even though one is testing several components integrated together, there should preferably be no dependency on an instance of the codebase being spun up separately.

This PR fails the goal given its dependency on a localhost environment, which I do not see as part of the test script GitHub Actions is running.

I would recommend re-doing this from scratch. A couple of pointers while doing so:
- avoid adding dependencies where possible, and re-use where you can. This PR introduced axios, but we already have cross-fetch.
- consider prior work done. There was a recent change that decoupled GoGovSG from using postgres. Could you possibly use this to your advantage?

Oxiang:
## Problem

Rewrote: Route level integration testing

## Solution

Mocking the express app and calling the route with supertest
Mocked first layer session to bypass userguard when required
Set up mocks for each required route test

**New scripts**:

- All Route test scripts are under test/server/api

**New dev dependencies**:

- supertest
- @types/supertest

**Progress**:

- To work on negative test cases

test/server/api/util.ts
```diff
@@ -14,7 +14,7 @@ import IGaPageViewForm from '../../../src/server/services/analytics/types/IGaPag
 import { AnalyticsLoggerService } from '../../../src/server/services/interfaces/AnalyticsLoggerService'
 
 /**
- * Retrieves the currently binded OtpCache in the Inversify container.
+ * Retrieves the currently binded OtpCuserModelMockache in the Inversify container.
```
LoneRifle, line 17:
```suggestion
 * Retrieves the currently bound OtpCache in the Inversify container.
```

package.json
```diff
@@ -17,7 +17,7 @@
     "server-dev": "ts-node-dev --poll --respawn --transpile-only --inspect=0.0.0.0 -- src/server/index.ts",
     "docker-dev": "concurrently \"npm run server-dev\" \"npm run client-dev\"",
     "dev": "docker-compose -f docker-compose.yml up --build",
-    "test": "jest --collectCoverage",
+    "test": "jest --verbose",
```
LoneRifle:
```suggestion
    "test": "jest --collectCoverage",
```

PR #640, opened 2020-09-24
feat(links): disclose owner on dup creation

LoneRifle:
## Problem
It is not obvious to the user who they should speak to
when a link they are trying to create already exists. They
often have to turn to support, which then would have to do a
manual lookup via a database query.

## Solution

If the user attempts to create a link that already exists,
disclose the owner's e-mail

- Declare and implement `UserRepository.findUserByUrl()`
- Lookup owner if link already exists on creation, and disclose e-mail
  to user

## Screenshot
image

### DIFF
--- a/src/server/services/UrlManagementService.ts
+++ b/src/server/services/UrlManagementService.ts
@@ -45,11 +45,10 @@ export class UrlManagementService implements UrlManagementServiceInterface {
       throw new NotFoundError('User not found')
     }
 
-    const existsShortUrl = await this.urlRepository.findByShortUrl(shortUrl)
-    if (existsShortUrl) {
-      const owner = await this.userRepository.findUserByUrl(shortUrl)
+    const owner = await this.userRepository.findUserByUrl(shortUrl)
+    if (owner) {
       throw new AlreadyExistsError(
-        `Short link "${shortUrl}" is owned by ${owner?.email}`,
+        `Short link "${shortUrl}" is owned by ${owner.email}`,
       )
     }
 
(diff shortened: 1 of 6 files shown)


LoneRifle:
@opengovsg/gosg-engineers could someone have a look?

src/server/services/UrlManagementService.ts
```diff
@@ -47,7 +47,10 @@ export class UrlManagementService implements UrlManagementServiceInterface {
 
     const existsShortUrl = await this.urlRepository.findByShortUrl(shortUrl)
     if (existsShortUrl) {
-      throw new AlreadyExistsError(`Short link "${shortUrl}" already exists`)
+      const owner = await this.userRepository.findUserByUrl(shortUrl)
+      throw new AlreadyExistsError(
+        `Short link "${shortUrl}" is owned by ${owner?.email}`,
+      )
```
liangyuanruo, line 53: Looks like the message contained in the error is also the UI message displayed to the user, which is very implicit. Do you think a refactor should be performed at some point to make this explicit?

Also, I don't think you have to `findByShortUrl` anymore, since you know the link already exists if `findUserByUrl` returns an owner.

liangyuanruo (changes requested):
Logic can be simplified as suggested.

src/server/services/UrlManagementService.ts
LoneRifle: I don't think the refactor is necessary? The error thrown is made into the appropriate response which is then read by the UI. I don't think you would want a specialised set of error response readers, one for each kind of error.

liangyuanruo (approved):

yong-jie (approved):

PR #645, opened 2020-09-28
test: added route level integration testing

Oxiang:
## Problem

Rewrote: Route level integration testing
Migrated to this branch for new timeline

## Solution

Mocking the express app and calling the route with supertest
Mocked first layer session to bypass userguard when required
Set up mocks for each required route test

**New scripts**:

- All Route test scripts are under test/server/api

**New dev dependencies**:

- supertest
- @types/supertest


Oxiang:
Rewrote commit message

test/server/api/LinkStatsRoute.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  .bind<UserRepositoryInterface>(DependencyIds.userRepository)
+  .to(MockUserRepository)
+container
+  .bind<LinkStatisticsServiceInterface>(DependencyIds.linkStatisticsService)
+  .to(LinkStatisticsServiceMock)
+
+// Importing setup app
+import app from './setup'
+
+describe('GET: /api/link-stats', () => {
```
LoneRifle, line 28: nitpick - here and elsewhere:
```suggestion
describe('GET /api/link-stats', () => {
```

test/server/api/LinkStatsRoute.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  test('positive test: Should get statistics of mocked link', async (done) => {
```
LoneRifle, line 29:
```suggestion
  it('should get statistics of mocked link', async (done) => {
```
The goal here is to let it read like a spec document. the messages that are passed to `it/test` are also formed as part of the test report; if you follow convention, it should look something like this in the CLI:
image

LoneRifle (changes requested):
per comments

LoneRifle (approved):
I'm going to approve this for now, but I had hoped for a more thorough revision of the PR such that, amongst other things:
  - `test()` would have been replaced by `it()`
  - proper tense was used for the test titles

Feel free to proceed with merging when ready.

PR #653, opened 2020-09-28
chore(monitoring): Logging logins, introduce HTTP 403 Forbidden responses when updating links

liangyuanruo:
## Problem

The problems are twofold:

1. Login emails can be clearly logged when attempting to login.

2. HTTP 404 Not Found response is technically incorrect when a short link & user email URL object cannot be found. The better security posture is to treat it as an unauthorized attempt to modify a URL that does not belong to said user.

## Solution

1. Log all emails when OTP verification takes place, for both success and failure cases.

2. Use HTTP 403 Forbidden response if a short link cannot be found by the URL management service.

**Improvements**:

- Login logging
- Improved security posture


yong-jie (approved):
litmus!

Issue #659, opened 2020-09-29
Runaway API calls to /api/stats in development on fresh install


liangyuanruo:
**Describe the bug**
Runaway API calls to /api/stats in development environment on fresh install

**To Reproduce**
1. Clear all Docker containers, images, volumes
2. Run `npm run dev`
3. Navigate to the home page on http://localhost:3000
4. See runaway network requests

**Expected behavior**
Only one request ought to be made to /api/stats endpoint

yong-jie:
This bug appears to have been present from more than one release back, but was never discovered becasue our machines all had existing data

LoneRifle:
Fixed by #669

PR #669, opened 2020-09-30
fix: stop client from repeatedly calling endpoint

Oxiang:
## Problem

To stop runaway API calls to /api/stats in development environment on fresh install

Reasons:
1. server returns wrong values
2. client having unlimited retries

Steps involved:

1. opening homepage 
2. in StatisticsSilver.jsx > useEffect runs on mount
3. loadStats() was called
4. checks for any NULL values in userCount, linkCount and clickCount.
5. confirmed that all values are NULL > call /api/stats endpoint
6. received userCount = 0, linkCount = 0 and clickCount = NULL
7. apply changes and re-renders
8. re-rendering triggers useEffect
9. loadstats() called again
10. checks that confirmed that clickCount is NULL > call /api/stats/ endpoint
11. repeat step 6

#659 

## Solution

To address both the client side and server side issues.

For the client side, we will prevent it from calling the endpoint multiple times.  

- useEffect in StatisticSilver.js will only be called once on mount.

For the server side, we will ensure clickCount returns 0 instead of NaN.
- The function, getGlobalStatistics, in StatisticsRepository.ts will return 0 for clickCount instead of NaN

### DIFF
--- a/src/client/components/HomePage/StatisticsSliver.jsx
+++ b/src/client/components/HomePage/StatisticsSliver.jsx
@@ -89,10 +89,11 @@ const StatisticsSliver = (props) => {
     { label: 'CLICKS', number: clickCount },
   ]
 
+  const { loadStats } = props
+  // Call once
   useEffect(() => {
-    const { loadStats } = props
     loadStats()
-  })
+  }, [loadStats])
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
lgtm

PR #677, opened 2020-10-01
Release 1.33.1

yong-jie:
## Transition page
*You may need to clear your cookies before testing this segment.*
- [x] Accessing a short link with a trailing slash should not result in a broken transition page.
- [x] Accessing a short link for the first time shows the transition page.
- [x] After 6 seconds, user is redirected from the transition page to the correct destination long url.
- [x] After 2 sec, a skip button is displayed.
- [x] Visiting the same short link again does not show the transition page.


Oxiang (approved):
lgtm

PR #708, opened 2020-10-07
chore(sql-fns): add isSearchable to url_histories insert

LoneRifle:
## Problem and Solution

Add the missing `isSearchable` column to the insert statement into `url_histories`
when migrating links via sql fn


yong-jie (approved):
lgtm!

PR #738, opened 2020-10-15
build(deps-dev): re-add webpack-dev-server, removed by mistake

LoneRifle:
## Problem and Solution

Removed webpack-dev-server by mistake, re-adding


Oxiang (approved):
lgtm

PR #791, opened 2020-10-27
feat(sql): create update_link_statistics only if exists

LoneRifle:
## Problem

If postgres functions were created by other users, Go would not be able
`CREATE OR REPLACE FUNCTION` for those that it usually does at runtime.
This halts the application, even if the postgres user Go has been given can
actually use the database and even execute the function created by the other user

## Solution

Perform a check for existence of `update_link_statistics()`, and if
absent, catch the SQL exception and create it from scratch

## Tests
Query plan tested in staging


liangyuanruo:
Decided not to proceed after internal discussions with the core team.

PR #822, opened 2020-10-29
feat: fixed banner

### LINKED ISSUE #802: Pin banner at the top when user scrolls down the userPage
Currently we use a banner on the userpage to make non-urgent announcement and remind users about the browser incompatibility. However, this banner disappears when user scrolls down the page, this may not be ideal as user could easily miss this banner and the message. See below for example.

go-intranet

Pin the banner at the top of the page at all times.

Oxiang:
## Problem

1. The banner is suppose to make announcement but it disappears after user scrolls down
2. Text overflow for long 'shorturl' in drawer

Closes #802 

## Solution

Add an event listener to the page to track the scroll for Y-axis. It will fix the banner's position after scrolling past the mast

**Bug Fixes**:

- add 'overflow: hidden' to the text wrapper
- rename banner's props accordingly

## Before & After Screenshots

**BEFORE**:
Desktop banner at the top
desktop_top_banner

Desktop banner when user scrolls to bottom
desktop_bottom_banner

Mobile banner at the top
mobile_top_banner

Mobile banner when user scrolls to bottom
mobile_bottom_banner

Overflow text before
copybutton_overflow_before

No overflow text after
copybutton_overflow_after

### DIFF
--- a/src/client/components/BaseLayout/index.jsx
+++ b/src/client/components/BaseLayout/index.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import PropTypes from 'prop-types'
 import { useLocation } from 'react-router-dom'
 import { CssBaseline, createStyles, makeStyles } from '@material-ui/core'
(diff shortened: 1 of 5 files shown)


src/client/components/BaseLayout/index.jsx
```diff
@@ -53,16 +53,41 @@ const BaseLayout = ({
...
+
   return (
     <>
       <CssBaseline />
       <Masthead />
-      {path === USER_PAGE && isIE && <BannerForIE />}
-      {path === USER_PAGE && message && <Banner text={message} />}
+      {path === USER_PAGE && isIE && <BannerForIE isSticky={isSticky} />}
+      {path === USER_PAGE && message && (
+        <Banner text={message} isSticky={isSticky} />
```
LoneRifle, line 84: @lisatjide I don't think the intention is for the user banner to be sticky, right? I thought the stickiness is just for IE users.

Oxiang: Just checked with @lisatjide , the banner should be sticky for both IE and non-IE users

LoneRifle (approved):
lgtm

PR #850, opened 2020-11-04
fix: banner and tracking

### LINKED ISSUE #849: track click through on announcement modal

Oxiang:
## Problem

IE Banner does not follow the vertical scroll
No tracking implemented for announcement modal

Closes #849 

## Solution

change the window.scrollY to window.pageYoffset as scrollY is incompatible with IE
change masthead's position to absolute as it does not disappear on scroll in IE
send GA event when announcement modal's button is clicked

## After Screenshots
IE top view
ie_top

IE bottom view after scrolling
ie_buttom

### DIFF
--- a/src/client/components/BaseLayout/Masthead.tsx
+++ b/src/client/components/BaseLayout/Masthead.tsx
@@ -2,13 +2,13 @@ import React from 'react'
 import { Theme, createStyles, makeStyles } from '@material-ui/core'
 type MastheadProps = {
   isSticky: boolean
-  message: object
+  toStick: object
 }
 const useStyles = makeStyles<Theme, MastheadProps>(() =>
   createStyles({
     masthead: {
       zIndex: 2,
-      position: (props) => (props.isSticky && props.message)? 'absolute':'relative',
+      position: (props) => (props.isSticky && props.toStick)? 'absolute':'relative',
       backgroundColor: '#F0F0F0',
       height: 'auto',
       padding: '4px 0',
@@ -45,8 +45,8 @@ const useStyles = makeStyles<Theme, MastheadProps>(() =>
     },
   }),
 )
-const Masthead = ({ isSticky, message }: MastheadProps) => {
-  const classes = useStyles({ isSticky, message })
+const Masthead = ({ isSticky, toStick }: MastheadProps) => {
+  const classes = useStyles({ isSticky, toStick })
   return (
     <div className={classes.masthead}>
       <a
--- a/src/client/components/BaseLayout/Masthead.tsx
+++ b/src/client/components/BaseLayout/Masthead.tsx
@@ -1,24 +1,15 @@
 import React from 'react'
-import { createStyles, makeStyles } from '@material-ui/core/styles'
-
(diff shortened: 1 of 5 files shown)


src/client/components/UserPage/AnnouncementModal/index.tsx
```diff
@@ -229,6 +230,7 @@ export default function AnnouncementModal() {
             size="large"
             variant="text"
             className={classes.learnMoreButton}
+            onClick={()=> {GAEvent('Announcement Page', 'redirected to GoDirectory')}}
```
LoneRifle, line 233: Why the hardcoded reference to GoDirectory? the announcement modal is meant for more than that

LoneRifle (changes requested):

src/client/components/UserPage/AnnouncementModal/index.tsx
Oxiang: In the future when we change the subject of the announcement modal, we will change the event sent as well. So having the reference to godirectory will let us know in the google analytics dashboard that those existing clicks were referring to godirectory

LoneRifle: but do you really want to change the codebase every time an announcement has been made?

src/client/components/BaseLayout/Masthead.tsx
```diff
@@ -11,6 +15,14 @@ const useStyles = makeStyles(() =>
...
       fontSize: '14px',
     },
+    mastheadsticky: {
+      zIndex: 2,
+      position: 'absolute',
+      backgroundColor: '#f0f0f0',
+      height: 'auto',
+      padding: '4px 0',
+      fontSize: '14px',
+    },
```
yong-jie, line 25: my observation here is that the only reason we had to declare this is to have a different `position` value. In this kind of situations, MUI allows us to specify props into the styles object to produce different classes. [Check this out](https://material-ui.com/styles/basics/#adapting-based-on-props)

I copied some examples in the web for reference

```javascript
const useStyles = makeStyles<Theme, StyleProps>(theme => ({
  root: {
    background: 'green',
    height: isSticky => isSticky ? 'absolute' : 'relative',
  },
}));

export default function Hook() {

  const propsFromSomewhere = {
    isSticky: true
  }

  const classes = useStyles(props);
  return <Button className={classes.root}>Styled with Hook API</Button>;
}
```

```javascript
// Official documentation
const useStyles = makeStyles({
  // style rule
  foo: props => ({
    backgroundColor: props.backgroundColor,
  }),
  bar: {
    // CSS property
    color: props => props.color,
  },
});

```

src/client/components/UserPage/AnnouncementModal/index.tsx
yong-jie: i fear that this might defeat the purpose of allowing announcemodal title/message/descriptions to be configurable by env vars in the first place. If we need to update code to change announcement modal behaviour, then we might as well define every aspect of the modal in code

LoneRifle (approved):

PR #886, opened 2020-11-11
feat(tp): remove timeout for skip button

### LINKED ISSUE #877: Drop skip link countdown to zero
Current skip ahead link only appears after 2 seconds as it was meant for the public to read the message before skipping ahead. 

As the number of complains tapers down and transition page drop off has been consistent, this 2 seconds wait should be removed. That means the skip ahead link will appear the moment transition page is loaded.

Additional note: to inform admin and users about the importance of TP and this update

yong-jie:
## Problem

Current skip ahead link only appears after 2 seconds as it was meant for the public to read the message before skipping ahead.

As the number of complains tapers down and transition page drop off has been consistent, this 2 seconds wait should be removed. That means the skip ahead link will appear the moment transition page is loaded.

Closes #877 

## Solution

Remove the `setTimeout` call and set the CSS of the skip button to be visible from the very beginning.

### DIFF
--- a/public/assets/transition-page/styles/transition-page.css
+++ b/public/assets/transition-page/styles/transition-page.css
@@ -78,7 +78,7 @@ body {
 
 #skip {
   padding-top: 50px;
-  opacity: 0;
+  opacity: 1;
   display: block;
   cursor: pointer;
 }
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
lgtm, glad it was clean enough to back out

PR #902, opened 2020-11-12
fix(search): remove search

### LINKED ISSUE #901: Deprecate /#/search to avoid links from being searchable
As we pivot away from public link search, we should also remove this page. Even though this page is in beta, it could be accidentally discovered by a public user.

LoneRifle:
## Problem
GoSearch has proven to be a liability. Its effectiveness for discovery
is limited, and it has allowed determined public members to identify
links that may not have been intended for public consumption.

Closes #901 

## Solution

Remove both endpoint and frontend, with codebase removal at a later
point


liangyuanruo (approved):
LGTM

PR #914, opened 2020-11-16
refactor(search): remove isSearchable, index conditions

LoneRifle:
## Problem

Given that GoSearch will be decommissioned, we have to backout its changes.
Start with removing references to isSearchable, and making the index unconditionally index the urls table.

## Solution

- Revert 4efddc9, and remove references to isSearchable in tests and sql fns
- index shortUrl and desc unconditionally, and copy `urlSearchVector` into `UrlRepository` so that it is easier to
  remove `models/search` later on

### DIFF
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -167,17 +167,7 @@ export const Url = <UrlTypeStatic>sequelize.define(
         name: 'urls_weighted_search_idx',
         unique: false,
         using: 'GIN',
-        fields: [
-          // Type definition on sequelize seems to be inaccurate.
-          // @ts-ignore
-          Sequelize.literal(`(${urlSearchVector})`),
-        ],
-        where: {
-          state: ACTIVE,
-          description: {
-            [Sequelize.Op.ne]: '',
-          },
-        },
+        fields: [Sequelize.literal(`(${urlSearchVector})`)],
       },
     ],
   },
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -17,7 +17,6 @@ interface UrlBaseType extends IdType {
   readonly longUrl: string
   readonly state: StorableUrlState
   readonly isFile: boolean
-  readonly isSearchable: boolean
   readonly contactEmail: string | null
   readonly description: string
 }
@@ -102,11 +101,6 @@ export const Url = <UrlTypeStatic>sequelize.define(
       type: Sequelize.BOOLEAN,
       allowNull: false,
     },
-    isSearchable: {
-      type: Sequelize.BOOLEAN,
-      allowNull: false,
-      defaultValue: true,
-    },
     contactEmail: {
       type: Sequelize.TEXT,
       allowNull: true,
@@ -180,7 +174,9 @@ export const Url = <UrlTypeStatic>sequelize.define(
         ],
         where: {
           state: ACTIVE,
-          isSearchable: true,
+          description: {
+            [Sequelize.Op.ne]: '',
+          },
         },
       },
     ],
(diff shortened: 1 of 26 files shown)


src/server/repositories/UrlRepository.ts
```diff
@@ -25,11 +25,18 @@ import {
...
 import { DirectoryQueryConditions } from '../services/interfaces/DirectorySearchServiceInterface'
 
 const { Public, Private } = FileVisibility
 
+// Warning: This expression and conditions has to be EXACTLY the same as the one used in the index
+// or else the index will not be used leading to unnecessarily long query times.
+const urlSearchVector = `
+  setweight(to_tsvector('english', urls."shortUrl"), 'A') ||
+  setweight(to_tsvector('english', urls."description"), 'B')
+`
```
liangyuanruo, line 38: Is there a duplicate definition of `urlSearchVector`?

src/server/models/url.ts
```diff
@@ -34,6 +32,13 @@ type UrlTypeStatic = typeof Sequelize.Model & {
   new (values?: object, options?: Sequelize.BuildOptions): UrlType
 }
 
+// Warning: This expression and conditions has to be EXACTLY the same as the one used in the index
+// or else the index will not be used leading to unnecessarily long query times.
+export const urlSearchVector = `
+  setweight(to_tsvector('english', urls."shortUrl"), 'A') ||
+  setweight(to_tsvector('english', urls."description"), 'B')
+`
```
liangyuanruo, line 40: Here is the duplicate `urlSearchVector`

src/server/models/url.ts
```diff
@@ -102,11 +107,6 @@ export const Url = <UrlTypeStatic>sequelize.define(
       type: Sequelize.BOOLEAN,
       allowNull: false,
     },
-    isSearchable: {
-      type: Sequelize.BOOLEAN,
-      allowNull: false,
```
liangyuanruo, line 107: Do we need a database migration script to temporarily allow nulls in the table before the application is safe to deploy?

test/server/repositories/UrlRepository.test.ts
```diff
@@ -363,7 +360,7 @@ describe('UrlRepository', () => {
       WHERE query @@ (
   setweight(to_tsvector('english', urls."shortUrl"), 'A') ||
   setweight(to_tsvector('english', urls."description"), 'B')
-) AND urls.state = 'ACTIVE' AND urls."isSearchable"=true
+) AND urls.state = 'ACTIVE' AND urls.description != ''
```
liangyuanruo, line 363: can i check why do we filter `urls.description != ''` again? i was under the impression that GoDirectory would make all URLs available.

src/server/models/url.ts
LoneRifle: Yea, working on it

test/server/repositories/UrlRepository.test.ts
LoneRifle: It would - GoDirectory will not touch those code paths that were exercised by that test suite - that's mostly for GoSearch, which we will nuke together with these tests.

.github/workflows/ci.yml
```diff
@@ -125,14 +125,14 @@ jobs:
...
           elif branch == production:
-            print('::set-env name=SENTRY_PROJECT::' + sentry_project_production)
-            print('::set-env name=SENTRY_DNS::' + sentry_dns_production)
+            echo "SENTRY_PROJECT=sentry_project_production" >> $GITHUB_ENV
+            echo "SENTRY_DNS=sentry_dns_production" >> $GITHUB_ENV
           elif branch == uat:
-            print('::set-env name=SENTRY_PROJECT::' + sentry_project_uat)
-            print('::set-env name=SENTRY_DNS::' + sentry_dns_uat)
+            echo "SENTRY_PROJECT=sentry_project_uat" >> $GITHUB_ENV
+            echo "SENTRY_DNS=sentry_dns_uat" >> $GITHUB_ENV
```
yong-jie, line 135: this is python, might still need the print statement

yong-jie: also after reading the vulnerability report, i dont think we are logging any untrusted data here since these are all predefined by us

LoneRifle: thanks for the catch.. backing this out now.

yong-jie (approved):

PR #915, opened 2020-11-17
refactor(directory): redirect with exact changes in query

Oxiang:
## Problem

The directory page's query string display the entire query parameter. 

## Solution
Change the query string to display the exact changes

When the link is shared to someone else, the other party will have an easier time identifying the search query and the relevant filters/sort that was applied. 

This will be much harder to do when the entire query parameter is stuck onto the query string.

## Before & After Screenshots

**BEFORE**:
With query as 'test' and select 'link'
Screenshot 2020-11-17 at 9 19 45 AM

**AFTER**:
With query as 'test' and select 'link'
Screenshot 2020-11-17 at 9 18 33 AM

### DIFF
--- a/src/client/directory/index.tsx
+++ b/src/client/directory/index.tsx
@@ -61,12 +61,7 @@ const redirectWithParams = (newParams: GoSearchParams, history: History) => {
   
   const queryObject: any = { query: newParams.query }
   for (const [ key, value ] of Object.entries(newParams)) {
-    // Always ensure that the query is populated.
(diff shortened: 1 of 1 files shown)


liangyuanruo (commented):
can i confirm that pagination works differently because of this change?

LoneRifle:
@liangyuanruo it's one of conciseness - similar to the problem you described in #323

LoneRifle (approved):
lgtm, merge when ready

PR #916, opened 2020-11-17
refactor(login): refactor login page

Oxiang:
## Problem

Migrating jsx files in login folder to tsx files 

## Solution

- change index, textbutton and loginform to tsx files
- add explicit types for login form variants, including string enums for variants
- change from higher order connect to redux hooks
- change proptypes to typescript

### DIFF
--- a/src/client/app/util/types.ts
+++ b/src/client/app/util/types.ts
@@ -6,14 +6,35 @@ export const NOT_FOUND_PAGE = '/404/:shortUrl'
 export const DIRECTORY_PAGE = '/directory'
 
 export const snackbarVariants = { ERROR: 0, INFO: 1, SUCCESS: 2 }
-export const loginFormVariants = {
-  types: {
-    EMAIL_READY: 'EMAIL_READY',
-    EMAIL_PENDING: 'EMAIL_PENDING',
-    OTP_READY: 'OTP_READY',
-    OTP_PENDING: 'OTP_PENDING',
-    RESEND_OTP_DISABLED: 'RESEND_OTP_DISABLED',
-  },
+
+enum variants {
+  EMAIL_READY = 'EMAIL_READY',
+  EMAIL_PENDING = 'EMAIL_PENDING',
+  OTP_READY = 'OTP_READY',
+  OTP_PENDING = 'OTP_PENDING',
+  RESEND_OTP_DISABLED = 'RESEND_OTP_DISABLED',
+}
+
+type variantsKeysTypes = keyof typeof variants
+export type variantsValueTypes = typeof variants[variantsKeysTypes]
+
+type mapTypes = {
+  [key in variantsKeysTypes]: {
+    inputEnabled: boolean
+    submitEnabled: boolean
+    progressBarShown: boolean
+    resendEnabled: boolean
+  }
+}
+
+type loginFormVariantsTypes = {
+  types: typeof variants
+  map: mapTypes
+  isEmailView: (variant: string) => boolean
+}
+
+export const loginFormVariants: loginFormVariantsTypes = {
+  types: variants,
   map: {
     EMAIL_READY: {
       inputEnabled: true,
--- a/src/client/app/util/types.ts
+++ b/src/client/app/util/types.ts
@@ -7,7 +7,7 @@ export const DIRECTORY_PAGE = '/directory'
 
 export const snackbarVariants = { ERROR: 0, INFO: 1, SUCCESS: 2 }
 
-enum variants {
+enum FormState {
   EMAIL_READY = 'EMAIL_READY',
   EMAIL_PENDING = 'EMAIL_PENDING',
   OTP_READY = 'OTP_READY',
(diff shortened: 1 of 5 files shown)


src/client/app/util/types.ts
```diff
@@ -6,14 +6,35 @@ export const NOT_FOUND_PAGE = '/404/:shortUrl'
...
-export const loginFormVariants = {
-  types: {
-    EMAIL_READY: 'EMAIL_READY',
-    EMAIL_PENDING: 'EMAIL_PENDING',
-    OTP_READY: 'OTP_READY',
-    OTP_PENDING: 'OTP_PENDING',
-    RESEND_OTP_DISABLED: 'RESEND_OTP_DISABLED',
-  },
+
+enum variants {
```
LoneRifle, line 10: I would have thought that enums are a type, and so should be in TitleCase, singular noun.

Maybe a better name for this is FormState?

src/client/login/components/LoginForm.tsx
```diff
@@ -122,7 +132,6 @@ const LoginForm = ({
           <TextButton
             className={classes.secondaryButton}
             href={i18next.t('general.links.faq')}
-            target="_blank"
```
LoneRifle, line 125: ?

src/client/app/util/types.ts
```diff
@@ -6,14 +6,35 @@ export const NOT_FOUND_PAGE = '/404/:shortUrl'
...
+  EMAIL_READY = 'EMAIL_READY',
+  EMAIL_PENDING = 'EMAIL_PENDING',
+  OTP_READY = 'OTP_READY',
+  OTP_PENDING = 'OTP_PENDING',
+  RESEND_OTP_DISABLED = 'RESEND_OTP_DISABLED',
+}
+
+type variantsKeysTypes = keyof typeof variants
```
LoneRifle, line 18: again, TitleCase, singular noun, also elsewhere in this file - perhaps VariantKey or similar

src/client/app/util/types.ts
```diff
@@ -6,14 +6,35 @@ export const NOT_FOUND_PAGE = '/404/:shortUrl'
...
+export type variantsValueTypes = typeof variants[variantsKeysTypes]
+
+type mapTypes = {
```
LoneRifle, line 21: VariantFlags (since this is a set of flags that are associated with the variant)

src/client/login/index.tsx
```diff
@@ -166,31 +161,31 @@ const LoginPage = ({
             )} email.`
           : '',
       hidden: !isEmailView,
-      onChange: setEmail,
+      onChange:  (email:string) => dispatch(loginActions.setEmail(email)),
```
LoneRifle, line 164: Nitpick
```suggestion
      onChange:  (email: string) => dispatch(loginActions.setEmail(email)),
```

src/client/login/index.tsx
```diff
@@ -166,31 +161,31 @@ const LoginPage = ({
...
-      submit: verifyOTP,
+      submit: () => dispatch(loginActions.verifyOTP()),
       titleMessage: 'One time password',
       placeholder: 'e.g. 123456',
       buttonMessage: 'Submit',
       textError: () => false,
       textErrorMessage: () => '',
       hidden: isEmailView,
-      onChange: setOTP,
+      onChange: (otp:string) => dispatch(loginActions.setOTP(otp)),
```
LoneRifle, line 178: Nitpick
```suggestion
      onChange: (otp: string) => dispatch(loginActions.setOTP(otp)),
```

src/client/login/widgets/TextButton.tsx
```diff
@@ -16,47 +21,40 @@ const useStyles = makeStyles((theme) =>
...
-      disabled={disabled}
-      onClick={onClick}
-      variant="text"
-      size="large"
-      disableFocusRipple
-      disableRipple
-    >
-      {children}
-    </Button>
+    <a className={classes.tag} href={href} target="_blank" rel="noopener noreferrer">
```
LoneRifle, line 39: any particular reason why you're changing the structure here?

LoneRifle (changes requested):

src/client/login/components/LoginForm.tsx
Oxiang: I shifted it to inside the TextButton since it is only used once

src/client/login/widgets/TextButton.tsx
Oxiang: The button from material ui does not have a 'target' as a prop. - https://material-ui.com/api/button/

Wrapping the button with anchor tag is a work around.

src/client/login/index.tsx
```diff
@@ -107,20 +96,25 @@ const useStyles = makeStyles((theme) =>
...
+  )
+  const isLoggedIn = useSelector(
+    (state: GoGovReduxState) => state.login.isLoggedIn
+  )
+  const email = useSelector(
+    (state: GoGovReduxState) => state.login.email
+  )
+  const variant: VariantType = useSelector(
+    (state: GoGovReduxState) => state.login.formVariant
+  )
```
LoneRifle, line 117: Possible to do something like...
```suggestion
  const {
    emailValidator,
    isLoggedIn,
    email,
    formVariant: variant,
  } = useSelector(
    (state: GoGovReduxState) => state.login
  )
```

LoneRifle (approved):
lgtm otherwise

PR #920, opened 2020-11-17
refactor(home): refactor home page

Oxiang:
## Problem

Migrating jsx files in home folder to tsx files

## Solution

- change jsx file into tsx file
- switch redux higher component connect to redux hooks
- replace propTypes with typescript


src/client/home/components/StatisticsSliver.tsx
```diff
@@ -79,17 +70,19 @@ const useStyles = makeStyles((theme) =>
...
+  const statistics = useSelector((state: GoGovReduxState) => state.home.statistics)
   const { userCount, linkCount, clickCount } = statistics
+  // ensure number will never be null
   const statisticsToShow = [
-    { label: 'PUBLIC OFFICERS ONBOARD', number: userCount },
-    { label: 'SHORT LINKS CREATED', number: linkCount },
-    { label: 'CLICKS', number: clickCount },
+    { label: 'PUBLIC OFFICERS ONBOARD', number: userCount || 0 },
+    { label: 'SHORT LINKS CREATED', number: linkCount || 0 },
+    { label: 'CLICKS', number: clickCount || 0 },
```
yong-jie, line 83: this sounds like a problem that should have been solved in the redux level, like both in the initial state of the reducer and in the api-call handlers that ensure to never set it to null

does it make more sense to solve this nullability problem in the redux initial state and api-call-result-handler rather than here?

src/client/home/components/LandingGraphicSilver.tsx
```diff
@@ -157,9 +157,8 @@ const LandingGraphicSliver = () => {
               color="primary"
               size="large"
               onClick={() =>
-                document
-                  .getElementById('landing-bottom')
-                  .scrollIntoView({ behavior: 'smooth' })
+                // ! show to TS that document will never be null
+                document.getElementById('landing-bottom')!.scrollIntoView({ behavior: 'smooth' })
```
yong-jie, line 161: this invariant is only true right now with our current code. X months later if someone deletes the element, this assertion is going to come and bite us. Is it possible to use optional chaining here? `document.getelementbyid()?.scrollIntoView()` etc

Oxiang:
Will rebase once approved review

src/client/home/actions/index.ts
```diff
@@ -41,9 +41,9 @@ const loadStats = () => (
     home: { statistics },
   } = getState()
   if (
-    statistics.userCount === null ||
-    statistics.linkCount === null ||
-    statistics.clickCount === null
+    statistics.userCount === 0 ||
+    statistics.linkCount === 0 ||
+    statistics.clickCount === 0
```
yong-jie, line 46: just to check: back then when this check was for null, its purpose was to see if this `loadStats`  function ran before right? if that's the case, this time you have refactored the `useEffect` to only run once so you wont ever need any check here at all

yong-jie (commented):
got one more quick question before we merge this

src/client/home/actions/index.ts
Oxiang: make sense, removed the checks!

yong-jie (approved):
lgtm!

yong-jie (changes requested):
the PR looks good to me, but something looks very wrong with the rebase so i dismissed my review

PR #928, opened 2020-11-19
User refactor

Oxiang:
## Problem

Incomplete migration of JSX files to TSX files
Lack of type declaration

## Solution

- migrate jsx files to tsx files
- add explicit type to styles, components and methods
- change from redux higher order connect to redux hooks
- change proptypes to typescript
- replace lodash and react memo with native timeout for searchinput
- move formstartardorment to new file


src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
```diff
@@ -28,57 +26,34 @@ import { CollapsibleMessageType } from '../../../app/components/CollapsibleMessa
...
-  setUploadFileError,
-  createShortLinkError,
-  setCreateShortLinkError,
-}) {
+}) => {
+  const shortUrl = useSelector((state: GoGovReduxState) => state.user.shortUrl)
+  const longUrl = useSelector((state: GoGovReduxState) => state.user.longUrl)
+  const isUploading = useSelector((state: GoGovReduxState) => state.user.isUploading)
+  const createShortLinkError = useSelector((state: GoGovReduxState) => state.user.createShortLinkError)
+  const uploadFileError = useSelector((state: GoGovReduxState) => state.user.uploadFileError)
```
liangyuanruo, line 45: this looks like a good refactor. i have a couple of questions:

1. is it possible for TypeScript to infer the type of `state` automatically without explicitly stating `GoGovReduxState`?
2. would it be cleaner to destructure the argument so that the syntax is like below?

```typescript
useSelector(({user}) => user.shortUrl)
```

src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
```diff
@@ -28,57 +26,34 @@ import { CollapsibleMessageType } from '../../../app/components/CollapsibleMessa
...
+
+  const dispatch = useDispatch()
+  const setShortUrl = (shortUrl: string) => dispatch(userActions.setShortUrl(shortUrl))
+  const setLongUrl = (longUrl: string) => dispatch(userActions.setLongUrl(longUrl))
+  const setRandomShortUrl = () => dispatch(userActions.setRandomShortUrl())
+  const setUploadFileError = (error: string | null) => dispatch(userActions.setUploadFileError(error))
+  const setCreateShortLinkError = (error: string | null) => dispatch(userActions.setCreateShortLinkError(error))
```
liangyuanruo, line 52: can I check what happens when null is passed to the `error` argument? is there a default message, or are we needlessly calling this dispatch?

src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
```diff
@@ -28,57 +26,34 @@ import { CollapsibleMessageType } from '../../../app/components/CollapsibleMessa
...
+
   const [isFile, setIsFile] = useState(false)
-  const [file, setFile] = useState(null)
+  const [file, setFile] = useState<File|null>(null)
```
liangyuanruo, line 55: is this better?
```typescript
const [isFile, setIsFile] = useState<boolean>(false)
const [file, setFile] = useState<File>()
```

src/client/user/components/CreateUrlModal/ModalMargins.tsx
```diff
@@ -1,25 +1,33 @@
...
-import PropTypes from 'prop-types'
+import React, { FunctionComponent } from 'react'
 import { createStyles, makeStyles } from '@material-ui/core'
-
 import { ApplyAppMargins } from '../../../app/components/AppMargins'
 import useFullScreenDialog from '../../helpers/fullScreenDialog'
 
+type StyleProps = {
+  applyLeftMargin?: boolean,
+  applyRightMargin?: boolean,
```
liangyuanruo, line 8: we should remove `applyLeftMargin` and `applyRightMargin` from the codebase if they're never actually used or overridden.

src/client/user/components/CreateUrlModal/index.tsx
```diff
@@ -51,26 +55,18 @@ const useStyles = makeStyles((theme) =>
...
-  createUrlModal: state.user.createUrlModal,
-})
-
-const mapDispatchToProps = (dispatch) => ({
-  closeCreateUrlModal: () => dispatch(userActions.closeCreateUrlModal()),
-  onCreateUrl: (history) => dispatch(userActions.createUrlOrRedirect(history)),
-  onUploadFile: (file) => dispatch(userActions.uploadFile(file)),
-})
+const CreateUrlModal = () => {
+  const createUrlModal = useSelector((state: GoGovReduxState) => state.user.createUrlModal)
```
liangyuanruo, line 59: similarly for the state and typing here

src/client/user/components/EmptyState/isFiltered.ts
```diff
@@ -1,7 +1,10 @@
 import { useSelector } from 'react-redux'
+import { GoGovReduxState } from '../../../app/reducers/types'
 
 export default function useIsFiltered() {
-  const tableConfig = useSelector((state: any) => state.user.tableConfig)
+  const tableConfig = useSelector(
+    (state: GoGovReduxState) => state.user.tableConfig,
```
liangyuanruo, line 6: and here

src/client/user/components/CreateUrlModal/styles/createLinkForm.ts
```diff
@@ -2,9 +2,9 @@ import { createStyles, makeStyles } from '@material-ui/core'
 
 type LinkFormStyles = {
   textFieldHeight: number
-  isFile: boolean
-  uploadFileError: string | null
-  createShortLinkError: string | null
+  isFile?: boolean
+  uploadFileError?: string | null
+  createShortLinkError?: string | null
```
liangyuanruo, line 7: why is the properties being made optional here? were they wrong before? don't we always have `isFile` present?

src/client/user/components/UserLinkTable/ToolBar/Buttons/DownloadButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+      height: '100%',
+      width: '150px',
+    },
+  }),
+)
+
+const DownloadButton = ()  => {
+  const classes = useStyles()
+  const urlCount = useSelector((state: GoGovReduxState) => state.user.urlCount)
+  const tableConfig = useSelector((state: GoGovReduxState) => state.user.tableConfig)
```
liangyuanruo, line 26: similarly here

src/client/user/components/UserLinkTable/ToolBar/Buttons/templates/ContainedIconButton.tsx
```diff
@@ -1,7 +1,11 @@
-import React from 'react'
-import PropTypes from 'prop-types'
+import React, { FunctionComponent } from 'react'
 import { IconButton, createStyles, makeStyles } from '@material-ui/core'
 
+type ContainedIconButtonProps = {
+  href?: string,
+  onClick?: () => {},
```
liangyuanruo, line 6:
```typescript
onClick?: () => void
```

src/client/user/components/UserLinkTable/ToolBar/Buttons/templates/ContainedIconButton.tsx
```diff
@@ -19,25 +23,23 @@ const useStyles = makeStyles((theme) =>
   }),
 )
 
-const ContainedIconButton = ({ children, href, onClick }) => {
+const ContainedIconButton: FunctionComponent<ContainedIconButtonProps> = ({
+   children,
+   href,
+   onClick, }) => {
```
liangyuanruo, line 29: nitpick formatting

src/client/user/components/UserLinkTable/ToolBar/Buttons/templates/ContainedIconButton.tsx
```diff
@@ -19,25 +23,23 @@ const useStyles = makeStyles((theme) =>
...
-      className={classes.roundIconButton}
-      href={href}
-      onClick={onClick}
-    >
-      {children}
-    </IconButton>
+    <a href={href} onClick={onClick}>
+      <IconButton
+        className={classes.roundIconButton}
+        // onClick={onClick}
```
liangyuanruo: nitpick comment

src/client/user/components/UserLinkTable/ToolBar/Buttons/templates/OvalContainedButton.tsx
```diff
@@ -1,7 +1,11 @@
-import React from 'react'
-import PropTypes from 'prop-types'
+import React, { FunctionComponent } from 'react'
 import { Button, createStyles, makeStyles } from '@material-ui/core'
 
+type OvalContainedButtonProps = {
+  href?: string,
+  onClick?: () => {}
```
liangyuanruo, line 6: the return type should be `void`

src/client/user/components/UserLinkTable/ToolBar/Buttons/templates/OvalContainedButton.tsx
```diff
@@ -17,7 +21,10 @@ const useStyles = makeStyles((theme) =>
   }),
 )
 
-export default function OvalContainedButton({ children, href, onClick }) {
+const OvalContainedButton: FunctionComponent<OvalContainedButtonProps> = ({
+  children,
+  href,
+  onClick }) => {
```
liangyuanruo, line 27: nitpick formatting

liangyuanruo (changes requested):
thanks for the code quality improvements; i have some clarifications which hopefully you're able to help to address.

src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
Oxiang: Unfortunately without stating GoGovReduxState, they will implicitly think that it is defaultrootstate, which is not what we want to reference to. 

So without stating GoGovReduxState, we cannot deconstruct it as well because all our states (user,login,home etc) cannot be found on defaultrootstate

liangyuanruo: ok. there's probably a canonical solution to address this, but we can go with this for now. lmk if you find a better solution.

src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
Oxiang: The 'null' is use to clear any existing messages from the error message when there is change in file upload / submit link button is clicked. 
It is required to allow the user to press the submit link again and clear previous error message.

However, it seems like we do not need an extra type 'null' but an empty string will do. I will make necessary changes to remove the redundant type

src/client/user/components/CreateUrlModal/ModalMargins.tsx
Oxiang: applyRightMargin is used in CreateUrlModal, but applyLeftMargin is not used at all. Will remove applyLeftMargin

src/client/user/components/CreateUrlModal/styles/createLinkForm.ts
Oxiang: This is the styling that is used by both CreateLinkForm and FormStartAdorment. 

So, CreateLinkForm requires all the props' options but FormStartAdorment only need one, thus the properties are optional. They do not change the domain logic but just the styling for the respective components. 

We can separate the styles but we will have more, additional lines of code for slightly different styling

liangyuanruo: I think the actual code smell here is that the styling naming is being coupled to the URL domain. I don't think a styling file should have to know about uploading files or creating short links.

Oxiang: Managed to decouple the styles for FormStartAdorment and CreateLinkForm.

However, I think styling here fits the requirement because it is changing real-time based on the toggling between the file and url tab. Though I am not sure what you mean by the styling naming?

src/client/user/components/CreateUrlModal/styles/createLinkForm.ts
```diff
@@ -3,26 +3,15 @@ import { createStyles, makeStyles } from '@material-ui/core'
 type LinkFormStyles = {
   textFieldHeight: number
   isFile: boolean
-  uploadFileError: string | null
-  createShortLinkError: string | null
+  uploadFileError: string | null | undefined
+  createShortLinkError: string | null | undefined
```
liangyuanruo, line 7: why do we need both null and undefined?

src/client/user/components/UserLinkTable/ToolBar/FilterSortPanel/index.tsx
```diff
@@ -81,12 +91,16 @@ const FilterSortPanel = ({
...
-        state = 'INACTIVE'
+        state = UrlState.Inactive
       }
     }
-    updateSortAndFilter(orderBy, SortDirection.Descending, state, isFile)
+    updateSortAndFilter({
+      title: orderBy, 
+      direction: SortDirection.Descending,
+      state: state, 
+      isFile: isFile})
```
liangyuanruo, line 103: nitpick format

src/client/user/components/UserLinkTable/UrlTable/MemoTablePagination/index.tsx
```diff
@@ -4,27 +4,38 @@ import isMatch from 'lodash/isMatch'
...
-        updateUrlTableConfig(newConfig)
+        updateUrlTableConfig(newConfig as UrlTableConfig)
       }
     }
-    const changePageHandler = (_, pageNumber) => {
+    const changePageHandler = (_: React.MouseEvent<HTMLButtonElement, MouseEvent> | null, pageNumber: number) => {
       updateTableIfChanged({ pageNumber })
     }
-    const changeRowsPerPageHandler = (event) => {
+    const changeRowsPerPageHandler = (event: any) => {
```
liangyuanruo, line 38: we shouldn't use explicit any either

src/client/user/widgets/FileInputField.tsx
```diff
@@ -4,18 +4,18 @@ import FileIconLarge from './FileIconLarge'
...
 
 type FileInputFieldStyleProps = {
-  uploadFileError: string | null
+  uploadFileError: string | null | undefined
   textFieldHeight: number | string
 }
 
 type FileInputFieldProps = {
-  uploadFileError: string | null
+  uploadFileError: string | null | undefined
```
liangyuanruo, line 12: similarly why use `undefined` here?

src/client/user/components/CreateUrlModal/index.tsx
```diff
@@ -51,26 +55,18 @@ const useStyles = makeStyles((theme) =>
...
-  closeCreateUrlModal,
-  onCreateUrl,
-  onUploadFile,
-}) => {
   const isFullScreenDialog = useFullScreenDialog()
   const classes = useStyles({ isFullScreenDialog })
   const [step, setStep] = useState(0)
-  const incrementDecorator = (func) => async (...args) => {
+
+  const incrementDecorator = (func: any) => async (...args: any[]) => {
```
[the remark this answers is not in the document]
Oxiang: changed to hook instead

yong-jie (approved):

PR #939, opened 2020-11-23
refactor(redux): move creation of the Redux store singleton into its own file

liangyuanruo:
Following the [Redux style guide](https://redux.js.org/style-guide/style-guide#only-one-redux-store-per-app), this PR helps to make clear the singleton initialization of the Redux store by placing its creation into its own file and decoupling it from the creation of the React `<Root />` component.

**Improvements**:
- Greater clarity of Redux store creation
- Greater awareness of the use of the root reducer in Redux


yong-jie (approved):
lgtm

PR #954, opened 2020-11-24
refactor(app): refactor app domain

Oxiang:
## Problem

Incomplete migration of JSX files to TSX files

## Solution

- convert jsx files to tsx files
- convert js file to ts file
- change redux higher order connect to redux hooks
- assign types to function components, styles and methods
- replace proptypes with typescript


src/client/app/components/BaseLayout/BaseLayoutHeader.tsx
```diff
@@ -233,7 +240,7 @@ const BaseLayoutHeader = ({
             headers.map(
               (header) =>
                 (header.public ? !isLoggedIn : isLoggedIn) &&
-                !header.hidden && (
```
yong-jie, line 236: can i check what was the context behind this removal?

src/client/app/components/pages/RootPage/ScrollToTop.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const ScrollToTop = ({ children }: ScrollToTopProps) => {
+  const location = useLocation().pathname
+  const prevProps = usePrevious(location)
+  useEffect(() => {
+    if (!!prevProps && location !== prevProps) {
+      window.scrollTo(0, 0)
+    }
+  })
+  return <>{children}</>
```
yong-jie, line 24: wondering if `return children` would work

src/client/app/components/pages/RootPage/index.tsx
```diff
@@ -29,7 +28,11 @@ import {
 } from '../../../util/types'
 import theme from '../../../theme'
 
-const Root = ({ store }) => (
+type RootProps = {
+  store: any
```
yong-jie, line 32: is there a stricter subset of `any` that would work for root props? maybe the redux libraries export a store type that can be used here

yong-jie (changes requested):

src/client/app/components/pages/RootPage/ScrollToTop.tsx
yong-jie: okay nevermind scratch that, didnt work.

src/client/app/components/BaseLayout/BaseLayoutHeader.tsx
Oxiang: Headers' objects does not have 'hidden' field, so this was not used

yong-jie (approved):

Issue #955, opened 2020-11-24
Decouple login form input from Redux


liangyuanruo:
**Is your feature request related to a problem? Please describe.**
Each keystroke entering the email and OTP input fields results in an update to the Redux store.

**Describe the solution you'd like**
To decouple each input keystroke so that it no longer updates the Redux store. Instead, updates to Redux should only occur when action buttons such as "sign in" or "submit OTP" are clicked.

yong-jie:
agreed; would also like to add that OTP and email fields aren't needed beyond the component, might also be worth to just store it in state instead of tracking with redux.

PR #987, opened 2020-11-30
refactor/login: Refactor heavily-entangled LoginForm component

liangyuanruo:
## Problem

The login page and its components were heavily entwined. This made it difficult to reason about interactions and dependencies, or to remove its Redux dependenices.

Part of #955

## Solution

Refactor the `LoginForm` component to be fully functional. Instead of having inside knowledge of the login flow, it is now configured from outside the component. Instead of hardcoding if-else statements to decide which components to show, it now accepts `children` as props.

It has been upgraded to become a fully controlled component via the `value` prop, whereas previously it was relying on a hack using the `hidden` property to remove DOM input (i.e. user email) when the login page transits from email stage to the OTP stage.

### DIFF
--- a/src/client/login/components/LoginForm.tsx
+++ b/src/client/login/components/LoginForm.tsx
@@ -7,7 +7,6 @@ type LoginFormProps = {
   id: string,
   placeholder: string,
   buttonMessage: string,
-  hidden: boolean,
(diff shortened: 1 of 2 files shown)


yong-jie (approved):
lgtm!

PR #989, opened 2020-12-01
refactor: remove keystroke updates to Redux on Login page

### LINKED ISSUE #955: Decouple login form input from Redux

liangyuanruo:
## Problem

Closes #955

## Solution

Login component state management of email and OTP is now managed by `useState` hooks. The user's email is populated in Redux only upon successful login.

Other minor code improvements were also carried out, such as separating the `useEffect` hooks of login glob validation from that of fetching the login page message.

Unused Redux actions and reducers have been removed.

### DIFF
--- a/src/client/login/reducers/index.ts
+++ b/src/client/login/reducers/index.ts
@@ -8,9 +8,7 @@ import {
   LoginActionType,
   RESEND_OTP_DISABLED,
   RESEND_OTP_PENDING,
-  SET_EMAIL,
   SET_EMAIL_VALIDATOR,
-  SET_OTP,
   VERIFY_OTP_ERROR,
   VERIFY_OTP_PENDING,
 } from '../actions/types'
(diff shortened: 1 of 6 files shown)


src/client/login/reducers/index.ts
```diff
@@ -23,7 +21,6 @@ export const defaultEmailValidator = (email: string) =>
 const initialState: LoginState = {
   email: '',
```
LoneRifle, line 22: Would this be needed still? See also definition at LoginState type.

liangyuanruo: yes. we decoupled keystroke updates to Redux, but the full user email is still saved to Redux when the user clicks "Sign In".

liangyuanruo: The value is referenced when the user attempts OTP verification. It's also good to store the user email in Redux so that other components may reference it if necessary.

LoneRifle: Okay. Took a while to locate `GET_OTP_EMAIL_SUCCESS`, thanks!

LoneRifle (approved):
lgtm

PR #997, opened 2020-12-01
TSX Lint

### LINKED ISSUE #870: Lint TSX files
TSX files are not being linted. Look into this problem and fix the issue.

Oxiang:
## Problem

TSX is not included in the eslint. There were many linting errors that requires manual fix after running eslint fix. 

Closes #870 

## Solution

- Added .tsx into both the lint and lint:fix scripts
- Set new rules for tslint 
- Deconstruct props and create new prop types
- Added alt to img tags


liangyuanruo:
@Oxiang could you advise which are the commits we should be reviewing, given that it's an 88-file change?

Oxiang:
> @Oxiang could you advise which are the commits we should be reviewing, given that it's an 88-file change?

The fixes are made up of eslint fix and manual fix. Apart from commits 2 (54f8d08) and 6 (548cb66), which are fixed by the eslint, the rest of the commits are done manually and they should be reviewed.

src/client/user/components/Drawer/ControlPanel/LinkAnalytics/index.tsx
```diff
@@ -265,13 +262,13 @@ function Graphs({ data, shortUrl }: GraphsProps) {
...
-          <TabPanel value={tabValue} index={0} dir={theme.direction}>
+          <TabPanel value={tabValue} index={0}>
             <DeviceStatistics deviceClicks={data.deviceClicks} />
           </TabPanel>
-          <TabPanel value={tabValue} index={1} dir={theme.direction}>
+          <TabPanel value={tabValue} index={1}>
             <DailyStatistics dailyClicks={data.dailyClicks} />
           </TabPanel>
-          <TabPanel value={tabValue} index={2} dir={theme.direction}>
+          <TabPanel value={tabValue} index={2}>
```
liangyuanruo: removing this doesn't affect anything? e.g. mobile view

liangyuanruo (approved):
lgtm, but i think we need better coordination with @yong-jie as well, so as to avoid any massive merge conflicts.

src/client/user/components/Drawer/ControlPanel/LinkAnalytics/index.tsx
Oxiang: it doesn't seem like it affected anything. Both mobile and desktop view looks fine

Issue #977, opened 2020-11-26
Allow searching of longUrl on GoDirectory


lisatjide:
**Is your feature request related to a problem? Please describe.**
Currently the user is only able to search links based on the shorturl alone, however we notice users are also pasting in long links to search for shortlinks associated with it. 

Similar problem happens when user pasted go.gov.sg/short-link. 

**Describe the solution you'd like**
Fetch results when user searches for long link or go.gov.sg/short-link

PR #998, opened 2020-12-02
Include long links into search

Oxiang:
## Problem

Currently, users are only able to search keywords based on short links and description. However, users are also searching long links as recorded by google analytics.

Partially address #977 

## Solution

**Improvements**:

- Add longUrl as part of search vector
- Script for db migration to include longUrl into the search index
- description ranking is dropped to the lowest as they are no longer being filled by users
- shortUrl ranking still remains at the top as they are the most frequently used keywords


liangyuanruo (dismissed):
actually, instead of dropping the existing index, could we have both the old and new indexes co-exist? this will allow the database to support both old and new application versions, which will be necessary if there is a rollback.

src/server/models/url.ts
```diff
@@ -164,7 +164,7 @@ export const Url = <UrlTypeStatic>sequelize.define(
         fields: ['userId'],
       },
       {
-        name: 'urls_weighted_search_idx',
+        name: 'new_urls_weighted_search_idx',
```
LoneRifle, line 167: please, never claim something is new. a more appropriate name could be `url_entry_weighted_search_idx` or similar

LoneRifle (changes requested):

LoneRifle (approved):
lgtm

Issue #869, opened 2020-11-09
Reorganise backend files to use modules


liangyuanruo:
Backend source code would be clearer if grouped into modules instead of filetype.

Introduce a module system so that related files can live together.

e.g.

The module located at `src/server/modules/login` would contain the routes, controllers, services necessary for login APIs.

Models and repositories should live outside the modules in their own folders as they are shared across services.

PR #1002, opened 2020-12-02
refactor: move sentry, user codebase to modules

LoneRifle:
## Problem

Ongoing work for #869 

## Solution

- create the sentry module, move controller, service therein
- create the user module, move controller, services, types therein
- eliminate controller and service interfaces


src/server/modules/user/__tests__/UserController.test.ts
```diff
@@ -22,6 +22,7 @@ const userAnnouncement = {
 
 describe('UserController', () => {
   const controller = new UserController(
+    // @ts-ignore - no need for private properties in UrlManagementService
```
liangyuanruo, line 25: sorry, could you elaborate more on why `@ts-ignore` is needed here?

liangyuanruo (approved):

src/server/modules/user/__tests__/UserController.test.ts
LoneRifle: tsc complains of the absence of the private properties that it expects to find in UrlManagementService. We cannot simply add private properties to the mock, since... I decided that removing interfaces is for the best.

PR #1006, opened 2020-12-02
refactor(user): use interface for controller ctor arg again

LoneRifle:
## Problem

Ongoing work for #869 

## Solution
Use an interface for the ctor argument that usually receives
the UrlManagementService. This allows us to helpfully locate
impls for the argument type using the IDE.

- Create a new interfaces dir with UrlManagementService in it
- Take care with naming to minimise confusion while still keeping
  names consistent and clear, despite using identical names for
  both interface and impl

Take the trouble also to correct the linting config for the pre-commit hook

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,10 +186,10 @@
   },
   "lint-staged": {
     "**/*.(js|jsx)": [
-      "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.json --fix"
+      "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.js.json --fix"
     ],
-    "**/*.ts": [
-      "eslint --ext .ts --no-eslintrc -c .eslintrc.ts.json --fix"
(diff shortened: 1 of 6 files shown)


liangyuanruo (approved):
can we have a better name than `CanManageLinks`? what about sticking to `UrlManagementService`? 

also, is it really necessary to move away from interfaces?

LoneRifle:
> can we have a better name than `CanManageLinks`? what about sticking to `UrlManagementService`?
> 
> also, is it really necessary to move away from interfaces?

I'll think of a better name. I really don't think programming to interfaces is what we want; we only have a single concrete implementation for a management service

LoneRifle:
Okay, I've come round. I neglected to consider how support in the IDE would be like, and it is really nice to be able to seek implementations for a given interface easily. 

I would personally would have preferred to do away with interfaces altogether, but jest's ability to mock classes with private properties leaves much to be desired.

Issue #984, opened 2020-11-30
Move click count out of URL table


lisatjide:
**Is your feature request related to a problem? Please describe.**
As we are moving forward with features that require bulk database operations, we want to move click count (a frequently updated field) in the URL table.

This is because updating the click count atomically requires a row lock, and it is unlikely that bulk operations will be able to acquire the necessary locks to proceed if a link is being visited constantly.

LoneRifle:
Sole outstanding issue tracked by #1097

PR #1011, opened 2020-12-07
Mirror URL table's click column

yong-jie:
## Problem

As we are moving forward with features that require bulk database operations, we want to move click count (a frequently updated field) in the URL table.

This is because updating the click count atomically requires a row lock, and it is unlikely that bulk operations will be able to acquire the necessary locks to proceed if a link is being visited constantly.

Part 1 of #984 

## Solution

Visit our [wiki page](https://github.com/opengovsg/GoGovSG/wiki/URL-Click-Column-DB-Migration) for in-depth discussion of solution.

Proposed migration approach involves multiple phases.

1) Deploy this PR that will let sequelize create a new `url_clicks` table.

2) Run the migration script in this PR that will populate the `url_clicks` table, as well as create the necessary hooks to keep them always in sync.

3) Deploy a separate PR that changes application code to do read/writes to `url_clicks` table.

4) Run a migration script to remove hooks and `url` table's click column.

**Side effects of PR**:

- Original `Clicks` model renamed to `DailyClicks`.

## Deploy Notes

Deploy code first to let sequelize create the tables before running migration script.

**New scripts**:

- `20201207_synchronise_url_clicks.sql` : Copies click count into url_clicks table, and adds DB hooks to keep it in sync.

### DIFF
--- a/src/server/db_migrations/20201207_synchronise_url_clicks.sql
+++ b/src/server/db_migrations/20201207_synchronise_url_clicks.sql
@@ -1,0 +1,38 @@
+-- This SQL file is used to populate the url_clicks table with data from the
+-- url_table, and also add the necessary hooks such that they will always be
+-- in sync. These include 1) when a url_table row's click is changed, and
+-- 2) when a new url_table row is created.
+BEGIN TRANSACTION;
+
+INSERT INTO "url_clicks" ("shortUrl", "clicks", "createdAt", "updatedAt")
+SELECT "shortUrl", "clicks", "createdAt", "updatedAt" from urls;
+
+CREATE OR REPLACE FUNCTION update_clicks()
+RETURNS TRIGGER AS $$
+BEGIN
+    UPDATE url_clicks
+    SET "clicks" = NEW."clicks"
+    WHERE url_clicks."shortUrl" = NEW."shortUrl";
+    RETURN NEW;
+END; $$ LANGUAGE PLPGSQL;
+
+CREATE TRIGGER url_table_click_incremented
+    AFTER UPDATE of "clicks" on urls
+    FOR EACH ROW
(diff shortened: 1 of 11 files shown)


src/server/db_migrations/20201207_synchronise_url_clicks.sql
```diff
@@ -0,0 +1,38 @@
+-- This SQL file is used to populate the url_clicks table with data from the
+-- url_table, and also add the necessary hooks such that they will always be
+-- in sync. These include 1) when a url_table row's click is changed, and
+-- 2) when a new url_table row is created.
+BEGIN TRANSACTION;
+
```
liangyuanruo, line 6: are we missing a `CREATE TABLE` statement?

src/server/models/statistics/daily.ts
```diff
@@ -3,32 +3,35 @@ import Sequelize from 'sequelize'
...
-    validate: {
-      is: /^[a-z0-9-]+$/,
+export const DailyClicks = <DailyClicksTypeStatic>sequelize.define(
+  'daily_stats',
+  {
+    shortUrl: {
+      type: Sequelize.STRING,
+      primaryKey: true,
+      validate: {
+        is: /^[a-z0-9-]+$/,
```
liangyuanruo, line 25: can you help to check if this regex is being repeated throughout the codebase? i think we should extract it into a single shared variable

liangyuanruo (approved):
some suggestions but lgtm!

src/server/models/statistics/daily.ts
yong-jie: 👍  i've extracted all uses and declared the variable in our shared directory for both frontend and backend reuse

PR #1027, opened 2020-12-08
refactor(route): allow all private route to redirect

### LINKED ISSUE #988: Investigate necessity for Login page GA to cater to GoDirectory redirects
The Google Analytics implementation on the login page has to cater to an edge case if the user arrives from the GoDirectory page, even though the user did not need to interact with the Login page at all.

Investigate why this is happening and remove the edge case.

```typescript
// in LoginPage component
if (location?.state?.previous !== '/directory') {
       GAPageView('EMAIL LOGIN PAGE')
       GAEvent('login page', 'email')
}
```

Oxiang:
## Problem

Currently, all private page routes (user and directory) will be directed to login page before redirecting back to the private page, on refresh. Directory page's redirect is coded as an edge case which can be generalised to allow redirect from all private routes

Closes #988 

## Solution

- Remove directory previous location check and enable all private route 

### Refactors
- Move GA recording for directory to directory page
- Change login defaultprops to typescript default value

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,10 +186,10 @@
   },
   "lint-staged": {
     "**/*.(js|jsx)": [
-      "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.json --fix"
+      "eslint --ext .js,.jsx --no-eslintrc -c .eslintrc.js.json --fix"
     ],
-    "**/*.ts": [
-      "eslint --ext .ts --no-eslintrc -c .eslintrc.ts.json --fix"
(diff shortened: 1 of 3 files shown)


LoneRifle (approved):
I've already fixed husky, so you may want to rebase/rework the commit

Oxiang:
> I've already fixed husky, so you may want to rebase/rework the commit

since our husky is the same and there is no conflict, i will just merge it

PR #1037, opened 2020-12-10
Search full shorturl

### LINKED ISSUE #977: Allow searching of longUrl on GoDirectory

Oxiang:
## Problem

Users are unable to search for shortUrl when they enter the entire url in the search box e.g. https://go.gov.sg/search-this.

Closes #977 

## Solution

Add another layer for keyword search to extract the shortUrl if the search query matches the regex pattern. 
e.g. **https://go.gov.sg/search-this** will extract **search-this**
If it does not match the regex, the original query will be used

Exclude query without https header 
e.g. go.gov.sg/search-this as users only search (guessing it is a copy and paste action) the full url link (as found in GA)

- add unit test for parse.ts


src/server/repositories/UrlRepository.ts
```diff
@@ -140,54 +132,14 @@ export class UrlRepository implements UrlRepositoryInterface {
...
-          state,
-          isFile,
-        ))
-
-    return urlsModel
-  }
-
-  private async getRelevantUrlsFromEmail(
-    query: string,
+  public async getRelevantUrlsFromEmail(
```
LoneRifle, line 135: Okay, this is a bad idea. you're allowing the arbitrary injection of SQL via `rankingAlgorithm`, `queryState` and `queryFile` through a public method. I know these are only accessed from a service, but I think that you have to trade-off between biz logic and security here.

LoneRifle (changes requested):

test/server/util/parse.test.ts
```diff
@@ -0,0 +1,18 @@
+import { extractShortUrl } from '../../../src/server/util/parse'
```
LoneRifle, line 1: If you're making a new test, could you kindly put it under `src/server/util/__tests__` instead please?

Oxiang: shifted the test to the path

src/server/repositories/UrlRepository.ts
Oxiang: make sense, I have reverted the parsing back to urlrepository and made them private again

src/server/util/__test__/parse.test.ts
```diff
@@ -0,0 +1,18 @@
+import { extractShortUrl } from '../parse'
```
LoneRifle, line 1: `__tests__`, not `__test__`

src/server/repositories/UrlRepository.ts
```diff
@@ -228,14 +229,16 @@ export class UrlRepository implements UrlRepositoryInterface {
     state: string | undefined,
     isFile: boolean | undefined,
   ): Promise<UrlDirectoryPaginated> {
+    // Extract shortUrls with regex
+    const newQuery = extractShortUrl(query)
```
LoneRifle, line 233: I'm not convinced that it is the server's responsibility to correct user input. @yong-jie wdyt?

LoneRifle (changes requested):

src/server/repositories/UrlRepository.ts
yong-jie: agreed in general that server shouldnt correct user input. in this control flow they're supposed to be searching for keywords, and a shortUrl isnt a keyword. 

Extracting the shortURL here would mean that either 1) our "keyword" header is incorrect, or  2) this function is doing things that it's not supposed to do. Both scenarios are undesirable

That being said, it's strange that we didnt give them a way to explicitly search a shortlink. I think even we as developers are finding this weird, which is why we're anticipating user mistakes like these. 

It might be worth to discuss whether we should have an option to search shortlinks too(?)

src/server/util/parse.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Checks for full short link URL in the query and extract out the short link.
+ * @param query The query to check.
+ */
+export function extractShortUrl(query: string) {
+  const regExp = new RegExp(`^https://${ogHostname}/(.*)`)
+  const capture = regExp.exec(query)
+  if (capture) {
+    return capture[1]
+  }
+  return query
```
yong-jie, line 28: returning `query` makes this function a misnomer. this function extracts the shortURL. if there's no shortURL to be found, my impression is that `nothing` would be extracted. would returning `null` be a more valid behaviour here?

then other functions that call it would check if there are any shortURLs that can be extracted, if not, use whatever fallback mechanisms available. either way, it's not the job of this function to fallback for them.

yong-jie (changes requested):

src/server/repositories/UrlRepository.ts
LoneRifle: I don't know. we are taking on some complexity, and the gain is somewhat questionable. should the API be that forgiving?

src/server/repositories/UrlRepository.ts
```diff
@@ -228,14 +229,16 @@ export class UrlRepository implements UrlRepositoryInterface {
     state: string | undefined,
     isFile: boolean | undefined,
   ): Promise<UrlDirectoryPaginated> {
+    // Extract shortUrls with regex
+    const newQuery = this.queryFilter(query)
```
LoneRifle, line 233: you could simply do...
```suggestion
    const newQuery =extractShortUrl(query) || query
```

LoneRifle (approved):
lgtm

yong-jie (approved):

PR #1038, opened 2020-12-10
fix: directory toggle ui

Oxiang:
## Problem

The arrow in the directory page toggle have been shifted down

## Solution

Change the css from inline-block to box

## Before & After Screenshots

**BEFORE**:
Screenshot 2020-12-10 at 10 45 19 AM

**AFTER**:
Screenshot 2020-12-10 at 11 01 08 AM


LoneRifle (approved):
I'm alright if the designers are alright with it - last I remembered the wireframes deliberately wanted it that way

PR #1054, opened 2020-12-14
Remove index

Oxiang:
## Problem

Now that the production database uses **urls_combined_weighted_search_idx**, **urls_weighted_search_idx** is obsolete and can be deleted.

The previous db migration script that was meant to create **urls_combined_weighted_search_idx** is unused as sequelize handles the index creation. The script can be remove as well.

## Solution

- New db_migration script to delete urls_weighted_search_idx index
- Remove 20201202_add_combined_search_index migration script from codebase


LoneRifle (changes requested):
we don't usually remove scripts for migrations. lgtm otherwise.

Oxiang:
> we don't usually remove scripts for migrations. lgtm otherwise.

In this case we did not use the migration script at all, would leaving it in cause confusion?

LoneRifle:
Ah, that's true. let me change it.

LoneRifle (approved):

yong-jie (approved):

PR #1055, opened 2020-12-14
Record search result clicks for search evaluation metric

Oxiang:
## Problem

Going forward, we need a way to evaluate GoDirectory's search quality.
One of the way is to track the ranking of the results that are clicked by the users and the total results returned by the search query.

Gathering this data will allow us to calculate a meaningful baseline score and set google analytics goals that can be use as benchmark for future improvements.

## Solution

- Enable google analytics event to record value
- Include result count when recording user's search query
- Record click ranking and its respective query


yong-jie (approved):
lgtm

PR #1091, opened 2020-12-21
Move reading and writing of url table's clicks column to sharded table

yong-jie:
## Problem

Part 2 of #984 

## Solution

**Features**:

- [x] Remove `clicks` column definition in Url model.
- [x] Add `getClicks` scope for Url model. In order to fetch the clicks object, repositories must use this scope.
- [x] Modify `UrlRepository` and `UserRepository` to use new `getClicks` scope when fetching urls.
- [x] Modify `UrlMapper`'s `persistenceToDto` method to throw an error when a url is supplied without an associated `UrlClicks` object. This allows it to guarantee a complete `StorableUrl` object to be returned for abstraction layers above the repository.
- [x] Tweak GoSearch queries to `JOIN ON "url_clicks"."shortUrl" = "url"."shortUrl"`
- [x] Tweak ranking algorithm's `Popularity` order to `ORDER BY url_clicks.clicks`  instead of `ORDER BY url.clicks`
- [x] Modify LinkStatisticsRepository to call the `sum` method of `UrlClicks` instead of `Url.clicks`
- [x] Fix the 20-30 tests that started breaking as a result of this change
- [x] Tweak `updateLinkStatistics` SQL function to write to `url_clicks` instead of `url` table

**Improvements**:

- [x] Remove `Relevance` filter option for url search. This was used for GoSearch, but not for GoDirectory.

## Deploy Notes

1) Deploy application code directly.

2) Once confident that no IO has been made on the Url.clicks column, run the migration script to delete it.

**New scripts**:

- `20201221_remove_url_clicks_column.sql` : The final step in this DB sharding task. Removes all triggers/functions and deletes the Url.clicks column

### DIFF
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -179,6 +179,7 @@ export const Url = <UrlTypeStatic>sequelize.define(
           },
         )
 
+        // TODO: change to create after DB triggers are removed.
         await UrlClicks.upsert(
           {
             shortUrl: url.shortUrl,
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -1,5 +1,5 @@
 import Sequelize from 'sequelize'
-import { UrlClicks } from './statistics/clicks'
+import { UrlClicks, UrlClicksType } from './statistics/clicks'
 import { ACTIVE, INACTIVE } from './types'
 import {
   SHORT_URL_REGEX,
@@ -24,7 +24,7 @@ interface UrlBaseType extends IdType {
 }
 
 export interface UrlType extends IdType, UrlBaseType, Sequelize.Model {
-  readonly clicks: number
+  readonly UrlClicks?: UrlClicksType
   readonly createdAt: string
   readonly updatedAt: string
   readonly email: string
@@ -144,11 +144,6 @@ export const Url = <UrlTypeStatic>sequelize.define(
       values: [ACTIVE, INACTIVE],
       defaultValue: ACTIVE,
     },
-    clicks: {
-      type: Sequelize.INTEGER,
-      allowNull: false,
-      defaultValue: 0,
-    },
     isFile: {
       type: Sequelize.BOOLEAN,
       allowNull: false,
@@ -231,6 +226,16 @@ export const Url = <UrlTypeStatic>sequelize.define(
         fields: [Sequelize.literal(`(${urlSearchVector})`)],
       },
     ],
+    scopes: {
+      getClicks: {
+        include: [
+          {
+            model: UrlClicks,
+            as: 'UrlClicks',
+          },
+        ],
+      },
+    },
   },
 )
 
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -1,4 +1,5 @@
 import Sequelize from 'sequelize'
+import { UrlClicks } from './statistics/clicks'
 import { ACTIVE, INACTIVE } from './types'
 import {
   SHORT_URL_REGEX,
@@ -178,6 +179,15 @@ export const Url = <UrlTypeStatic>sequelize.define(
         // eslint-disable-next-line no-use-before-define
         await writeToUrlHistory(
           url,
+          options as Sequelize.CreateOptions & {
+            transaction: Sequelize.Transaction
+          },
+        )
+
+        await UrlClicks.upsert(
+          {
+            shortUrl: url.shortUrl,
+          },
           options as Sequelize.CreateOptions & {
             transaction: Sequelize.Transaction
           },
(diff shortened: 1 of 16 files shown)


src/server/repositories/UrlRepository.ts
```diff
@@ -51,32 +47,40 @@ export class UrlRepository implements UrlRepositoryInterface {
   public findByShortUrl: (
     shortUrl: string,
   ) => Promise<StorableUrl | null> = async (shortUrl) => {
-    return (await Url.findOne({
+    const url = await Url.scope('getClicks').findOne({
       where: { shortUrl },
-    })) as StorableUrl | null
+    })
+    return this.urlMapper.persistenceToDto(url)
   }
```
yong-jie, line 54: this was an incredibly bad practice to blindly cast the output of `findOne` immediately to the DTO's `StorableUrl` type. it turned off typescript's typechecks and could have been fatal for the server if not caught in development.

I'm adding an extra step of getting the `urlMapper` to convert from raw `Url` to `StorableUrl`. Because this throws when data is incomplete, we can be better assured of type safety.

src/server/models/url.ts
```diff
@@ -182,6 +178,15 @@ export const Url = <UrlTypeStatic>sequelize.define(
...
         )
+
+        await UrlClicks.upsert(
+          {
+            shortUrl: url.shortUrl,
+          },
+          options as Sequelize.CreateOptions & {
+            transaction: Sequelize.Transaction
+          },
+        )
```
yong-jie, line 189: using `upsert` is necessary during this transition phase because the DB triggers are still in place, and this step clashes with what the trigger is trying to do. Once the trigger is removed, we can change `upsert` to `create` if we want to

liangyuanruo:
i took a look at the draft code, lgtm so far!

src/server/repositories/UrlRepository.ts
liangyuanruo: do u know why the previous developer felt it was necessary in the first place?

src/server/models/url.ts
liangyuanruo: file an issue and add a TODO?

liangyuanruo: on another note, should the typecasting of `options as Sequelize.CreateOptions` be revisited?

liangyuanruo (approved):
Code quality LGTM, but there is still the special treatment of the clicks column because of its prior existence in the Url table. For example, the type of `StorableUrl` was simply extended.

I agree the current implementation is backward compatible, but would there be value in separating them more fully? I haven't delved fully into the specifics, but I suspect we could discern between read concerns of having a URL object with total clicks (perhaps using a new DTO type), versus the previous use of `StorableUrl`.

liangyuanruo:
Happy to get this merged in first, but let me know what you think?

liangyuanruo:
> Once confident that no IO has been made on the Url.clicks column, run the migration script to delete it.

We should also check that there are no discrepancies between the Url and UrlClicks table, since there's no going back once the DB script is run.

i.e. the following query should return zero rows.

```sql
SELECT COUNT(*)
FROM "Url"
JOIN "UrlClicks"
ON "Url"."shortUrl" = "UrlClicks"."shortUrl"
WHERE "Url"."clicks" <> "UrlClicks"."clicks"
```

src/server/repositories/UrlRepository.ts
yong-jie: my guess is that previous developer was certain that running `Url.findOne` would always yield an object with all the necessary fields of `StorableUrl`, so did a quick typecast so that the return type matches what's defined in the interface.

But this isn't true anymore, considering a `Url` object by itself is now missing the `clicks` property.

yong-jie:
> I agree the current implementation is backward compatible, but would there be value in separating them more fully? I haven't delved fully into the specifics, but I suspect we could discern between read concerns of having a URL object with total clicks (perhaps using a new DTO type), versus the previous use of `StorableUrl`.

The current approach of combining urls and clicks reduces the number of abstraction layers that need to be changed (at present only repositories and models have been modified). Moving forward, I agree that having a clear separation would be cleaner for the codebase and could reduce coupling between these two models.

However, there would always be use-cases that necessitate a binding between urls and clicks, such as fetching urls ordered by click count. A full separation might never be possible if we want to cater to these cases.

src/server/models/url.ts
yong-jie: agreed, it will still pass typescript checks even without typecasting, and I added this to be consistent with `writeToUrlHistory`. This can be changed together in a subsequent PR when we change `upsert` to `create`.

src/server/repositories/UrlRepository.ts
liangyuanruo: i see

src/server/models/url.ts
liangyuanruo: ok, please include in issue

liangyuanruo:
I see, if there is never a need to read a URL without clicks, perhaps we can simply rename the function in the URL repository layer to `findByShortUrlWithTotalClicks`. The main thing is not to mislead the next developer coming along.

yong-jie:
sounds good. all actionables will be added to an issue together with the TODO. merging now.

PR #1107, opened 2021-01-03
Use enum for user redux actions instead of constant value strings

### LINKED ISSUE #937: Use enums for Redux actions
User actions are currently typed as constant value strings. For example, in `src/client/user/actions/types.ts`:

```typescript
/* User actions */
export const CREATE_URL_SUCCESS = 'CREATE_URL_SUCCESS'
export const GET_URLS_FOR_USER_SUCCESS = 'GET_URLS_FOR_USER_SUCCESS'
export const OPEN_CREATE_URL_MODAL = 'OPEN_CREATE_URL_MODAL'
...
```

This causes import bloat in `src/client/user/actions/index.ts`:

```typescript
import {
  // 40 lines of imports
} from './types'
```

Convert the implementation to use TypeScript `enum`s instead.

barnettx:
## Problem

Import bloat in `src/client/user/actions/index.ts` due to actions being stored as constant value strings in `src/client/user/actions/types.ts`

Closes #937

## Solution

Replace constant value strings with `UserAction` [String Enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) as suggested in issue.

## Tests

`npm run build` compiled successfully.

First time contributing to this repo, kindly let me if there are areas I can improve in.

### DIFF
--- a/src/client/user/actions/types.ts
+++ b/src/client/user/actions/types.ts
@@ -6,49 +6,52 @@ import {
 } from '../reducers/types'
 
 /* User actions */
-export const CREATE_URL_SUCCESS = 'CREATE_URL_SUCCESS'
-export const GET_URLS_FOR_USER_SUCCESS = 'GET_URLS_FOR_USER_SUCCESS'
-export const OPEN_CREATE_URL_MODAL = 'OPEN_CREATE_URL_MODAL'
-export const CLOSE_CREATE_URL_MODAL = 'CLOSE_CREATE_URL_MODAL'
-export const SET_SHORT_URL = 'SET_SHORT_URL'
-export const SET_LONG_URL = 'SET_LONG_URL'
-export const SET_EDITED_LONG_URL = 'SET_EDITED_LONG_URL'
-export const SET_RANDOM_SHORT_URL = 'SET_RANDOM_SHORT_URL'
-export const RESET_USER_STATE = 'RESET_USER_STATE'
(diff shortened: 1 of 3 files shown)


src/client/user/actions/types.ts
```diff
@@ -6,49 +6,52 @@ import {
...
+  SET_EDITED_DESCRIPTION = 'SET_EDITED_DESCRIPTION',
+  SET_USER_MESSAGE = 'SET_USER_MESSAGE',
+  SET_USER_ANNOUNCEMENT = 'SET_USER_ANNOUNCEMENT',
+  SET_URL_UPLOAD_STATE = 'SET_URL_UPLOAD_STATE',
+  SET_FILE_UPLOAD_STATE = 'SET_FILE_UPLOAD_STATE',
+}
 
 export type SetUrlUploadStateAction = {
-  type: typeof SET_URL_UPLOAD_STATE
+  type: typeof UserAction.SET_URL_UPLOAD_STATE
```
liangyuanruo, line 39:
```suggestion
  type: UserAction.SET_URL_UPLOAD_STATE
```

I think we can also take the opportunity to narrow the type even further for all these actions, as the corresponding action will always have that exact string value and no other.

liangyuanruo (changes requested):
Hi @barnettx ,

Thanks for contributing to the repository!

I've made a suggestion, but the code looks good otherwise. Happy to approve once the changes have been made.

Thanks!

liangyuanruo (approved):
thanks for your contribution!

PR #1134, opened 2021-01-19
chore(robots): disallow NLB crawler from parsing beyond the application bundle

LoneRifle:
Keep crawlers away from bundle.js as there are no real links in here

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,2 +1,7 @@
-User-agent: *
-Disallow: /bundle.js
+User-agent: NLB_archive_bot
+
+Allow: /
+Allow: /bundle.js
+Allow: /css*
+Allow: /assets/*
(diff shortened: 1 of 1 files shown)


liangyuanruo (changes requested):
Since Go is a React single-page app, disallowing `bundle.js` effectively means that the crawler is blind and can't even see the landing page at all. Is this what we want?

How about if we only allow loading the landing page, but disallow all other server routes, including `/api` and redirect endpoints?

From what I can see loading Go in Chrome Devtools, this should work:

```text
user-agent: <NLB crawler>

allow: /
allow: /bundle.js
allow: /css2
allow: /assets/*
allow: /s/*.woff2
disallow: *
```

liangyuanruo (approved):

PR #1151, opened 2021-02-01
refactor(modules): remove use of interfaces where possible

LoneRifle:
## Problem

The use of interfaces continues to be unwieldy in the Go codebase, often providing 
a layer of indirection for just one implementation, rather than multiple possible impls.

To ensure that the codebase remains generally free of clutter, we should prefer to 
use classes directly where it makes sense.

Part of work for #869

## Solution

Use `Pick<T, keyof T>` in lieu of interfaces so that we can properly
define the 1-1 relationship between eg a controller and its service,
while still allowing Jest to provide a mock for testing purposes.

Interfaces were declared for threat scanners for initial requirements
that multiple services may be used to scan for threats. An interface
was needed then to ensure consistency between the services. Given
however that we have decided on Cloudmersive and Google Safe Browsing
for file and url threats respectively, these interfaces are no longer
needed.

Remove dead code in CloudmersiveScanService

This PR also contains two linting changes that were auto-inserted by VSCode.


liangyuanruo (changes requested):
Say for example this code:

```typescript
  private fileTypeFilterService: Pick<FileTypeFilterService, keyof FileTypeFilterService>
```

If you wanted to do remove interfaces, why wouldn't you simplify it to the class itself?

```typescript
  private fileTypeFilterService:FileTypeFilterService
```

LoneRifle:
> Say for example this code:
> 
> ```ts
>   private fileTypeFilterService: Pick<FileTypeFilterService, keyof FileTypeFilterService>
> ```
> 
> If you wanted to do remove interfaces, why wouldn't you simplify it to the class itself?
> 
> ```ts
>   private fileTypeFilterService:FileTypeFilterService
> ```

Because then the consumer would expect the _private_ properties of FileTypeFilterService to be there too. This is okay with normal operation, but affects testing: without resorting to some really bad syntax, Jest is unable to provide a mock containing private properties to the consumer, and will complain that you are attempting to mock private properties, which you cannot do, but need.

liangyuanruo (commented):
Unless your intention is to make it easy to mock in the tests?

If so would it be better to parametrise this?

```typescript
type Service<T> = Pick<T, keyof T>
```

But this is just an interface, except this time the dependency is inverted - the interface is defined by the concrete class instead of the other way around.

liangyuanruo:
> The use of interfaces continues to be unwieldy in the Go codebase, often providing
a layer of indirection for just one implementation, rather than multiple possible impls.

But it's easy to write mocks that implement an interface, right?

> To ensure that the codebase remains generally free of clutter, we should prefer to
use classes directly where it makes sense.

But the implementation isn't using classes, it's basically using `Pick` to create a new interface?

> Because then the consumer would expect the private properties of FileTypeFilterService to be there too. This is okay with normal operation, but affects testing: without resorting to some really bad syntax, Jest is unable to provide a mock containing private properties to the consumer, and will complain that you are attempting to mock private properties, which you cannot do, but need.

I'm just not convinced this is a real problem.

LoneRifle:
> > The use of interfaces continues to be unwieldy in the Go codebase, often providing
> > a layer of indirection for just one implementation, rather than multiple possible impls.
> 
> But it's easy to write mocks that implement an interface, right?

Yes, hence what you see in the unit tests in the PR - the mocks are implementing what amounts to an interface

> 
> > To ensure that the codebase remains generally free of clutter, we should prefer to
> > use classes directly where it makes sense.
> 
> But the implementation isn't using classes, it's basically using `Pick` to create a new interface?

Correct - the consumer basically is saying "I want this class, but am only interested in its public properties". In that way, you _deliberately_ couple say, the controller with its service, so that you can _both_ avoid the clutter that comes with having a spurious interface, _and still_ have a somewhat elegant way for Jest to provide the mock

> 
> > Because then the consumer would expect the private properties of FileTypeFilterService to be there too. This is okay with normal operation, but affects testing: without resorting to some really bad syntax, Jest is unable to provide a mock containing private properties to the consumer, and will complain that you are attempting to mock private properties, which you cannot do, but need.
> 
> I'm just not convinced this is a real problem.

The alternative is having interfaces everywhere, mostly to provide a layer of indirection just for Jest, which I really do not wish to go out of my way to accommodate. The naming already is proving cumbersome, whatwith having the `*Interface` suffix with each dependency that you wish to inject

LoneRifle:
> Unless your intention is to make it easy to mock in the tests?
> 
> If so would it be better to parametrise this?
> 
> ```ts
> type Service<T> = Pick<T, keyof T>
> ```
> 
> But this is just an interface, except this time the dependency is inverted - the interface is defined by the concrete class instead of the other way around.

I did try to introduce a new global:
```ts
type Public<T> = Pick<T, keyof T>
```
but ESLint will have none of it, and I wasn't keen on introducing a codebase-specific global into eslintrc.js

PR #1152, opened 2021-02-01
refactor(modules): create display, qr, directory modules

LoneRifle:
## Problem

The infrastructure to generate QR codes is poorly organised and too dependent on
calling inversify to resolve its dependencies. This makes it inconsistent in structure
relative to the other controllers and modules in Go.

Whilst we are here, we should move RotatingLinksController into its own module

Part of ongoing work for #869 

## Solution
- Create display module, move RotatingLinksController into it
- Create qr module, move QrCodeService and assets into it
- Extract QrCodeController from `api/qr` into `modules/qr`
- move GoDirectory to module
- Inject config, existing inversify dependencies into classes
  wherever possible


src/server/modules/qr/QrCodeController.ts
```diff
@@ -0,0 +1,58 @@
...
+  ) {
+    this.ogUrl = ogUrl
+    this.qrCodeService = qrCodeService
+    this.urlRepository = urlRepository
+  }
+
+  private shortUrlExists: (shortUrl: string) => Promise<boolean> = async (
+    shortUrl,
+  ) => {
+    return !!(await this.urlRepository.findByShortUrl(shortUrl))
```
yong-jie, line 31: This will be the first time a controller calls a repository directly in our codebase. I'm getting [mixed responses](https://softwareengineering.stackexchange.com/questions/306890/is-it-bad-practice-that-controller-call-repository-instead-of-service) online whether this is a good practice.

Some factors complicate this on top of what was discussed on stackoverflow (will edit if i think of more):

1) This is really just an existence predicate. does it really need to be in a service?

Open to discussion on this matter, we can also talk about this in syncup

yong-jie (commented):
Agreed that the qrcode route was exposing too much logic and should be refactored into its own controller. Most of the code lgtm; i'm raising one discussion point to ponder about before changes go in.

src/server/modules/qr/QrCodeController.ts
LoneRifle: As noted, there really isn't much biz logic to isolate from the controller. While it is true that this is slightly inconsistent from the controller -> service -> repository pattern that is common in this codebase, I don't think it sets a precedent for going directly to repositories.

Given also that this is existing code lifted from the route into its own module, I would lean towards keeping it for now, and refactoring it later as the need arises.

yong-jie (approved):

PR #1172, opened 2021-02-08
refactor(modules): make auth module for login/out

LoneRifle:
## Context
- Move LoginController, LogoutController, and supporting cast into `modules/auth`
- Rework tests to use Jest functions where appropriate, to decouple from `test/`

Part of routine work for #869


yong-jie (approved):
lgtm

PR #1190, opened 2021-02-14
Url click table migration cleanup

### LINKED ISSUE #1097: Update onCreate hook for Urls model after migration is complete
When a url object is created, an `upsert` on url_clicks was used so that it does not clash with an existing database trigger. Once that trigger is removed, we can switch `upsert` to `create`.

At the same time, remove the typecast in the `upsert` operation; that cast is unnecessary and does not fail typechecks.

Also rename  `findByShortUrl` to `findByShortUrlWithTotalClicks` for clarity.

yong-jie:
## Problem

There are a few outstanding issues surrounding the url click table migration that needs to be cleaned up.

Closes #1097 

## Solution

- Change `upsert` to `create` in url's onCreate hook
- Remove unnecessary typecast of the transition object in `create`
- Rename `findByUrl` to `findByUrlWithTotalClicks` for better clarity

### DIFF
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -184,9 +184,7 @@ export const Url = <UrlTypeStatic>sequelize.define(
             shortUrl: url.shortUrl,
             clicks: 0,
           },
-          options as Sequelize.CreateOptions & {
-            transaction: Sequelize.Transaction
-          },
+          options,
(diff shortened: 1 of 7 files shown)


LoneRifle (approved):
lgtm

PR #1258, opened 2021-02-25
fix(gitpod): install localstack via gitpod, not dockerfile

LoneRifle:
Resolve the problem with gitpod once and for all


yong-jie (approved):
lgtm

Issue #777, opened 2020-10-22
Setup for.edu.sg


lisatjide:
**Is your feature request related to a problem? Please describe.**
Go.gov.sg is currently only available to gov.sg email users, hence, there is a group of users on edu.sg emails who are not able to access it. They are mainly users from the tertiary institutions such as polytechnics and ITEs. 

**Describe the solution you'd like**
A separate link shortener service for the edu.sg users. This will be hosted on for.edu.sg

liangyuanruo:
@lisatjide can you detail what changes to the application should be made before the infrastructure is set up?

PR #1275, opened 2021-03-04
refactor(locales): move to using static file

LoneRifle:
## Problem

We want to use a single codebase for several similar deployments, but 
with different themes and copywriting text. To that end, we need resource strings
to be easily substitutable. As of now, they are implemented as a TypeScript module
webpacked into the bundle

Part of work for #777 

## Solution

Move translation resource strings to static file, so that we can easily
substitute such files for deployments using the Go.gov.sg codebase

- `npm install i18next-http-backend`
- `app/assets/.../translation` -> `public/locales/.../translation.json`
  - Make use of i18next's in-built nesting feature in lieu of the
    string interpolation were doing in the `app/assets` ts file
- add i18next-http-backend to i18next bootstrap, so that we fetch
  translations from a static path instead of having it injected into
  the config
- wait on i18next initialization before proceeding to render
- FeatureListSliver - swop out `<Trans>` elements for i18next.t calls

## Deploy Notes

**New dependencies**:

- `i18next-http-backend` : backend that fetches translations from a defined path


yong-jie (commented):
looks like the `<Trans>` is also used in `LandingGraphicSliver.tsx`. Do you also plan to migrate that to vanilla i18next? we might be able to remove react-i18next if that's the case

LoneRifle:
> looks like the `<Trans>` is also used in `LandingGraphicSliver.tsx`. Do you also plan to migrate that to vanilla i18next? we might be able to remove react-i18next if that's the case

`<Trans>` is being used here to honour the style tags that are found in the resource strings.

yong-jie (approved):

PR #1281, opened 2021-03-08
chore(package-lock): upgrade to v2

LoneRifle:
## Problem

Gitpod now uses npm v7 by default, which uses v2 of the package lock format. This leads to large diffs against our original codebase whenever we `npm install` something while working using gitpod.

## Solution
Move to v2 of the lock file, and recommend the Go team to use npm v7 going forward. npm v6 "tries its best" to work with v2 lockfiles, so there is a degree of backward compatibility. That said, it is generally recommended that developers move to npm v7 for [deterministic builds](https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson) and to benefit from a more [reliable package manager](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/)


liangyuanruo:
From NPM's [blog post](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/):

> We first introduced npm v7 back in May. Today, we are excited to announce the release of npm v7.0.0, which will be shipping with Node.js 15.0.0 next week.

Will there be compatibility issues with the existing Node 14?

> npm 7.0.0. will not be marked as latest; your workflows will not get npm v7.0.0 by default unless you opt in by running npm install -g npm@7 or install Node.js 15.

Has it been marked as latest yet?

LoneRifle:
Gitpod already pairs Node 14 with npm 7 routinely. I suspect that Node only cares for what's in `node_modules`

LoneRifle:
Closing, blocked by #1286

PR #1286, opened 2021-03-09
refactor(qr): replace jsdom and d3 with cheerio

### LINKED ISSUE #1284: Replace jsdom and d3 libraries with leaner libraries such as cheeriojs
Currently, the jsdom and d3 libraries have been imported to generate HTML for the QRCodeService module. 

However, we only use a small subset of the functionalities contained within these libraries.

As such, we could look into alternative replacements to reduce the memory footprint of the client app and decrease loading times.

One suggestion is cheeriojs, a Javascript library that can parse, traverse and manipulate HTML, but does not implement  features such as visual rendering, applying CSS, loading external resources, or executing JavaScript. I think it's a close match for our needs.

Metrics that can be used to measure the effectiveness of this proposal include the client app load time as well as the time taken to generate QR Codes.

orbitalsqwib:
## Problem

Currently, the jsdom and d3 libraries have been imported to generate HTML for the QRCodeService module.
However, we only use a small subset of the functionalities contained within these libraries.

## Solution

Replaced jsdom and d3 libraries with cheerio to reduce client app memory footprint.
Closes #1284.

**New dependencies**:

- `cheerio` : Lightweight markup parser that can traverse and manipulate HTML.

**New dev dependencies**:

- `@types/cheerio` : Supplies types for cheerio


liangyuanruo:
Do you have any metrics to back up this PR?

orbitalsqwib:
as of now i haven’t evaluated the metrics, but potentially they would be the application load speed and the qr code generation speed

orbitalsqwib:
Update: I realised I can't evaluate the application load speed as the files are stored locally on my device, however I checked the bundle sizes for cheerio, jsdom and d3 on Bundlephobia and found that cheerio's bundle footprint is about 323.8kB minified, while jsdom and d3 are 2.9MB and 278.8kB minified respectively.

After gzipping, cheerio is at 88.5kB while jsdom and d3 are 726kB and 86.7kB respectively. Hence, there should theoretically be a 89% reduction in the bundle footprint for this code segment specifically.

liangyuanruo:
> Hence, there should theoretically be an 89% reduction in the bundle footprint for this code segment specifically.

Actually, there is zero change in bundle size because the QR code service is on the server.

LoneRifle:
@liangyuanruo another reason to replace jsdom and d3 is that of dependency management - cheerio is lighter and has fewer dependencies than jsdom+d3 (7 vs 23 + 30), and despite having a limited API relative to d3, is still suitable enough for our usecase of generating SVGs. Given that we only use a tiny subset of the incumbents (ie, jsdom to create a DOM and d3.select to manipulate the DOM), it makes sense to have the libraries replaced, so that generally, we minimise the number of dependency updates that we may get but not need, particularly from d3.

liangyuanruo:
Still doesn't sound like a very strong reason to change battle-tested server-side code, but if you insist... I see that we have unit tests in `qrcode.test.ts` so I don't suppose this will break anything.

LoneRifle:
> btw can help to remove the unnecessary typecast in the unit tests in `qrcode.test.ts`?
> 
> ```ts
> const buffer = (await qrCodeService.createGoQrCode(
>         testUrl,
>         ImageFormat.PNG,
>       )) as Buffer
> ```

@orbitalsqwib can get to that once he's subjected the codebase to k6 testing

liangyuanruo:
sure thing, thanks!

src/server/modules/qr/services/QrCodeService.ts
```diff
@@ -51,76 +48,79 @@ export class QrCodeService implements interfaces.QrCodeService {
...
+        .attr('x', `${textLocationX}`)
+        .attr('y', `${textLocationY + i * FONT_SIZE * LINE_HEIGHT}`)
+        .attr('text-anchor', 'middle')
+        .attr('font-family', 'sans-serif')
+        .attr('font-weight', '400')
+        .attr('font-size', `${FONT_SIZE}px`)
+        .text(lines[i])
+
+      svg.append(`${linkDOM('body').html()}`)
+    }
```
LoneRifle, line 120: So how about..
```suggestion
    lines.forEach((line, i) => {
      const linkDOM = cheerio.load('')
      linkDOM('body').append('<text></text>')
      linkDOM('text')
        .attr('x', `${textLocationX}`)
        .attr('y', `${textLocationY + i * FONT_SIZE * LINE_HEIGHT}`)
        .attr('text-anchor', 'middle')
        .attr('font-family', 'sans-serif')
        .attr('font-weight', '400')
        .attr('font-size', `${FONT_SIZE}px`)
        .text(line)

      svg.append(`${linkDOM('body').html()}`)
    })
```

LoneRifle (changes requested):
See this and the other request to remove the test

LoneRifle:
Note that this PR also upgrades package-lock to v2.

PR #1293, opened 2021-03-11
fix(LongUrlEditor): strip protocol from long URL

### LINKED ISSUE #1092: When pasting a link in update link field in the drawer, https:// is not stripped
**Describe the bug**
On link creation modal, when a long link is pasted into the original link field, the https:// will be stripped, however this behaviour is not consistent when updating the original link in the drawer.

**To Reproduce**
1. Select a link
2. Paste the new full url into the Original Link field.
3. See that the https:// still present.
Screenshot 2020-12-21 at 6 14 50 PM

zwliew:
## Problem

When editing the long URL in `LongUrlEditor`, the protocol is not automatically stripped. This behaviour does not match that of the link creation modal.

Closes #1092

## Solution

**Bug Fixes**:

- Strip the protocol from the long URL before updating the state of the `LongUrlEditor`.

## Tests

**Protocol should be stripped when editing the long URL in `LongUrlEditor`**:
1. Deploy the change
2. Change the long URL of an existing link to one prefixed with `http://` or `https://`
3. The prefix should be automatically stripped

### DIFF
--- a/src/client/user/components/Drawer/ControlPanel/widgets/LongUrlEditor.tsx
+++ b/src/client/user/components/Drawer/ControlPanel/widgets/LongUrlEditor.tsx
@@ -27,7 +27,9 @@ export default function LongUrlEditor() {
       leading={
         <PrefixableTextField
           value={editedLongUrl}
-          onChange={(event) => setEditedLongUrl(event.target.value)}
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
lgtm

PR #1294, opened 2021-03-11
refactor(pages): drop govtech-sgds via jsdelivr

LoneRifle:
## Problem

The transition and 404 pages currently have a dependency on
govtech-sgds, delivered via cdn.jsdelivr.com. While convenient, this
needs an entry on our CSP to allow jsdelivr, which malicious actors
might take advantage of.

## Solution

- Replace govtech-sgds with CSS specific to the masthead, and the
  Lion Head Symbol introduced earlier
- Remove jsdelivr entries in CSP

### DIFF
--- a/public/assets/not-found-page/styles/not-found-page.css
+++ b/public/assets/not-found-page/styles/not-found-page.css
@@ -40,6 +40,36 @@ img {
   padding: 6px 30px;
 }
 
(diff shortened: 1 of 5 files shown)


yong-jie (approved):
good idea, lgtm

PR #1302, opened 2021-03-12
build(webpack): prepare assets for edu build

orbitalsqwib:
## Problem

The upcoming edu build will be a reskin of the current go.gov.sg build and will need to serve a separate set of assets apart from the current set. However, the current client source code uses file paths that use string literals that cannot be changed to  load another set of file paths at runtime. 

In addition, the current assets within the project are scattered across multiple folders which pose an organizational and technical challenge to swap out for the new assets.

## Solution

By using Webpack's `resolve.alias` configuration, we can switch the client assets being emitted to the build folder at compile time, thereby swapping in the appropriate assets for the build.

On top of that, a reorganization of the asset folders have been carried out to consolidate the assets into a single directory, with aliases being used to keep file paths readable.

**Features**:

- dynamically loads assets at compile time for the client application
- selects appropriate asset to use for server-side scripts at run time

**Improvements**:

- reorganized and consolidated assets for client application

## Deploy Notes

- `ASSET_VARIANT` : decides which set of assets should be used during the build process and during runtime
    - `edu` : uses the assets for the edu build
    - defaults to use the gov build assets if empty or unrecognized


webpack.config.js
```diff
@@ -14,6 +14,9 @@ const requiredSentryEnvVar = [
   process.env.SENTRY_URL,
 ]
 
+const assetVariant = process.env.ASSET_VARIANT || 'gov'
+const assetResolveDir = assetVariant === 'edu' ? 'assets/edu/' : 'assets/gov/'
```
LoneRifle, line 18: Is the trailing `/` necessary? if we can drop that without affecting the build the favicon path in HtmlWebpackPlugin would be cleaner-looking

LoneRifle (changes requested):
lgtm. There's one more bit you have to figure out - fetching the appropriate locales file. you will need to duplicate `public/locales/en/translation.json` and have the appropriate path selected by the client at runtime

webpack.config.js
```diff
@@ -69,17 +93,13 @@ module.exports = () => {
...
-          'og:title': 'Go.gov.sg',
-          'og:type': 'article',
-          'og:description': 'The official Singapore government link shortener',
-          'og:image':
-            'https://s3-ap-southeast-1.amazonaws.com/gosg-public/gosg-landing-meta.jpg',
-        },
+        meta: metaVariant,
+      }),
+      new webpack.DefinePlugin({
+        ASSET_VARIANT: JSON.stringify(process.env.ASSET_VARIANT),
```
LoneRifle, line 102:
```suggestion
        'process.env.ASSET_VARIANT': JSON.stringify(process.env.ASSET_VARIANT),
```
And then do the suggested change mentioned earlier

src/client/app/i18n.ts
```diff
@@ -2,12 +2,23 @@ import i18next from 'i18next'
 import Backend from 'i18next-http-backend'
 import { initReactI18next } from 'react-i18next'
 
+declare let ASSET_VARIANT: string
+const assetVariant = ASSET_VARIANT || 'gov'
```
LoneRifle, line 6: Surely...
```suggestion
const assetVariant = process.env.ASSET_VARIANT || 'gov'
```

And then key the value in DefinePlugin as `process.env.ASSET_VARIANT`?

LoneRifle (changes requested):
lgtm otherwise

LoneRifle (approved):

PR #1307, opened 2021-03-16
refactor(server-views): refactor public assets and styles for dynamic server views

orbitalsqwib:
## Problem

Currently, our server view pages use static assets, which need to be converted to dynamic assets based on the `ASSET_VARIANT` env var.

Part of work for #777 

## Solution

- reorganized public assets and styles to prepare them for path switching
- implemented dynamic loading of assets at render time for server views
- implemented toggling of mastheads for server views in edu build


src/server/config.ts
```diff
@@ -198,3 +198,5 @@ export const dbPoolSize = Number(process.env.DB_POOL_SIZE) || 40
 export const sentryDns: string | undefined = process.env.SENTRY_DNS
 
 export const assetVariant: string = process.env.ASSET_VARIANT || 'gov'
+export const displayHostname =
+  assetVariant === 'gov' ? 'Go.gov.sg' : 'Edu.edu.sg'
```
LoneRifle, line 202:
```suggestion
  assetVariant === 'gov' ? 'Go.gov.sg' : 'For.edu.sg'
```

src/server/views/transition-page.ejs
```diff
@@ -6,42 +6,39 @@
...
     <meta charset="UTF-8">
-    <title>Go.gov.sg</title>
-    <link href="/assets/shared/shared-styles.css" rel="stylesheet">
-    <link href="/assets/shared/colours.css" rel="stylesheet">
+    <title><%- displayHostname %></title>
+    <link href="/styles/shared/shared-styles.css" rel="stylesheet">
+    <link href="/<%- assetVariant %>/assets/styles/colours.css" rel="stylesheet">
     <link href="https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@400;500;700&display=swap" rel="stylesheet">
-    <link href="/assets/transition-page/styles/transition-page.css" rel="stylesheet">
+    <link href="/styles/transition-page/transition-page.css" rel="stylesheet">
```
LoneRifle, line 13: By convention, `/assets/` should be the top-level dir

LoneRifle (changes requested):
lgtm otherwise. take care when writing commit messages:
- subject line should be no more than 72 chars, 50 char limit generally preferable
- things done should be written in the imperative, so use present tense, not past

See Chris Beam's article for more information

LoneRifle (approved):

PR #1308, opened 2021-03-16
fix(url-validation): use error response from server

### LINKED ISSUE #1175: Show appropriate error message when user try to shorten a go.gov.sg link
Currently when a user tries to shorten a go.gov.sg link, it will show a "Server error" message. This error message is not easily understandable and can lead to user confusing that there might be some other problems with their account.

A more accurate error message should be "You cannot shorten a go.gov.sg link"

Screenshot 2021-02-09 at 5 45 54 PM

orbitalsqwib:
## Problem

Server errors are currently hard-coded to display "Server Error",  instead of the actual error response returned from the server.

Closes #1175

## Solution

Fixed the error handling code to display the actual error message from the response.


LoneRifle (changes requested):
As discussed - this might disclose issues with the service that we might not want to disclose. We might be able to handle the validation error more specifically; look at code blocks above to figure out the handling

LoneRifle (approved):

PR #1310, opened 2021-03-17
fix(masthead): add default asset variant for masthead

orbitalsqwib:
## Problem

Masthead does not show up if `ASSET_VARIANT` env var is not set

## Solution

Added defaults to use 'gov' as default value for `ASSET_VARIANT`


LoneRifle (approved):
lgtm

PR #1317, opened 2021-03-17
fix(qr-fonts): implement ibm plex sans for qrcode url

### LINKED ISSUE #433: Avoid confusion between I, 1 and I in rendered URL
I was recently sent a go.gov.sg url. The last character seemed to be either a 1 or a I. (I tried both but both failed. I guess there's another problem in there somewhere but let's deal with this first.) I think it is quite annoying for users to have to try both characters.

Could the code just avoid using "1" and "I" in the generated URL? 

Granted, the poster I received had a QR code I could scan but it just adds additional complexity for the user in some cases. In my case, I was accessing the file from a computer and could not scan the QR code. Making the generated URL clickable in the QR code could be workable too I guess.

gov url

orbitalsqwib:
## Problem

Currently, the font used in qr-code generation does not sufficiently distinguish between (capital i) `I`,`1` and (lowercase L) `l`.

Closes #433

## Solution

Implement IBM Plex Sans as the font for qr code generation instead of the default sans-serif font.

- added appropriate styles to qr generation in `QRCodeService` to render IBM Plex Sans Regular
- added wget command in Dockerfile to install IBM Plex Sans Regular in the container instance to give
sharp access to the font

### Before
https __go gov sg_il1il1i1l1i1i1i1i1111ll1l111-1

### After
https __go gov sg_il1il1i1l1i1i1i1i1111ll1l111

## Deploy Notes

Note that IBM Plex Sans is sourced from Google Fonts when exporting to SVG, while sharp uses a container-installed instance of the same font from IBM's Plex typeface repository ([link](https://github.com/IBM/plex)).

As such, modifications have been made to the appropriate Dockerfiles to get the container instance to download the appropriate font and subsequently install via `fc-cache -f`, which force rebuilds the font configurations on Alpine Linux.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -9,6 +9,10 @@ RUN sudo apt-get update \
 RUN sudo mkdir -p /docker-entrypoint-initaws.d
 RUN sudo chown gitpod /docker-entrypoint-initaws.d
 
+# Installs IBMPlexSans-Regular.ttf for QRCodeService.
+RUN sudo wget https://github.com/IBM/plex/blob/master/IBM-Plex-Sans/fonts/complete/ttf/IBMPlexSans-Regular.ttf?raw=true -O /usr/share/fonts/truetype/IBMPlexSans-Regular.ttf
+RUN sudo fc-cache -f
+
 USER gitpod
 
 ENV NODE_ENV=development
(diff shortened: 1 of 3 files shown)


LoneRifle (approved):
lgtm - @NatMaeTan can you give it a once over?

LoneRifle:
I see the following error when trying to spin up gitpod:
Screenshot_20210318-090656

PR #1326, opened 2021-03-18
build(edu-pipeline) implement ci build pipeline for edu

orbitalsqwib:
## Problem

Currently, there is no pipeline for the edu variant of go.gov.sg to be built and deployed.

Part of work #777 

## Solution

- Implemented `build-edu` and `deploy-edu` actions for continuous integration
- Implemented build args for `ASSET_VARIANT` in Dockerfile so that production builds can inject the env var into the container properly


LoneRifle (approved):
lgtm

PR #1337, opened 2021-03-22
refactor(redux-actions): refactor redux types to use generics

### LINKED ISSUE #938: Extract Redux actions patterns with generics
_Blocked by #937_

Duplicate boilerplate code exists where Redux actions are present.

For example, in `src/client/user/actions/types.ts` each action is duplicating the same `{ type, payload }` pattern:

```typescript
export type SetLastCreatedLinkAction = {
  type: typeof SET_LAST_CREATED_LINK
  payload: string
}

export type WipeUserStateAction = {
  type: typeof WIPE_USER_STATE
}

(... 22 more actions)
```

Extract the repeated `{ type, payload }` pattern via using [Generics](https://www.typescriptlang.org/docs/handbook/generics.html). For example:

```typescript
interface ReduxAction<T extends string, P> {
    readonly type: T
    readonly payload?: P
}

enum UserActions {
    SomeAction = 'SomeAction',
    NoPayloadAction = 'NoPayloadAction'
}

type SomeAction = ReduxAction<typeof UserActions.SomeAction, string>
type NoPayloadAction = ReduxAction<typeof UserActions.NoPayloadAction, void>
```

By addressing this issue, not only will we adhere to the DRY principle, but it will also help ensure that all Redux actions are typed correctly throughout the codebase.

orbitalsqwib:
## Problem

Duplicate boilerplate code exists where Redux actions are present. By addressing this issue, not only will we adhere to the DRY principle, but it will also help ensure that all Redux actions are typed correctly throughout the codebase.

Closes #938

## Solution

- Implemented generic redux actions interfaces through intersection type extensions
- Refactored redux type definitions to use generic interfaces

### DIFF
--- a/src/client/app/actions/types.ts
+++ b/src/client/app/actions/types.ts
@@ -23,4 +23,7 @@ export interface HasPayload<P> {
   payload: P
 }
 
+export type ReduxPayloadAction<T extends string, P> = ReduxAction<T> &
+  HasPayload<P>
+
 export type AllThunkDispatch = ThunkDispatch<GoGovReduxState, void, AllActions>
--- a/src/client/app/actions/types.ts
+++ b/src/client/app/actions/types.ts
@@ -15,4 +15,12 @@ export type AllActions =
   | HomeActionType
   | DirectoryActionType
 
+export interface ReduxAction<T extends string> {
+  type: T
+}
+
+export interface HasPayload<P> {
+  payload: P
+}
+
 export type AllThunkDispatch = ThunkDispatch<GoGovReduxState, void, AllActions>
--- a/src/client/login/actions/types.ts
+++ b/src/client/login/actions/types.ts
@@ -1,3 +1,5 @@
+import { HasPayload, ReduxAction } from '../../app/actions/types'
+
 export const GET_OTP_EMAIL_SUCCESS = 'GET_OTP_EMAIL_SUCCESS'
 export const GET_OTP_EMAIL_PENDING = 'GET_OTP_EMAIL_PENDING'
 export const GET_OTP_EMAIL_ERROR = 'GET_OTP_EMAIL_ERROR'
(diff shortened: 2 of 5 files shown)


src/client/app/actions/types.ts
```diff
@@ -15,4 +15,12 @@ export type AllActions =
   | HomeActionType
   | DirectoryActionType
 
+export interface ReduxAction<T extends string> {
+  type: T
+}
+
+export interface HasPayload<P> {
+  payload: P
+}
```
liangyuanruo, line 24: Suggest defining a single type to save on the repeated use of ampersand, since the keys are always used together.

```typescript
export type ReduxAction<T extends string, P> = {
  type: T,
  payload: P
}
```

orbitalsqwib: Initially I tried using this manner of implementing the interface but ran into several issues with the compiler complaining about the void parameters.

In addition, I tried implementing optional parameters and default parameters but those also didn’t work properly.

In the end the current implementation was suggested by Alwyn as a means to accommodate types which had no payload

LoneRifle: So since the comment is about the &, I would suggest having a second type that just intersects the two together, ie,
```js
export type ReduxActionWithPayload<T extends string, P> = ReduxAction<T> & HasPayload<P>
```

liangyuanruo: > Initially I tried using this manner of implementing the interface but ran into several issues with the compiler complaining about the void parameters.

Very strange. I'd wager that was a strange quirk, and you'll actually be able to get it working if you try again.

orbitalsqwib: I tested the single type method again and ran into the same problem. It seems that Typescript expects every parameter within the suggested interface above to be required (truthy), thus providing it with a falsy value such as void (which evaluates to undefined) will throw an error:

`Property 'payload' is missing in type '{ type: "SET_INITIAL_STATE"; }' but required in type 'ReduxActionTest<"SET_INITIAL_STATE", void>`

orbitalsqwib: I've implemented LoneRifle's solution in the latest commit

src/client/user/actions/types.ts
```diff
@@ -35,143 +37,141 @@ export enum UserAction {
   SET_FILE_UPLOAD_STATE = 'SET_FILE_UPLOAD_STATE',
 }
 
-export type SetUrlUploadStateAction = {
-  type: UserAction.SET_URL_UPLOAD_STATE
-  payload: boolean
-}
+export type SetUrlUploadStateAction = ReduxPayloadAction<
+  typeof UserAction.SET_URL_UPLOAD_STATE,
```
LoneRifle, line 41: is this correct? you're changing the type from `UserAction.SET_URL_UPLOAD_STATE` to `typeof UserAction.SET_URL_UPLOAD_STATE`?

src/client/login/actions/types.ts
```diff
@@ -9,49 +11,39 @@ export const SET_EMAIL_VALIDATOR = 'SET_EMAIL_VALIDATOR'
...
-export type GetOtpEmailSuccessAction = {
-  type: typeof GET_OTP_EMAIL_SUCCESS
-  payload: string
-}
+export type GetOtpEmailSuccessAction = ReduxPayloadAction<
+  typeof GET_OTP_EMAIL_SUCCESS,
+  string
+>
+
+export type GetOtpEmailPendingAction = ReduxAction<typeof GET_OTP_EMAIL_PENDING>
```
yong-jie, line 19: not sure if relevant, but [this](https://stackoverflow.com/questions/44884838/what-is-const-type-pattern-in-typescript) mentions that we can just use `type GET_OTP_EMAIL_PENDING = 'GET_OTP_EMAIL_PENDING'` in line 2 for example and we wont have to use typeof here anymore

LoneRifle: If the purpose of doing so however is to just avoid `typeof` in the type parameter, that's fairly redundant

LoneRifle (approved):

src/client/login/actions/types.ts
yong-jie: i think `typeof GET_OTP_EMAIL_PENDING` evaluates to `string`, whereas `type GET_OTP_EMAIL_PENDING = "GET_OTP_EMAIL_PENDING"` is a string that can only be that value. So it depends on how specific we'd like this `ReduxAction<T>` to be

PR #1338, opened 2021-03-22
refactor(babelrc): migrate babelrc to babel.config.json

### LINKED ISSUE #1264: Use project-wide babel.config.json configured to target supported browsers
It is good practice in Babel 7.x to use a project-wide `babel.config.json` instead of `.babel.rc`, which is a relic of Babel 6.x (we are on 7.x).

The configuration file should also explicitly specify the supported application browsers to ensure longer-term compatibility of JS build.

See https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading, as well as the FormSG codebase for an example.

orbitalsqwib:
## Problem

It is good practice in Babel 7.x to use a project-wide babel.config.json instead of .babel.rc, which
is a relic of Babel 6.x (we are on 7.x).

Closes #1264

## Solution

- implemented babel.config.json based on FormSG's current configuration


babel.config.json
```diff
@@ -0,0 +1,15 @@
...
+      "corejs": 3,
+      "useBuiltIns": "entry",
+      "targets": {
+        "browsers": ["IE 11", "not dead"]
+      }
+    }],
+    "@babel/preset-react"
+  ],
+  "plugins": [
+    "@babel/plugin-proposal-class-properties"
```
liangyuanruo, line 13: Consider adding some more plugins for completeness:
1. @babel/plugin-proposal-private-methods
2. @babel/plugin-proposal-decorators
3. @babel-plugin-proposal-optional-chaining

babel.config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "presets": [
+    ["@babel/preset-env", {
+      "corejs": 3,
+      "useBuiltIns": "entry",
```
liangyuanruo, line 5: If you take a look at the [docs](https://babeljs.io/docs/en/babel-preset-env#usebuiltins), you will notice that the following lines need to be available in the top-level application entrypoint for `entry` mode:

```typescript
import "core-js/stable"
import "regenerator-runtime/runtime"
```

babel.config.json
```diff
@@ -0,0 +1,15 @@
...
+      "targets": {
+        "browsers": ["IE 11", "not dead"]
```
liangyuanruo, line 7: you should have a quick clarification with the team on whether it is still necessary to support IE11, given that official support for it ends in Aug 2021.

liangyuanruo (changes requested):
This can be a sensitive part of the codebase, so make sure to test thoroughly (with IE11 if necessary).

babel.config.json
orbitalsqwib: Adding these in next commit, thanks for the catch!

babel.config.json
orbitalsqwib: The team decided to keep supporting IE 11 since there are still members of the public using the transition page, and there may be officers who may be using Go through their GSIBs.

orbitalsqwib:
I've tested the latest commit with Chrome, Firefox, Safari and also verified that at minimum the transition and landing pages for IE 11 work properly (though the transition page has some styling issues that leave the background completely white).

liangyuanruo:
You also need to test functionality, since this PR is modifying the JS transpilation process.

orbitalsqwib:
As for functionality, I’ve done the end to end tests for the platforms specified in my previous comment with the exception of IE11, as there were some issues with the VM i was using

LoneRifle:
snyk test errored out, not an actual failure

babel.config.json
LoneRifle: I feel that is premature, given the general direction for this codebase. Let's hold off including this for now

LoneRifle (approved):

PR #1344, opened 2021-03-23
fix(webpack): fix webpack for IE 11 support

orbitalsqwib:
## Problem

Currently, webpack 5 builds to target es6 by default, which breaks IE 11 since it only supports up to es5. Additionally, webpack 5 drops quite a few polyfills, which triggers warnings to provide fallbacks even if unused. There is also a transpilation issue with react components where IE would break if the transpilation order was incorrect.

## Solution

- Changed webpack build target to es5
- Added fallbacks for unused polyfills that webpack 5 drops from webpack 4
- Specified transpilation order to fix an issue with IE, where it would break react components otherwise

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -39,7 +39,14 @@ const metaVariant = assetVariant === 'edu' ? eduMetaTags : govMetaTags
 
 module.exports = () => {
   const jsBundle = {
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
lgtm

PR #1347, opened 2021-03-24
feat(maildev): log OTP in console through maildev in dev mode

orbitalsqwib:
## Problem

Currently, the implementation for maildev does not include support for Gitpod, as maildev is run on another container instance and thus cannot be ported to Gitpod as is. Furthermore, the development workflow is slightly slowed down due to having to get the OTP manually from maildev's visual interface through `localhost:1080`.

## Solution

By logging the OTP in the console, developers should be able to retrieve the OTP there and log into the application without having to open `localhost:1080`. In addition, this pull request also implements maildev for Gitpod through running a maildev instance on the same container instance.

- logs the OTP in console for logins through maildev during development mode for convenience
- implements maildev support for gitpod
- refactored transport option configurations for server side config to reuse some SES variables for maildev

### Environmental Variables
- Reused `SES_HOST` and `SES_PORT` for configuring maildev host and smtp ports.
    - `docker-compose.yml` sets `SES_HOST` and `SES_PORT` to `maildev` and `25` respectively
    - `gitpod.Dockerfile` sets `SES_HOST` and `SES_PORT` to `localhost` and `1025` respectively

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -8,8 +8,8 @@ tasks:
   - before: export ACCESS_ENDPOINT=`gp url 4566`
     init: npm install
     command: /tmp/wait-for-it.sh localhost:4566 -t 0 -- ./init-localstack.sh && echo "Run npm run docker-dev to begin"
-  - init: npm install maildev -g
-    command: maildev --verbose
+  - before: npm install maildev -g
+    command: echo "maildev.on('new', ({ html }) => console.log('The login OTP is ' + html.match(/\d{6}/)[0]))" >> $(which maildev) && maildev --verbose
 ports:
   # postgres
   - port: 5432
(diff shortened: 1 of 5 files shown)


.gitpod.yml
```diff
@@ -8,6 +8,8 @@ tasks:
   - before: export ACCESS_ENDPOINT=`gp url 4566`
     init: npm install
     command: /tmp/wait-for-it.sh localhost:4566 -t 0 -- ./init-localstack.sh && echo "Run npm run docker-dev to begin"
+  - init: npm install maildev -g
+    command: maildev --verbose
```
LoneRifle, line 12: Hang on, we can do that? If so, can you figure a way to supply the CMD arg to the docker image on local machine?

LoneRifle (approved):
lgtm

PR #1375, opened 2021-04-07
chore(e2e): skip frontend js errors

LoneRifle:
Avoid spurious JavaScript errors thrown by the frontend, which we have 
observed in commits up to release 1.53 and seem unrelated to the actual
functioning of the front and backend


yong-jie:
Might be worth merging #1382 in favor of this if it can reliably eliminate the frontend js errors we're facing.

PR #1438, opened 2021-05-13
fix(mergify): correct dependabot title regex

LoneRifle:
## Problem and Solution

Correct the regex used to identify dependabot PRs for merging


liangyuanruo (approved):
Thanks for fixing this!

PR #1480, opened 2021-05-18
fix/privacy: remove exposure of long URLs from GoDirectory

liangyuanruo:
## Problem

Exposing the long URL on GoDirectory may reveal sensitive URLs if users are not savvy.

## Solution

Remove the long URL from both the UI and the `/search` API.

**Improvements**:

- Users cannot use GoDirectory to mine links for inactive links

## Screenshots

image


yong-jie (approved):
lgtm!

PR #1481, opened 2021-05-18
feat/warning: Warn users that their links will be indexable by search engines

liangyuanruo:
## Problem

Users may upload sensitive content that ends up being discovered by other users on the directory, or indexed by search engines.

## Solution

Add a suitable message to the creation modal for both link and file creation, that has been endorsed by @lisatjide .

**Improvements**:

- Users should (hopefully!) think twice about making sensitive documents public.

## Screenshots

LInk creation
image

File creation
image


lisatjide (approved):
Looks good to me!

PR #1649, opened 2021-07-28
build(deps): relock dependencies to fix security vulnerabilities

yong-jie:
Fixes dependency problems highlighted by snyk. Also fixed problems highlighted by npm audit, all this done by removing package-lock and rerunning npm install.

Some dependencies had breaking changes, which caused the typescript compiler to fail. This PR also contains the fixes for them:

- `express-session` requires the `req.session` object to be manually typed. I did this using declaration merging.
- Update `connect-redis` to a newer version that is compatible with `express-session`'s breaking change. 
- MUI deprecated a prop in the `TablePagination` component. I checked the type declaration for information and used the new one
- MUI also changed the class name of some components, which broke e2e tests which were selecting them. I changed the selectors to use the new `MuiCollapse-root` element.


LoneRifle (approved):
lgtm

PR #1651, opened 2021-07-29
feat(user-route): validate type of req.body.email

yong-jie:
## Problem

`req.body` is based on user-controlled input and so `req.body.email` could be crafted with a custom .trim() property. Calling .trim() before validating the variable's type is unsafe.

## Solution

Enforce that the type is a string.


LoneRifle (approved):
lgtm

PR #1741, opened 2021-11-08
fix/audit: Periodic security patch for the application

liangyuanruo:
Periodic security patch for the GoGovSG codebase, via `npm audit fix`.

The patch did not upgrade major semvers, on closer examination those were related only to `devDependencies` such as `jest` and `webpack-dev-server`.


LoneRifle (approved):
Lgtm

PR #1755, opened 2021-12-23
feat(lambda): create lambda for migration of url(s)

wanlingt:
## Problem

Transfer of ownership is currently being done manually and directly in the prod database, which exposes us to unnecessary risks. Further details can be found in the Orphan Links Management [Design Doc](https://docs.google.com/document/d/13WgLctEXyoWiMsnpvx49GzR4WamDQwnSnJg7btODRnM/edit?usp=sharing)

## Solution

Create an AWS lambda for both types of link transfer operations:
- `migrate_url_to_user` - to transfer a single link to a new email address (all lowercase)
- `migrate_user_links` - to transfer all links belonging to an account to another account (all lowercase)

The lambda handler executes the relevant SQL function, using the arguments provided in the `event` object.

**Details**:

- The `DATABASE_URL` env var is provided directly in the lambda's environmental variables configuration, instead of GitHub secrets

## Tests

Invoke the test events in each lambda function by filling in the appropriate parameters, and check that the changes are reflected in the database

## Deploy Notes

Add `serverless deploy` to `deploy` stage for deployment of lambda functions.

Note: `deploy-edu` remains unchanged.

- `SERVERLESS_STAGE` : one of `staging`, `production`

**New dependencies**:

- `serverless-plugin-include-dependencies` : This Serverless [plugin](https://www.serverless.com/plugins/serverless-plugin-include-dependencies) makes deployed functions smaller by only adding the modules relevant to each handler


src/server/serverless/lambda-migrate-user-links/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+    await pgClient.connect().then(() => {
+      console.log('Connected')
+    })
+
+    const sqlUDF = fs.readFileSync(sqlScriptPath)
+    const sqlScript = `SELECT migrate_user_links($1, $2)`
+    const values = [fromUserEmail, toUserEmail]
+
+    const queryUDFResult = await pgClient.query(sqlUDF.toString())
```
LoneRifle, line 25: I'm still not very comfortable of the idea of overwriting the function every single time we want to invoke it. Is there no way to avoid this at all?

wanlingt: Thanks for pointing it out, didn't realise it's already stored in the DB - have removed the lines pertaining to the SQL function

LoneRifle (approved):
Lgtm

PR #1794, opened 2022-03-29
build(deps): Upgrade aws-sdk from 2.977.0 to 2.1102.0

yong-jie:
Verified that there are no breaking changes on staging


gweiying (approved):
lgtm

PR #1800, opened 2022-04-01
Fix/url history defaults

gweiying:
## Problem

Lambda functions for migrating links were not running successfully on `health-staging-db` as the `description` column in `url_histories` tables had a constraint that enforced non-null values but did not set a default value.

Related to this, the two SQL migration functions did not include `description` as an indexed column from `url` table when inserting the link into the `url_histories` table. Hence, when a new migration attempted to occur, the `description` in the `url` table was not being copied into the new row in `url_histories` table, so the column defaulted to `null`, violating the non-null constraint. 

## Solution

- Updated SQL functions to include `description` as an indexed column on `url` table when inserting the link into the `url_histories` table
- Added empty string as a default on `description` column in `url_histories` table in database models

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

- [x] Test `migrate-short-link` on `health-staging`
- [x] Test `migrate-user-links` on `health-staging`

## Deploy Notes

#### Code changes
- [x] Fix migration functions to add `description` when migrating url
- [x] Update Sequelize models to add default empty string “” for `description` column in `url_histories` table

#### Update current databases

Health
- [x] Run migration to fix this in `health-staging`
  - [x] ```ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';```
  - [x] Update migration function `migrate-user-links`
  - [x] Update migration function `migrate-shortUrl-to-user`
- [ ] Verify that this does not occur in `health-prod` since the model should be fixed 

Edu
- [x] Run migration to fix this in `edu-staging`
  - [x] ```ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';```
  - [x] Add migration function `migrate-user-links`
  - [x] Add migration function `migrate-shortUrl-to-user`
- [x] Run migration to fix this in  `edu-prod`
  - [x] ```ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';```
  - [x] Add migration function `migrate-user-links`
  - [x] Add migration function `migrate-shortUrl-to-user`

GoGov
- [x] Run migration to fix this in `gogov-staging`
  - ~```ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';```~ (Not necessary in `gogov-staging` and `gogov-prod` as previous migrations have added a default)
  - [x] Update migration function `migrate-user-links`
  - [x] Update migration function `migrate-shortUrl-to-user`
- [x] Run migration to fix this in `gogov-prod`
  - ~```ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';```~ (Not necessary in `gogov-staging` and `gogov-prod` as previous migrations have added a default)
  - [x]  Update migration function `migrate-user-links`
  - [x] Update migration function `migrate-shortUrl-to-user`

___

- [ ] Look into debugging count for number of rows affected

### DIFF
--- a/src/server/db_migrations/20220406_add_description_default_value.sql
+++ b/src/server/db_migrations/20220406_add_description_default_value.sql
@@ -1,0 +1,12 @@
+-- This script was written to fix an incorrect url_histories model.
+-- The description column in the url_histories model had a non-null 
+-- constraint without a default value. As edusg and healthsg tables
+-- were set up with the faulty model, there were errors writing into 
+-- the tables when description was empty.
+-- This script was written for migration on edusg and healthsg.
+
+BEGIN TRANSACTION;
+
+ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';
+
+COMMIT;
--- a/src/server/db_migrations/20220406_add_description_default_value.sql
+++ b/src/server/db_migrations/20220406_add_description_default_value.sql
@@ -7,6 +7,6 @@
 
 BEGIN TRANSACTION;
 
-ALTER TABLE url_histories ADD "description" text NOT NULL DEFAULT '';
+ALTER TABLE url_histories ALTER "description" SET DEFAULT '';
 
 COMMIT;
--- a/src/server/models/url.ts
+++ b/src/server/models/url.ts
@@ -69,6 +69,7 @@ export const UrlHistory = <UrlHistoryStatic>sequelize.define('url_history', {
   description: {
     type: Sequelize.TEXT,
     allowNull: false,
+    defaultValue: '',
   },
 })
 
--- a/src/server/serverless/lambda-migrate-url-to-user/migrate_url_to_user.sql
+++ b/src/server/serverless/lambda-migrate-url-to-user/migrate_url_to_user.sql
@@ -16,6 +16,8 @@
 --                                      isSearchable column
 --   16 Nov  2020 @LoneRifle:           Update function's url_history insertion step to remove 
 --                                      isSearchable column
+--   01 Apr  2021 Alexis Goh:           Update function's url_history insertion step to include
+--                                      description column
 -- =============================================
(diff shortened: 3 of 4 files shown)


yong-jie:
Verified the following info about the `description` columns of each DB:

gosg-staging has '' as default
gosg-production has '' as default
edu-staging doesnt
edu-production doesnt

Did a bit of reading up as to how this happened, I suspect it's due to the differences between the migration file and the sequelize model definition.

Given that the `gov` variant was spun up way earlier, we had to run migration scripts to add the `description` column. [This migration script](https://github.com/opengovsg/GoGovSG/blob/f7ad44b8fda906a60b9991f0efb50fd05d762a76/src/server/db_migrations/20200618_add_description_and_contact_column.sql#L10) declared a `''` as default. However, subsequent variants (`edu`, `health`) directly make use of the sequelize models, which did not have a default value.

yong-jie (commented):
Both the code and strategy looks good to me! i'd just like to propose one more addition to this PR which would be to explicitly commit a migration sql file into the codebase, similar to what has been done in [this directory](https://github.com/opengovsg/GoGovSG/tree/develop/src/server/db_migrations).

Even though the migration script is probably gonna just be a one-liner wrapped in a transaction, there's value in tracking the list of migrations we've done in production, and reduce the risk of executing incorrect commands on the database

gweiying:
> Both the code and strategy looks good to me! i'd just like to propose one more addition to this PR which would be to explicitly commit a migration sql file into the codebase, similar to what has been done in [this directory](https://github.com/opengovsg/GoGovSG/tree/develop/src/server/db_migrations).
> 
> Even though the migration script is probably gonna just be a one-liner wrapped in a transaction, there's value in tracking the list of migrations we've done in production, and reduce the risk of executing incorrect commands on the database

Fixed in 20f898d6a

yong-jie (approved):
lgtm!

PR #1821, opened 2022-04-26
feat: add health variant & centralise assetVariant switch

### LINKED ISSUE #1812: Modify codebase to support 3 asset variants
Part of #1801 

- [ ] Create assets and locales folders for `/health` (temporarily use Gosg's assets)
- [ ] Fix ternary operators to allow for different assets - `assetVariant` currently allows only switching between `'gov'` and `'edu'` variants

orbitalsqwib:
## Problem

In preparation for the new GoGovSG stack targeted at healthcare providers, a new `health` variant needs to be set up to serve the appropriate assets and color themes.

Additionally, the `assetVariant` environment variable controls which variant of the GoGovSG application is being built. However, it is repeatedly imported in multiple places, which may introduce unintended bugs or issues that are hard to track down due to the multiple points of failure.

Closes #1812 

## Solution

This PR duplicates assets from the `gov` and `edu` variants to create placeholder assets for the `health` variant, which should be replaced when the actual assets and design for the variant are finished. Moreover, it also centralizes the `assetVariant` env var import, which is then imported by other files, which ensures that there is only a single point where the environmental variable is read and processed for use in other files.

**Features**:

- Adds the `health` variant to the codebase

**Improvements**:

- Cleans up and centralizes the `assetVariant` environment variable switch
- Removes the default `gov` variant that `assetVariant` defaults to when no value for the environment variable is provided
- Migrates the webpack configuration file to use Typescript

**Bug Fixes**:

- Updates the CI/CD actions and Jest tests to use the appropriate `assetVariant` value

## Tests

- [ ] Check that the variant is deployed to the correct url(s) (staging.for.sg / for.sg)
- [ ] Check that the variant displays the correct assets
- [ ] Check that the variant's features works as expected

## Deploy Notes

Note: This PR moves several dev dependencies to be normal dependencies as listed below:
- `webpack`
- `@sentry/webpack-plugin`
- `clean-webpack-plugin`
- `html-webpack-plugin`

**New dev dependencies**:

- `ts-node` : required to use a typescript webpack configuration file


gweiying (approved):
lgtm! 

Thanks for centralising the switch, it is much cleaner than before. Also, great work on finding and ironing out the bugs in tests and Sentry with the default `gov` variant!

yong-jie:
lgtm! Mind sharing the thought process behind moving webpack stuff (compile/transpile-time) away from dev-dependencies?

minor nitpicking that we can examine in the future:
- ts-ignore in the webpack config file (have we exhausted alternative approaches to get those plugins to work without using ts-ignore?)
-

orbitalsqwib:
> lgtm! Mind sharing the thought process behind moving webpack stuff (compile/transpile-time) away from dev-dependencies?
> 
> minor nitpicking that we can examine in the future:
> 
>     * ts-ignore in the webpack config file (have we exhausted alternative approaches to get those plugins to work without using ts-ignore?)

Replying to this for future documentation:

After a short discussion, these concerns have been resolved, reasons below:

- Moving the webpack plugins used in the typescript-based webpack configuration file was necessary as it would probably have to be transpiled once to commonjs before webpack could use the config, and therefore plugins in dev dependencies might not be "picked up" during the transpile process for webpack to use afterwards - though this could also be a false positive thrown by eslint. (more testing required in this case, but probably better to be safe than sorry)

- Due to incorrect typings, the `HTMLWebpackPlugin`'s `chunkSortMode` does not have a `none` option available, despite it being an option listed in their [documentation](https://github.com/jantimon/html-webpack-plugin). Additionally, the plugins array also restricts the types allowed in it to what was previously specified (e.g: plugins A, B and C are specified in an object first, and D is injected into the object's plugins array later, there will be a type error as the array is typed as accepting only plugins of type `A | B | C`). As this is incorrect behaviour, we have opted to use a `ts-ignore` to remedy the false positive.

PR #1831, opened 2022-05-10
chore: add transactional lock on CREATE OR REPLACE function

gweiying:
## Problem

We run multiple instances in production. On initialisation, each instance attempts to upsert a CREATE OR REPLACE FUNCTION. If multiple instances attempts to update the function on the database at the same instance, a `SequelizeDatabaseError: tuple concurrently updated` error can result, as we observed on 6th May during a release. 

This `tuple concurrently updated` occurs because Postgres does not implement locking for DDL on functions, [detailed here](https://www.postgresql.org/message-id/3473.1393693757%40sss.pgh.pa.us), hence such errors can occur when launching multiple instances that all attempt to upsert the function at once.

More details on this can be found in the writeup [here](https://docs.google.com/document/d/1EyMssJ8wDxu7el0OE3VPeFtLEhfDqpoGszhkYSUhzxM/edit#).

## Solution

We add an advisory transactional lock  with a unique ID on the CREATE OR REPLACE FUNCTION transaction. When an instance boots up and attempts to run that transaction concurrently as another instance, the instance has to acquire the lock before it can run the transaction. This way, the transactions are forced to run one after another and will never attempt to update the function concurrently. 

#### Notes
1) We considered other solutions such as to remove the on-initialization upsert of the function, however this would mean that we will need to deploy any function changes manually if we go forward with this change. After discussing, we eventually decided against it as this could result in possible code drift between the application code on Github, and the database function itself which could be difficult to debug.

2) From [Postgres documentation](https://www.postgresql.org/docs/14/explicit-locking.html#ADVISORY-LOCKS), these locks are not enforced by the database but by the application - `These are called advisory locks, because the system does not enforce their use — it is up to the application to use them correctly.`

**Bug Fixes**:

- Added pg_advisory_xact_lock

## Tests

- [x] Run through manual test on staging environment to verify that application is working as normal
- [x] Load test on staging to verify that database performance is not affected by the lock

### DIFF
--- a/src/server/modules/analytics/repositories/LinkStatisticsRepository.ts
+++ b/src/server/modules/analytics/repositories/LinkStatisticsRepository.ts
@@ -25,8 +25,15 @@ const timeZone = 'Asia/Singapore'
 
 /**
  * This function is used to update the relevant link statistics tables, when called.
+ *
+ * The SQL query acquires a transaction lock to prevent concurrent executions of
+ * the CREATE OR REPLACE FUNCTION. The transaction lock is enforced by lock key, so
+ * the id in the lock has to be a unique (64-bit key) per non-blocking transaction,
+ * i.e. Pg_advisory_xact_lock(uniqueId).
  */
-export const updateLinkStatistics = `CREATE OR REPLACE FUNCTION update_link_statistics (inputShortUrl text, device text)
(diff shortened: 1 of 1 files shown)


yong-jie (approved):
lgtm!

PR #1850, opened 2022-05-31
Chore/health copy

gweiying:
## Solution

Key changes:
- Asset, copy and color changes 
  - Note: We use `dangerouslySetInnerHtml` for the new copy changes as some of the new copies include links
- Add `sanitize-html` library for sanitising html
- Add `date-fns-tz` library for automatically generating copyright year 
- Fix UI states for outline button based on design [here](https://www.figma.com/file/dCK0KWd5qWELIuMuhqA7Lh/Design-System?node-id=148%3A1099)
- Fix console warnings for deprecated arguments and imports on Material UI library


yong-jie (approved):
did a broad glance and generally lgtm. using dangerously set HTML seems okay since we have full control over the copy

PR #1876, opened 2022-07-05
chore: update Sequelize library to 6.21.2 to mitigate the potential S…

thanhdatle:
…QL Injection vulnerability

## Problem

Closes [insert issue #]

## Solution

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


docker-compose.yml
```diff
@@ -50,6 +50,7 @@ services:
     volumes:
       - ./public:/usr/src/gogovsg/public
       - ./src:/usr/src/gogovsg/src
+      - ~/Projects/ContainerLogs:/root/.npm/_logs
```
yong-jie, line 53: this might have been an accidental commit. looks like a volume mount you created to debug yesterday

src/server/modules/user/__tests__/UserController.test.ts
```diff
@@ -107,7 +107,9 @@ describe('UserController', () => {
       const res: any = httpMocks.createResponse()
       res.badRequest = jest.fn()
 
-      urlManagementService.createUrl.mockRejectedValue(new ValidationError(''))
+      urlManagementService.createUrl.mockRejectedValue(
+        new ValidationError('', []),
+      )
```
yong-jie, line 112: could you share what API changes were made in sequelize that necessitated this change?

package.json
```diff
@@ -94,7 +94,7 @@
     "reflect-metadata": "^0.1.13",
     "regenerator-runtime": "^0.13.8",
     "sanitize-html": "^2.7.0",
-    "sequelize": "^6.6.5",
+    "sequelize": "^6.21.2",
```
yong-jie, line 97: i see that there's a [breaking change](https://github.com/sequelize/sequelize/issues/14519) related to the sequelize injection vulnerability. give me a minute to quickly check if those changes affect our codebase as well

yong-jie: okay looks like we dont use `:` replacements in our codebase, so we should be fine

docker-compose.yml
thanhdatle: This is intentional actually, the folder will be created when we run the docker-compose, basically, it will help with our debugging moving forward.

src/server/modules/user/__tests__/UserController.test.ts
thanhdatle: This happens because, the ValidationError constructor is expecting 2 parameters instead of 1.
`constructor(message: string, errors: ValidationErrorItem[], options?: ErrorOptions);`

docker-compose.yml
yong-jie: the `~/Projects/ContainerLogs` part seems specific to your computer, or am i misunderstanding this?

thanhdatle: on a second thought, I changed to ./.logs folder and updated the gitignore file. I think this looks better.

yong-jie (approved):
lgtm!

PR #1882, opened 2022-07-12
fix(lambda): return correct number of migrated links from lambda functions

### LINKED ISSUE #1806: Migration functions does not return accurate count of affected rows
**Describe the bug**

The functions to migrate links from one user to another currently do not return the accurate number of affected rows. For exampl, for `migrate-user-links`, the number of affected rows always returns as 1 even if more than 1 row has been transferred.

Screenshot 2022-04-06 at 4 33 38 PM

**Expected outcome**

Functions should return the accurate number of affected rows.

halfwhole:
## Problem

Currently, the lambda functions to migrate links between users always returns the number of rows affected as 1, regardless of the actual number of links transferred.

Closes #1806 

## Solution

- `migrate_user_links` has been updated to return the actual number of rows affected when updating the `urls` table
- As for `migrate_url_to_user`, it might not be necessary to return the number of rows affected, because if not for an error, there should always be exactly 1 link transferred. Is it ok to just remove this then?

## Tests

- [x] Test `migrate_user_links` on staging
- [x] Test `migrate_url_to_user` on staging

~~I've tested this locally by running the javascript code on node (without using aws lambda, which I'm not sure how to use yet), and locally it seems to work well~~

EDIT: have tested invoking the lambda functions both locally and on staging with the help of @yong-jie and it looks good

### DIFF
--- a/src/server/serverless/lambda-migrate-url-to-user/index.js
+++ b/src/server/serverless/lambda-migrate-url-to-user/index.js
@@ -16,13 +16,11 @@ async function handler(event) {
     const sqlScript = `SELECT migrate_url_to_user($1, $2)`
     const values = [shortUrl, toUserEmail]
 
-    const { rowCount } = await pgClient.query(sqlScript, values)
+    await pgClient.query(sqlScript, values)
 
     pgClient.end().then(() => console.log('Disconnected'))
 
-    statusMsg = `URL successfully migrated. ${JSON.stringify(
-      rowCount,
-    )} rows affected`
+    statusMsg = 'URL successfully migrated.'
   } catch (err) {
     console.log(err)
     pgClient.end()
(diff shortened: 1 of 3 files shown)


yong-jie:
> `migrate-url-to-user` will alwas only have 1 link transferred. Is it ok to just remove this then?

yup sure thing, we can remove it! 👍 

Just for future consideration: in general it probably isn't very harmful to be overly-verbose. In an unlikely scenario where something goes wrong with the DB function, and it transfers more than one link at a time, logging the number returned would have helped us detect the problem early.

yong-jie (approved):
lgtm. tested together with @halfwhole and verified that it works on go-staging. We noted that the DB function does not get deployed automatically with the rest of the CI/CD pipeline, and therefore we will need to manually update the functions prior to release.

PR #1885, opened 2022-07-14
chore: update readme with for.sg details

halfwhole:
## Problem

The readme covers only the gov and edu setups, but not health

## Solution

Updated the readme to include the health environment


gweiying (approved):
lgtm

PR #1901, opened 2022-08-02
WIP: update transition page flow and wording

darrelhong:
This PR contains some improvements I believe would improve the go.gov.sg transition page.

After discussing about gogov during interview, I wanted to make some suggestions to the project as I felt the page could be improved. Since gogov is open-source, I thought I might as well propose the changes with practical implementation. I will explain the reasoning behind each change in the list below but bear in mind that some changes are related to each other.

#### 1. remove “Beware of phishing!’ prompt.
It might be just me but I feel like this prompt is a little doomsday-ee. Imagine a user just trying to visit their gov pages and the first thing they see is ‘Beware of phishing!’, I think it does not serve the purpose well as most people would not even be aware that the page is for anti-phishing. The first time I saw the page, I thought it looked a bit alarming, and was like *"wat do u mean make sure the link starts with go.gov.sg, that’s how i got here"*. To replace it, I have some options one which I list in the following points.

#### 2. Remove the timer and users must click before proceeding to page. 
I’m sure this point has been heavily discussed within the team before deciding on the current implementation. The aim of the timer is to give people enough time to check their URLs before the page automatically redirects. The two problems with this are 1. users who are slow process information would not be able to do anything within the time. 2, even for people who understand the prompts, they would simply wait or press skip because it is the most prominent, in fact the only call to action. In essence, I feel removing the sense of urgency is the most important reason for removing the timer, so people don't feel pressured or anything. This change should optimally be combined with the next point for the best outcome.

#### 3. Better prompts for people who actually care about what the page is about
In addition to removing the timer, i think it’ll be better to link users to a page explaining the what the page is about. In fact, this link already exists on the current page, but I would have never known if not for looking into the source code. I think one of the main aims of the page is to educate the public on the importance of checking their urls but I think no one is going to think about something that will pass by in a few seconds. That’s why I updated the prompt to say “Why are go.gov.sg links safe?” as it better conveys the idea that go.gov.sg links are safe. This combined with removing the timer the allows users to absorb the information at their own time, and those who are really interested can actually find out more, although probably only 1% of users will ever visit the site. With these changes, i18n also has to be considered if not you’ve essentially blocked non-english speaking users from accessing sites. Ultimately, it is down to whether the benefits of removing the timer outweighs the cost of adding one more click. From the perspective of educating the public, I think it is much better to let users click on their own.

#### 4. Better page heading
This point is linked to the above. The current heading is “Check you address bar” which again I feel does not bring across what the page is about. I think its is better to convey the idea that this page is an official page therefore I updated the heading to say so. 

#### 5. Tell users what page they will be redirected too.
I’m not sure if this would be too tedious to implement, but I think the ideal way is to let people know what page or what agency website they will be redirected to. The main thing is to do with user control and mental model. Users enter a link and expect to directly go to the page they intended, but instead they are presented with a interstitial page with no info of where it’ll redirect next, which is just not aligned with the mental model of web browsing. The best way is to show users a short description of the page to be redirected to, but then old urls will not have this field. ~~Or perhaps Url.description might be suitable.~~ (not being used) On top of that, you still have to consider i18n implication.. Another option maybe simply show the url on the page, but I think that will lead to too much information on the page.

---

Overall, I do think admit that removing the timer and show the page name might be a little confusing, as people have already gotten used to the current behavior, but I feel at the very least the headings and links to more information can be updated to convey a more accurate message.

Lastly, I have to end with a disclaimer that I am just giving some suggestions and I definitely do not have the full context of the entire team. Perhaps the team has already considered all these options and ruled them out, or perhaps have conducted focus testing groups to arrive at the current implementation. If so feel free to disregard this PR, but I still feel it is worth sharing.

## Example screenshots

**simple header**:

simple header wide

**short page description**:

page desc header wide

**show url**:
a little too much info i think

page url header wide

### Example FAQ page
faq

### DIFF
--- a/public/assets/edu/styles/colours.css
+++ b/public/assets/edu/styles/colours.css
@@ -11,6 +11,7 @@
     --transition-top-half-color: #2b2e4a;
     --transition-top-half-content-color: #ffffff;
     --transition-top-half-content-link-color: white;
+    --transition-top-half-content-link-hover-color: darkslategray;
     --transition-links-color: #000000;
     --transition-skip-span-color: #48426d;
     --transition-footer-p-color: #bbbbbb;
--- a/public/assets/gov/styles/colours.css
+++ b/public/assets/gov/styles/colours.css
@@ -11,6 +11,7 @@
     --transition-top-half-color: #384a51;
     --transition-top-half-content-color: #ffffff;
     --transition-top-half-content-link-color: white;
+    --transition-top-half-content-link-hover-color: darkslategray;
     --transition-links-color: #000000;
     --transition-skip-span-color: #456682;
     --transition-footer-p-color: #bbbbbb;
--- a/public/assets/health/styles/colours.css
+++ b/public/assets/health/styles/colours.css
@@ -11,6 +11,7 @@
     --transition-top-half-color: #472F40;
     --transition-top-half-content-color: #ffffff;
     --transition-top-half-content-link-color: white;
+    --transition-top-half-content-link-hover-color: darkslategray;
     --transition-links-color: #000000;
     --transition-skip-span-color: #6D4559;
     --transition-footer-p-color: #bbbbbb;
--- a/public/assets/styles/transition-page/transition-page.css
+++ b/public/assets/styles/transition-page/transition-page.css
@@ -27,19 +27,34 @@ body {
   position: absolute;
   bottom: -56px;
   left: 50%;
+  margin-right: -50%;
   transform: translateX(-50%);
 }
 
 .top-half-content h3 {
   font-weight: 600;
-  margin-bottom: 6px;
+}
+
+.top-half-content p {
+  font-size: 16px;
 }
 
 .top-half-content a {
   display: block;
   color: var(--transition-top-half-content-link-color);
-  font-size: 16px;
-  margin-bottom: 35px;
+}
+
+#title {
+  margin-bottom: 16px;
+}
+
+#prompt {
+  margin-bottom: 16px;
+}
+
+#countdown {
+  font-size: 13px;
+  display: inline-block;
 }
 
 .browser-image {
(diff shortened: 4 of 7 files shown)


darrelhong:
update with example faq page

darrelhong:
would love to hear your thoughts @yong-jie @LoneRifle @liangyuanruo

LoneRifle:
cc @leonardloo who proposed the original ideas under discussion, and @shanty-ogp our UX writer

liangyuanruo:
@darrelhong thanks for your suggestions, they are very well thought out.

I think you rightly point out some UX issues on this page that may not allow the app to fully safeguard the user.

A couple of years back we ran experiments to find out how forcing a user interaction or delaying the duration of the transition page would affect user engagement. Results showed that the drop off rate was non negligible; a few percentage of users never made it to the target page for every second of delay introduced.

Given that the primary goal of the app is to bring users to verified resources, and user education being a secondary goal, we decided to keep the page transition low friction and not force user interaction.

It's possible that users may be used to this interaction by now and we are not at the same baseline as two years ago, perhaps we can consider that moving forwards. I especially like the idea of informing users what the target page is going to be, that will make it harder for phishers to replicate, yet is reasonably straightforward to implement.

darrelhong:
agree that the biggest priority is to get users to the destination page, so an additional click will probably be too intrusive and create too much friction especially as users might already used to the current behaviour.

i have updated the PR with only updates to the wording and included some css fixes. Only thing I would like to note is if some URLs are super long, they might look even more suspicious, and with the new wording and display of url its seems like there's too much information to be digested in a few seconds. So definitely needs more deliberation. As you mentioned, probably the most useful change if any, is just to show the destination url.

with url

without url

liangyuanruo:
> with the new wording and display of url its seems like there's too much information to be digested in a few seconds ... probably the most useful change if any, is just to show the destination URL

just an idea, but it _could_ be sufficient to simply show the top-level domain e.g. _drive.google.com_ instead of _https://drive.google.com/xxxxxxxxxx_.

in any case, i think I've set enough context. will let @thanhdatle take this PR forward as tech lead :)

darrelhong:
+1 can consider domain name + ellipsis to denote a longer URL
Screenshot 2022-08-08 at 1 23 28 PM

darrelhong:
.

PR #1904, opened 2022-08-03
feat: add OTP metrics and logging

halfwhole:
## Problem

We'd like to have custom metrics to count the number of OTP successes/failures, and log down both success and failure cases as well

Notion issue [here](https://www.notion.so/opengov/Add-metrics-logging-for-OTP-emails-sent-37210652dacf4d91ae9ef2186361989d)

## Solution

Added utility to collect custom Datadog metrics for Go
- Added statsd from `hot-shots` for custom metrics collection (as per the [Datadog documentation](https://docs.datadoghq.com/integrations/node/))
- Added the prefix `go.` to our custom metrics, to differentiate them from other teams' metrics

Added logging for OTP successes/failures and two custom metrics, `go.otp.success` and `go.otp.failure`

## Tests

- [x] Tested on staging, these OTP metrics show up as expected on Datadog's metrics explorer

## Deploy Notes

**New dependencies**:

- `hot-shots` : client to collect and send custom metrics to Datadog

### DIFF
--- a/src/server/util/dogstatsd.ts
+++ b/src/server/util/dogstatsd.ts
@@ -1,0 +1,7 @@
+import { StatsD } from 'hot-shots'
+
+const dogstatsd = new StatsD({
+  prefix: 'go.',
+})
+
+export default dogstatsd
--- a/src/server/util/dogstatsd.ts
+++ b/src/server/util/dogstatsd.ts
@@ -1,7 +1,9 @@
 import { StatsD } from 'hot-shots'
+import { DEV_ENV } from '../config'
 
 const dogstatsd = new StatsD({
   prefix: 'go.',
+  mock: DEV_ENV,
 })
 
 export default dogstatsd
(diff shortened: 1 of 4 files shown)


src/server/util/dogstatsd.ts
```diff
@@ -0,0 +1,7 @@
+import { StatsD } from 'hot-shots'
+
+const dogstatsd = new StatsD({
+  prefix: 'go.',
+})
+
```
yong-jie, line 6: was reading the documentation for `hot-shots`, and they mention the constructor of `StatsD` takes in a `host` parameter. If unspecified, it defaults to check `DD_AGENT_HOST`.

I see that in our elastic beanstalk environment this appears to be defined via `ebextensions`. However, will this pose any problems in local development? 

When `DD_AGENT_HOST` is `undefined`, it assumes the `host` to be `127.0.0.1` port `8125`. There probably won't be any hosts listening on that address/port.

halfwhole: there should be no issues on local too since UDP is used by default, but yep it doesn't seem great to send out messages when there's no one listening - will try using the `mock` option on statsd to disable it on dev environment

yong-jie (approved):
lgtm!

PR #1910, opened 2022-08-09
(backend) Feat/link audit

gweiying:
## Problem

We want to provide an audit trail for link creators, so that they can know who has changed what on their link, and when.

See [Notion story](https://www.notion.so/opengov/As-a-user-I-want-to-know-what-has-changed-on-my-link-and-by-whom-cf452a2d99864ac8850f01e0f6721296) for details on user journey.

We currently store this data in our `url_histories` table. We parse the url records and expose an endpoint for the frontend to retrieve the records.

## Solution

- Add repository layer to retrieve url records
- Add service layer to parse url records to return pairwise change sets
- Add controller and exposes endpoint `/api/link-audit`

For authenticated query:
```
GET /api/link-audit?url=<shortUrl>&limit=10&offset=10
```

Returns: (JSON) HttpStatus 200 for success
```
{
  changes: [ // ordered from most recent
    {
      "type": "update",
      "key": "state",
      "prevValue": "ACTIVE",
      "currValue": "INACTIVE",
      "updatedAt": "2022-04-06 09:15:26.272+00",
    },
    {
      "type": "update",
      "key": "longUrl",
      "prevValue": "https://google.com",
      "currValue": "https://twitter.com",
      "updatedAt": "2022-03-01 13:59:01.341+00',
    },
    {
      "type": "create",
      "key": "longUrl",
      "prevValue": "",
      "currValue": "https://abc.com",
      "updatedAt": "2022-01-30T04:45:55.204Z"
    }
  ],
  offset: 10, 
  limit: 10, 
  count: 13, // total count of results
}
```

`changes` is an array of type `changeSets[]`

**Values:**
- **type (string):** One of `'create', 'update'`
- **key (string):** One of `'state','userEmail','longUrl'`
- **prevValue (string):** string representation of previous key value
- **currValue (string):** string representation of current key value
- **updatedAt**

Refer to [Notion doc](https://www.notion.so/opengov/20220803-Link-Audit-7cd6cc4040bd4e90b9a331ea5ddbf721) for technical design.

## Tests

## Future work

**Improving pagination**
We use the simplest approach to pagination currently (via limit and offset). There are various ways to improve this:
- primary-key-based offsets e.g. “where id < last_id_seen” in the query
- `ORDER BY id` rather than something like `ORDER BY created_at`

The rationale for these optimizations are documented in the [Notion technical document](https://www.notion.so/opengov/20220803-Link-Audit-7cd6cc4040bd4e90b9a331ea5ddbf721) - we can do these optimisations at a later date if we evaluate that this is a performance issue.

### DIFF
--- a/src/server/modules/audit/LinkAuditController.ts
+++ b/src/server/modules/audit/LinkAuditController.ts
@@ -1,0 +1,18 @@
+import { Request, Response } from 'express'
+
+import { injectable } from 'inversify'
+
+@injectable()
+export class LinkAuditController {
+  public getLinkAudit: (req: Request, res: Response) => Promise<void> = async (
+    req,
+    res,
+  ) => {
+    const { url: shortUrl, offset, limit } = req.query
+    console.log(shortUrl, offset, limit)
+    res.status(200).send()
+    return
+  }
+}
+
+export default LinkAuditController
--- a/src/server/modules/audit/LinkAuditController.ts
+++ b/src/server/modules/audit/LinkAuditController.ts
@@ -1,17 +1,48 @@
 import { Request, Response } from 'express'
-
-import { injectable } from 'inversify'
+import { inject, injectable } from 'inversify'
+import { LinkAuditService } from './interfaces'
+import { DependencyIds } from '../../constants'
+import jsonMessage from '../../util/json'
 
 @injectable()
 export class LinkAuditController {
+  private linkAuditService: LinkAuditService
+
+  public constructor(
+    @inject(DependencyIds.linkAuditService)
+    linkAuditService: LinkAuditService,
+  ) {
+    this.linkAuditService = linkAuditService
+  }
+
   public getLinkAudit: (req: Request, res: Response) => Promise<void> = async (
     req,
     res,
   ) => {
-    const { url: shortUrl, offset, limit } = req.query
-    console.log(shortUrl, offset, limit)
-    res.status(200).send()
-    return
+    const { url, offset, limit } = req.query
+    const shortUrl = url as string
+    if (!shortUrl) {
+      res.status(404).send(jsonMessage('Short url does not exist'))
+      return
+    }
+    const user = req.session?.user
+    if (!user) {
+      res.status(401).send(jsonMessage('User session does not exist'))
+      return
+    }
+    try {
+      const linkStats = await this.linkAuditService.getLinkAudit(
+        user.id,
+        shortUrl,
+        parseInt(limit as string, 10) || undefined,
+        parseInt(offset as string, 10) || undefined,
+      )
+      res.status(200).json(linkStats)
+      return
+    } catch (error) {
+      res.status(404).send(jsonMessage(error.message))
+      return
+    }
   }
 }
 
--- a/src/server/modules/audit/LinkAuditController.ts
+++ b/src/server/modules/audit/LinkAuditController.ts
@@ -40,6 +40,9 @@ export class LinkAuditController {
       res.status(200).json(linkStats)
       return
     } catch (error) {
+      if (error instanceof Error) {
+        res.status(400).send(jsonMessage(error.message))
+      }
       res.status(404).send(jsonMessage(error.message))
       return
     }
--- a/src/server/modules/audit/interfaces/LinkAuditService.ts
+++ b/src/server/modules/audit/interfaces/LinkAuditService.ts
@@ -1,8 +1,20 @@
-interface LinkChangeSet {
-  type: string
+import { UrlHistoryRecord } from './UrlHistoryRepository'
+
+export type LinkChangeKey =
+  | 'description'
+  | 'isFile'
+  | 'state'
+  | 'userEmail'
+  | 'longUrl'
+
+export type LinkChangeType = 'create' | 'update'
+
+export interface LinkChangeSet {
+  type: LinkChangeType
+  key: LinkChangeKey
   prevValue: string
   currValue: string
-  updatedAt: Date
+  updatedAt: string
 }
 
 export interface LinkAudit {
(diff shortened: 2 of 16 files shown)


yong-jie:
just dropping a quick ack here that i've started reviewing this PR, but will need a bit more time after the dust settles to finish 🙏

src/server/modules/audit/interfaces/LinkAuditService.ts
```diff
@@ -0,0 +1,61 @@
+import { UrlHistoryRecord } from './UrlHistoryRepository'
+
+export type LinkChangeKey =
+  | 'description'
+  | 'isFile'
+  | 'state'
+  | 'userEmail'
+  | 'longUrl'
```
yong-jie, line 8: Something we could discuss about:

I was thinking it would be nice if we had a way to guarantee that every value here indeed corresponds to a key of `UrlType`. That way, in the future if someone accidentally renames or removes any key in `UrlType`, typescript compiler can crash and alert devs about the breaking change in this part of the codebase.

I started with this: `export type LinkChangeKey = keyof UrlType`. It gave the list of all possible keys in `UrlType`. We have now created an explicit dependency where anything of type `LinkChangeKey` must be a key of `UrlType`.

We now need to narrow the key space to just the 5 mentioned in your diff. The thing is, typescript syntax only allows for declaring supersets (using extends), but not subsets. Ideally we want to say that `UrlType extends LinkChangeKey { <additional keys here> }`, but this conflates link audit features with the pure data model types.

Instead, we declare our own 'reverse extend' type, like this: `type ReverseExtends = <U, T extends U> = T`. Whatever we feed into `U` will now be reverse-extended (i.e. a subset type)

As an example, `export type LinkChangeKey = ReverseExtends<keyof UrlType, 'description' | 'isFile'>`. Now, any variable with the `LinkChangeKey` type must be either 'description' or 'isFile', and we also have a guarantee that these strings are valid keys of `UrlType`. If you delete the `description` field of `UrlType`, tsc now crashes, saying `"TS2344: Type '"isFile" | "description"' does not satisfy the constraint 'keyof UrlType'.   Type '"description"' is not assignable to type 'keyof UrlType'."`

We're close, I just can't think of a good name that can convey this utility type for future engineers looking at your PR. Was thinking along the lines of `Certain` or `Specific`, so it reads naturally in english: `type LinkChangeKey = Specific<keyof UrlType, ...>`, reading literally as 'specific/certain keys of UrlType'.

Would like to seek your input on this matter as well, and any complications you foresee if we explicitly declare a dependency between `LinkChangeKey` and `UrlType`

yong-jie (commented):
Nice work! PR looks functionally complete; i just have some minor nits you could consider addressing, and some things i'd to hear your thoughts on.

src/server/modules/audit/interfaces/LinkAuditService.ts
gweiying: I think this is going in a great direction. I see no specific complications with declaring a dependency between `LinkChangeKey` and `UrlType`. I was also thinking about the same problem but I didn't tackle it as thoroughly as you until you wrote the above review.

Similar to you, initially I was stuck on how to select only certain key strings from `keyof UrlType`. However, thinking about it differently, we could use `Pick` to select the properties that are mutable from `UrlType` (`'description' | 'isFile' | 'state' | 'longUrl'`), before applying `keyof`.

Restating it more clearly, this would take the form 
```
type LinkChangeKey = keyof (Pick<UrlBaseType, 'description' | 'isFile' | 'state' | 'longUrl'>)
```

Using the native implementation has the benefit of being immediately recognisable for future developers, but I can also see benefits of explicitly defining a `ReverseExtends` utility type. 

What do you think?

src/server/modules/audit/LinkAuditController.ts
```diff
@@ -0,0 +1,49 @@
...
+      const linkStats = await this.linkAuditService.getLinkAudit(
+        user.id,
+        shortUrl,
+        parseInt(limit as string, 10) || undefined,
+        parseInt(offset as string, 10) || undefined,
+      )
+      res.status(200).json(linkStats)
+      return
+    } catch (error) {
+      res.status(404).send(jsonMessage(error.message))
```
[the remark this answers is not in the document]
gweiying: Good point, fixed here [e11c08d](https://github.com/opengovsg/GoGovSG/pull/1910/commits/e11c08daf7edc689b9f4569584af9a5335670692)

src/server/modules/audit/services/LinkAuditService.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+    return changeSets
+  }
+
+  getLinkAudit: (
+    userId: number,
+    shortUrl: string,
+    limit?: number,
+    offset?: number,
+  ) => Promise<interfaces.LinkAudit | null> = async (
```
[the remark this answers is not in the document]
gweiying: Good catch! Was a relic from past edit history, fixed [7faef26](https://github.com/opengovsg/GoGovSG/pull/1910/commits/7faef26affbb640b5e8fa63445cadb4dd7e970cc)

src/server/modules/audit/interfaces/LinkAuditService.ts
yong-jie: ooooh i like your idea, it doesn't rely on any strange magic and we don't need to introduce any additional utility types to achieve what we want. i vote to go with this approach 👍

gweiying: Updated in [c32e269](https://github.com/opengovsg/GoGovSG/pull/1910/commits/c32e26997d379dfafd67d6f3a410e67d35632858)

yong-jie: lgtm! seeing the `userEmail` key being excluded feels bad, maybe LinkChangeKey can be derived from `UrlHistoryRecord` instead of `UrlBaseType` in the future 🤔  but looks good for now!

yong-jie (approved):
lgtm! feel free to merge when ready~

PR #1911, opened 2022-08-10
chore: log OTP verification successes and failures

halfwhole:
## Problem

Apart from logging down successes/failures for OTP generation, we'd like to do so for OTP verification as well. Currently for OTP verification we have some console outputs - these can be seen on CloudWatch, but moving them to logger will make them also visible on Datadog

## Solution

Change `console.info` to `logger.info` and `logger.error`

## Tests

- [x] Tested on staging to ensure that OTP verification logs are visible on both CloudWatch and Datadog


gweiying (approved):
lgtm

PR #1912, opened 2022-08-10
Feat/postman

gweiying:
## Solution

Implemented Postman fallback mailer.

### Notes
Postman fallback mailer will only be activated if env variable `ACTIVATE_POSTMAN_FALLBACK` is set to `true`. If not, it should still use the default SES email transport. 

## Tests

- [x] add env variables to Go-staging
- [x] push to edge
- [x] verify that non-SGMail inboxes can receive OTP from go-staging (via Postman)
- [x] verify that SGMail inboxes can receive OTP from go-staging (via Postman)
- [x] verify that non-SGMail inboxes can receive OTP from edu-staging or health-staging (via SES)
- [ ] verify whether SGMail inboxes can receive OTP from edu-staging or health-staging (via SES)

## Other todos
- [ ] **Determine if we should apply Postman fallback to health and edu environments as well**

## Deploy Notes

- [ ] add env variables to prod 
  - [ ] `POSTMAN_API_KEY` - 1Password
  - [ ] `POSTMAN_API_URL` - `https://api.postman.gov.sg/v1/transactional/email/send`
  - [ ] `ACTIVATE_POSTMAN_FALLBACK` - `true`
- [ ] deploy to prod

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@sentry/webpack-plugin": "^1.15.1",
     "@types/express-rate-limit": "^5.1.3",
     "aws-sdk": "^2.1101.0",
+    "axios": "^0.27.2",
     "babel-polyfill": "^6.26.0",
     "bcrypt": "^5.0.1",
     "body-parser": "^1.18.3",
--- a/src/server/services/email.ts
+++ b/src/server/services/email.ts
@@ -1,10 +1,19 @@
 /* eslint class-methods-use-this: ["error", { "exceptMethods":
-  ["mailOTP", "initMailer"] }] */
+  ["mailOTP", "initMailer", "sendPostmanMail", "sendTransporterMail"] }] */
 
 import { injectable } from 'inversify'
 import nodemailer from 'nodemailer'
+import axios from 'axios'
 import assetVariant from '../../shared/util/asset-variant'
-import { logger, ogUrl, otpExpiry, transporterOptions } from '../config'
+import {
+  activatePostmanFallback,
+  logger,
+  ogUrl,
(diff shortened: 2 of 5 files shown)


src/server/services/email.ts
```diff
@@ -60,4 +94,36 @@ export class MailerNode implements Mailer {
...
+    return this.sendTransporterMail(mailBody)
+  }
+
+  mailOTP(email: string, otp: string, ip: string): Promise<void> {
+    if (!email || !otp) {
+      logger.error('Email or OTP not specified')
+      return Promise.resolve()
+    }
+
+    const emailHTML = `Your OTP is <b>${otp}</b>. It wii'll expire in ${Math.floor(
```
halfwhole, line 113:
```suggestion
    const emailHTML = `Your OTP is <b>${otp}</b>. It will expire in ${Math.floor(
```

package.json
```diff
@@ -39,6 +39,7 @@
     "@sentry/webpack-plugin": "^1.15.1",
     "@types/express-rate-limit": "^5.1.3",
     "aws-sdk": "^2.1101.0",
+    "axios": "^0.27.2",
```
yong-jie, line 42: tech debt to address next time: the codebase currently using `cross-fetch` for making HTTP requests. axios might not be necessary

yong-jie (approved):

package.json
gweiying: Aha! Thanks for highlighting, I was poking around and couldn't find a HTTP request module.

yong-jie: no worries we can address this next time!

PR #1913, opened 2022-08-10
Feat/sns consumer lambda

yong-jie:
## Problem

We need a way to capture SES email sending events, which are currently published to an SNS topic.

## Solution

Create a lambda that acts as a subscriber to said SNS topic. The actual subscription needs to be done manually, and we can probably just use one single lambda (gov-production maybe?) as the subscriber. In the future, we can split the SES events for gov, edu and health into separate SNS topics and have their respect lambdas subscribe to them.

## Deploy Notes

Grab the ARN of the newly created lambda and add it as a subscriber to the SNS topic.

## Alternatives considered

We can also introduce an API endpoint in our application, but that introduces unnecessary coupling of the app to this auxiliary feature.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -37,4 +37,16 @@ functions:
           - 'src/server/serverless/lambda-migrate-user-links/**'
           - 'package-lock.json'
           - 'package.json'
+  capture-ses-events:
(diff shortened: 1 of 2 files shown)


gweiying (approved):
lgtm

PR #1917, opened 2022-08-15
feat: added tags table and many to many rel to urls

thanhdatle:
**Features**:

- Details ...
- To include tags in link creation API


src/server/mappers/TagMapper.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable class-methods-use-this, lines-between-class-members, no-dupe-class-members */
+import { injectable } from 'inversify'
+import { StorableTag } from '../repositories/types'
+import { TagType } from '../models/tag'
+import { Mapper } from './Mapper'
+
+@injectable()
+export class TagMapper implements Mapper<StorableTag, TagType> {
+  persistenceToDto(tagType: TagType): StorableTag
+  persistenceToDto(tagType: TagType | null): StorableTag | null {
```
gweiying, line 10: Just wondering why we need this class `TagMapper` if we don't permute the data. 

Here, we're only checking for existence and returning the object as is if it exists, otherwise returning null. We could skip this mapper class and return the object directly since we don't permute it.

src/server/mappers/TagMapper.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (!tagType) {
+      return null
+    }
+    return {
+      id: tagType.id,
+      tagKey: tagType.tagKey,
+      tagString: tagType.tagString,
+    }
+  }
+}
```
gweiying, line 20: Wondering what the purpose of the `Mapper` class is since we're not permuting the object.

Here, we're only checking for existence and returning the object as is if it exists, otherwise we return null - the same can be achieved without this `Mapper` layer since the object retrieved from the DB would have the same attributes/ return null too otherwise.

src/server/mappers/UrlMapper.ts
```diff
@@ -14,11 +14,21 @@ export class UrlMapper implements Mapper<StorableUrl, UrlType> {
...
     if (!urlClicks || !Number.isInteger(urlClicks.clicks))
       throw new Error('UrlClicks object not populated.')
+
+    const tagStrings: string[] = []
+    const { tags } = urlType
+    if (tags) {
+      for (let i = 0; i < tags.length; i += 1) {
+        tagStrings.push(tags[i].tagString)
+      }
+    }
```
gweiying, line 24: Small TS-related suggestion, might want to simplify it with the `.map()` function.

src/server/modules/user/index.ts
```diff
@@ -20,6 +20,10 @@ type LinkInformationProperties = {
 
 type ShortUrlOperationProperty = UserIdProperty & ShortUrlProperty
 
+type OptionalTagsProperty = {
+  tags?: string[] | undefined
```
gweiying, line 24: small nit, redundant `undefined` here.

`tags?: string[] | undefined` is same as `tags?: string[]` because of the optional property `?`


 https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#exact-optional-property-types---exactoptionalpropertytypes

src/server/models/tag.ts
```diff
@@ -0,0 +1,65 @@
...
+      /**
+       * Use the replica database for read queries. To be enabled
+       * when realtime data is not needed.
+       */
+      useReplica: {
+        useMaster: undefined,
+      },
+    },
+    hooks: {
+      afterCreate: async (_: TagType, options) => {
```
gweiying, line 49: Can we add an inline comment here to clarify that it is necessary to check for a wrapper transaction because there's a downstream operation (assigning the tag to the url)?

gweiying (changes requested):
Looked through the PR briefly, overall it seems great, the structure seems good and most things made sense to me. Just left a couple of small inline nits. 

One larger missing point I wanted to highlight is that tags are not written into the UrlHistory table upon creation in this PR. For audit sake, it would probably be important to store the `tag -> url` mapping in the `url_history` table. Would you think it would make sense to add this functionality within this PR?

(edit, 1 Sep 11.35am) Chatted with Dat about this and we decided that the cleanest way to do this is to serialise the tags into the `url` table as well, and have the `afterCreate`, `afterUpdate` hooks handle the storing of changes to the `url_history` table.

gweiying:
Other comments for easier reviewing in the future:

Would it be possible to add more detailed git commit messages so that I can step through the commits slowly and see what changes are added with each commit? Currently, all the git commit messages are the same - Having separate git commit messages might make reviewing large PRs easier so that engineers can step through the PR commit-by-commit, since it was quite a large number of files that changed at once.

halfwhole (commented):
for discussion: was thinking that we should add tests for validating the `urlSchema` somewhere, but there doesn't seem to be anywhere in the codebase where these validators are tested. where could we add them, if we want to do so at all?

also, is it worthwhile to update the type signature for `create` in `test/server/mocks/repositories/UrlRepository.ts` to match the new version, and make the mock for `TagRepository.ts` too? (not sure of this though, because we don't seem to be using them much)

src/server/models/tag.ts
```diff
@@ -0,0 +1,65 @@
...
+      unique: true,
+      validate: {
+        is: TAG_STRING_REGEX,
+      },
+    },
+    tagKey: {
+      type: Sequelize.STRING(25),
+      allowNull: false,
+      validate: {
+        is: TAG_KEY_REGEX,
```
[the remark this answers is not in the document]
gweiying: +1 good point!

src/server/mappers/TagMapper.ts
thanhdatle: Generally, it's a good idea to separate Model object and client facing object since they can be quite different.

src/server/mappers/UrlMapper.ts
thanhdatle: I fixed this

src/server/modules/user/index.ts
thanhdatle: nice catch, will fix this

src/server/models/tag.ts
thanhdatle: A tag is only created to be associated to a link. If somehow the association operation failed, it makes sense to remove the tag.

src/server/models/tag.ts
```diff
@@ -0,0 +1,65 @@
...
+    tagString: {
+      type: Sequelize.STRING(25),
+      allowNull: false,
+      unique: true,
+      validate: {
+        is: TAG_STRING_REGEX,
+      },
+    },
+    tagKey: {
+      type: Sequelize.STRING(25),
```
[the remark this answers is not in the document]
thanhdatle: already fixed this

src/server/models/url.ts
```diff
@@ -25,6 +26,7 @@ export interface UrlBaseType extends IdType {
 
 export interface UrlType extends IdType, UrlBaseType, Sequelize.Model {
   readonly UrlClicks?: UrlClicksType
+  tags?: TagType[]
```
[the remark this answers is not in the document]
thanhdatle: already fixed this

src/server/models/tag.ts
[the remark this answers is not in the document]
thanhdatle: generally we should do both but I changed the tag length constraint in the DB to 255

src/server/api/user/validators.ts
```diff
@@ -34,6 +35,17 @@ export const urlSchema = Joi.object({
...
+    .max(3)
+    .optional()
+    .items(
+      Joi.string().custom((tag: string, helpers) => {
+        if (!isValidTag(tag)) {
+          return helpers.message({ custom: `tag: ${tag} format is invalid` })
+        }
+        return tag
+      }),
+    ),
```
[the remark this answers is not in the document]
thanhdatle: good idea, done

test/server/repositories/UrlRepository.test.ts
```diff
@@ -74,15 +79,27 @@ describe('UrlRepository', () => {
...
+    addTag: jest.fn(),
+  }
+
   const baseStorableUrl = {
     ...baseTemplate,
     ...baseUrlClicks,
   }
+  const baseStorableUrlWithTags = {
+    ...baseStorableUrl,
+    tags: ['Tag1', 'Tag2'],
```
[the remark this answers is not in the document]
thanhdatle: already fixed!

src/server/repositories/UrlRepository.ts
```diff
@@ -54,28 +55,53 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+          ? this.fileBucket.buildFileLongUrl(file.key)
+          : properties.longUrl,
+        isFile: !!file,
+      }
+      const url = await Url.create(urlStaticDTO, {
+        transaction: t,
+      })
+      if (properties.tags) {
+        const tagCreationResponses = await Promise.all(
+          properties.tags.map(async (tag) => {
```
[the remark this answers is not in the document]
thanhdatle: good catch, done!

src/server/models/index.ts
```diff
@@ -6,11 +6,18 @@ import { WeekdayClicks } from './statistics/weekday'
 import { Devices } from './statistics/devices'
 import { UrlClicks } from './statistics/clicks'
 import { syncFunctions } from './functions'
+import { Tag } from './tag'
 
 // One user can create many urls but each url can only be mapped to one user.
 User.hasMany(Url, { as: 'Urls', foreignKey: { allowNull: false } })
 Url.belongsTo(User, { foreignKey: { allowNull: false } })
 
+export const UrlTag = sequelize.define('url_tag', {}, { timestamps: true })
```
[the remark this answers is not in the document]
thanhdatle: I find it's better to be explicit.

Issue #1928, opened 2022-08-18
Tech debt: Improve Coveralls coverage


gweiying:
Key culprit:
- Email sending
- New lambdas (?)

PR #1932, opened 2022-08-22
Fix/mail

### LINKED ISSUE #1915: Tech Debt: Remove usage of axios in postman feature
The codebase is using `cross-fetch`. The introduction of axios was a quick way to get http requests sent in a more familiar interface to the engineer. #1912 

We could change the axios call into a cross-fetch one some time after everything stables.

Back then there were problems with HTTP request libraries regarding browser-compat, making it necessary for us to use `cross-fetch`. If this problem is no longer still there, we can also consider taking a greater effort to migrate over from `cross-fetch` to `axios`.

**Additional context**
The postman feature was a quick fix that introduces email-OTP fallback.

gweiying:
## Problem

I broke our Coveralls tests unknowingly (fell below 75%) when I made the changes to include a Postman fallback option (#1912) without adding tests. 

Closes #1915 and #1928

## Solution

- Added test routes for email module
- Replaced `axios` with `cross-fetch`
- Fixed `try-catch` nesting for OTP generation and sending

## Other notes
~I would love to propose that we run Coveralls against PRs to `develop`. Currently, Coveralls runs on PRs to `release`, but that since merges to `releases` are usually after we complete feature development and are prepared to cut a stable version, the coverage tests comes a little belated.~ Seems like Coveralls runs against `develop` too, my bad.

## Tests

- See [Coveralls coverage now](https://coveralls.io/jobs/104841285/source_files/24604087889#L115)
- Jest coverage now Screenshot 2022-08-22 at 12 45 14

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5117,32 +5117,6 @@
       "integrity": "sha512-5LMaDRWm8ZFPAEdzTYmgjjEdj1YnQcpfrVajO/sn/LhbpGp0Y0H64c2hLZI1gRMxfA+w1S71Uc/nHaOXgcCvGg==",
       "dev": true
     },
-    "axios": {
-      "version": "0.27.2",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-0.27.2.tgz",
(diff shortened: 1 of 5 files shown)


yong-jie (approved):
lgtm! yes coveralls has always been there for `develop` branch as well, just that it normally passes and hence gets drowned in the 10-19 other green ticks in the PR checks 😂

PR #1933, opened 2022-08-22
Feat/update link with tags

thanhdatle:
To implement Update link API with tags

### DIFF
--- a/src/server/api/user/index.ts
+++ b/src/server/api/user/index.ts
@@ -6,6 +6,7 @@ import { container } from '../../util/inversify'
 import { MAX_FILE_UPLOAD_SIZE } from '../../../shared/constants'
 import {
   ownershipTransferSchema,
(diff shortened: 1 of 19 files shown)


src/server/api/user/index.ts
```diff
@@ -103,6 +104,12 @@ router.get(
   userController.getUrlsWithConditions,
 )
```
gweiying, line 106: nit, some inline documentation about what endpoint does would be helpful

src/server/modules/user/UserController.ts
```diff
@@ -208,6 +211,7 @@ export class UserController {
       res.ok({ urls, count })
       return
     } catch (error) {
+      console.log(`usercontroller.mesasge ${error.message}`)
```
gweiying, line 214: Typo here. Would we want to use the `logger.info` or the `console.log`?

PR #1937, opened 2022-08-24
feat(datadog): add metrics and set RUM user

halfwhole:
## Problem

We want to collect more useful product metrics for Go

## Solution

**Features**:

- Set user ID and email for datadog RUM on login, remove user on logout. The user ID/email helps us to track how many unique users are using our app over some time period
- `go.otp.generate.(success/failure)`: number of OTP generation successes/failures. This used to be `go.otp.(success/failure)`, but made it more specific to generation
- `go.otp.verify.(success/failure)`: number of OTP verification successes/failures
- `go.shortlink.clicks`: number of user clicks to existing short links
- `go.shortlink.create`: number of created short links, tagged with `isFile`
- `go.user.new`: number of new users

**Improvements**:

- Made some logger OTP messages more descriptive with user email

## Tests

- [x] Test on staging to ensure that new metrics are collected (except `go.user.new`, because I don't have any new account)
- [x] Test on staging to ensure that user details are collected on RUM

**Problem**:

Note (just FYI): we use `go.shortlink.(clicks/create)` instead of `go.shorturl.(clicks/create)` because the latter doesn't work. I suspect datadog is blocking `go.shorturl.(clicks/create)` from being used by dogstatsd, because I had previously tried [generating them from spans](https://docs.datadoghq.com/tracing/trace_pipeline/generate_metrics/), causing a naming conflict.

### DIFF
--- a/src/server/modules/user/UserController.ts
+++ b/src/server/modules/user/UserController.ts
@@ -3,7 +3,6 @@ import { inject, injectable } from 'inversify'
 
 import Sequelize from 'sequelize'
 import { OwnershipTransferRequest, UrlCreationRequest, UrlEditRequest } from '.'
-import dogstatsd from '../../util/dogstatsd'
 import jsonMessage from '../../util/json'
 import { DependencyIds } from '../../constants'
 import {
@@ -67,7 +66,6 @@ export class UserController {
         longUrl,
         file,
       )
-      dogstatsd.increment('shortlink.create', 1, 1, [`isfile:${!!file}`])
       res.ok(result)
       return
     } catch (error) {
--- a/src/server/modules/user/UserController.ts
+++ b/src/server/modules/user/UserController.ts
@@ -3,6 +3,7 @@ import { inject, injectable } from 'inversify'
 
 import Sequelize from 'sequelize'
 import { OwnershipTransferRequest, UrlCreationRequest, UrlEditRequest } from '.'
+import dogstatsd from '../../util/dogstatsd'
 import jsonMessage from '../../util/json'
 import { DependencyIds } from '../../constants'
 import {
@@ -66,6 +67,7 @@ export class UserController {
         longUrl,
         file,
       )
+      dogstatsd.increment('shortlink.create', 1, 1, [`isfile:${!!file}`])
       res.ok(result)
       return
     } catch (error) {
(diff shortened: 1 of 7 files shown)


src/server/repositories/UserRepository.ts
```diff
@@ -49,7 +50,12 @@ export class UserRepository implements UserRepositoryInterface {
   public findOrCreateWithEmail: (email: string) => Promise<StorableUser> = (
     email,
   ) => {
-    return User.findOrCreate({ where: { email } }).then(([user, _]) => user)
+    return User.findOrCreate({ where: { email } }).then(([user, created]) => {
+      if (created) {
+        dogstatsd.increment('user.new', 1, 1)
```
thanhdatle, line 55: is it correct to increase the count here? 
Since the function is `findOrCreateWithEmail`, we might potentially run into existing user

thanhdatle (changes requested):

src/server/repositories/UserRepository.ts
halfwhole: yep it should be correct: the `created` in [`findOrCreate`](https://sequelize.org/docs/v6/core-concepts/model-querying-finders/#findorcreate) is a boolean telling us if a new user's been created, so the metric will increment only if the user is new

trying it out on staging, I also haven't observed any metric increments when logging in with my own existing email, so it matches what we would expect!

gweiying (approved):
looks great!

src/server/modules/user/UserController.ts
```diff
@@ -66,6 +67,7 @@ export class UserController {
         longUrl,
         file,
       )
+      dogstatsd.increment('shortlink.create', 1, 1, [`isfile:${!!file}`])
```
thanhdatle, line 70: Should we move this to the service layer? If this line is called somewhere else, caller might have to duplicate the datadog stat line as well

thanhdatle (changes requested):

src/server/modules/user/UserController.ts
halfwhole: makes sense, i'll move it down to url management service!

thanhdatle (approved):

halfwhole:
a note from @thanhdatle that the metric names should be extracted into another file - will raise this in another PR!

PR #1900, opened 2022-08-02
Chore/serverless upgrade

gweiying:
## Problem

Version 3 of the Serverless framework is out. We fix some breaking changes and upgrade our Serverless framework to v3 in this PR.

Closes #1883

## Solution

Changes:

`stage` 
Previously we injected  `SERVERLESS_STAGE` (takes values `staging` or `production`) into `serverless.yml` as an env variable to reuse the `serverless.yml` template for launching both our staging and production lambdas. However, [Serverless v3 requires `stage` to be resolved before](https://github.com/serverless/serverless/issues/9813) as it is used to load and resolve `.env` files e.g. `.env.development`, `.env.production`. As advised by the team, we switch to setting this via the CLI flag `--stage`.

`useDotEnv: true`
Explicitly enabled to read environment variables, as loading .env files is [kept opt-in](https://www.serverless.com/framework/docs/guides/upgrading-v3#deprecated-features-that-will-be-kept-in-v3)

`configValidationMode: error` 
Explicitly throw errors so that any misconfiguration on the file will throw an error before deployment. Severless has future plans to make [this the default mode](https://www.serverless.com/framework/docs/guides/upgrading-v3#deprecated-features-that-will-be-kept-in-v3), so swapping over early can't hurt

`--verbose`
CLI flag `-v` option is no longer recognized](https://www.serverless.com/framework/docs/guides/upgrading-v3#cli-commands-and-options), swapped to `--verbose`
 

## Tests

- [x] Test Lambda functions on staging to see if new Lambda functions still work

## Deploy Notes

Post deployment, to test lambda functions in production to see if new Lambda functions are working. Also ensure that configurations are not removed (VPC and env variable settings).

PR #1940, opened 2022-08-25
chore(lambda): remove excludeDevDependencies for capture-ses-events

halfwhole:
## Problem

GitHub action for the `edge` branch is failing at the `serverless deploy` step ([example](https://github.com/opengovsg/GoGovSG/runs/8007465383?check_suite_focus=true)). 

Likely because the serverless functions were upgraded to v3 in #1900 which involved removing `excludeDevDepdencies`, but the `capture-ses-events` lambda function from #1913 still had `excludeDevDependencies` that was incompatible with v3

## Solution

Removed `excludeDevDependencies: true` from `capture-ses-events`, action now passes

## Tests

- [x] Tested on health staging that's using SES, can see the logs for sending OTPs on CloudWatch

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -41,7 +41,6 @@ functions:
     memorySize: 128
     description: SNS subscriber for SES events
     package:
-      excludeDevDependencies: true
       patterns:
           - '!./**'
           - 'src/server/serverless/capture-ses-events/**'


yong-jie (approved):
thanks for the catch! lgtm

PR #1942, opened 2022-08-29
chore: send request logs to datadog

halfwhole:
## Problem

HTTP request logs are not being forwarded to Datadog 

## Solution

Forward logs to Datadog by configuring Morgan, our middleware for logging HTTP requests, to send these logs via `logger.info` rather than the default `process.stdout` (see [Morgan documentation](https://github.com/expressjs/morgan#stream), [Stack Overflow question](https://stackoverflow.com/a/28824464)). 

`logger.info` will then output the logs to both the console and Datadog. The `message` is also trimmed to remove an additional newline inserted by morgan.

**Improvements**:
- Removed date/time from morgan log format, because `logger` already takes down the time 

## Before & After Screenshots

After (on CloudWatch):
Screenshot 2022-08-29 at 1 36 53 PM

After (on Datadog):
Screenshot 2022-08-29 at 1 37 51 PM

I blacked out my IP address

## Tests

- [x] Test on staging to ensure request logs are forwarded to Datadog
- [x] Test on staging to ensure request logs are still available on CloudWatch

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -70,8 +70,14 @@ morgan.token('userId', (req: express.Request) =>
 )
 
 const MORGAN_LOG_FORMAT =
-  ':client-ip - [:date[clf]] ":method :url HTTP/:http-version" :status ' +
+  ':client-ip - ":method :url HTTP/:http-version" :status ' +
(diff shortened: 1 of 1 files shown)


gweiying (approved):
lgtm

PR #1919, opened 2022-08-16
feat(tagging): add frontend tagging on create new link form

halfwhole:
## Problem

Users should be able to add tags on the create link form

Notion issue [here](https://www.notion.so/opengov/As-a-user-I-want-to-add-tags-while-creating-link-s-so-that-I-can-identify-the-links-easily-7d9b280c038043b9b2210b8eba045fff)

## Solution

Added

These changes are purely on the frontend.

Pending questions:
- Restrictions on 

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

_What tests should be run to confirm functionality?_

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, etc._

**New environment variables**:

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

- `dependency` : dependency details

**New dev dependencies**:

- `dependency` : dependency details

PR #1950, opened 2022-09-06
Feat/link audit to include tags

thanhdatle:
## Deploy Notes

**New scripts**:

20220905_add_stringTags_column_to_url_and_url_history_tables.sql

-- This adds columns to both urls and url_histories
-- tables, which are backwards-compatible with the current
-- codebase as they default to empty strings.


src/server/modules/user/services/UrlManagementService.ts
```diff
@@ -128,18 +134,22 @@ export class UrlManagementService implements interfaces.UrlManagementService {
...
     conditions: UserUrlsQueryConditions,
   ) => Promise<UrlsPaginated> = (conditions) => {
     return this.userRepository.findUrlsForUser(conditions)
   }
+
+  getTagsWithConditions: (
+    conditions: UserTagsQueryConditions,
+  ) => Promise<string[]> = (conditions) => {
+    return this.tagRepository.findTagsWithConditions(conditions)
+  }
```
gweiying, line 152: Feels unintuitive that `getTagsWithConditions` is in the `UrlManagementService` when it doesn't actually interact with the `urlRepository` or anything. Also, the only reason we imported `tagRepository` into the service is for this function. What do you think of creating a separate `TagManagementService` for this function? Then future extensions to tagging can also belong to that service?

src/server/api/user/validators.ts
```diff
@@ -5,14 +5,62 @@ import {
...
+        .optional()
+        .min(1)
+        .custom((tag: string, helpers) => {
+          if (!isValidTag(tag)) {
+            return helpers.message({ custom: `tag: ${tag} format is invalid` })
+          }
+          return tag
+        }),
+    ),
+})
```
gweiying, line 46: Thinking that we can create a object schema for tags and reuse it, see https://stackoverflow.com/a/42575904

src/server/mappers/UrlMapper.ts
```diff
@@ -14,16 +14,27 @@ export class UrlMapper implements Mapper<StorableUrl, UrlType> {
     const { UrlClicks: urlClicks } = urlType
     if (!urlClicks || !Number.isInteger(urlClicks.clicks))
       throw new Error('UrlClicks object not populated.')
+
+    const tagStrings: string[] = []
+    const { tags } = urlType
+    if (tags) {
+      for (let i = 0; i < tags.length; i += 1) {
+        tagStrings.push(tags[i].tagString)
```
gweiying, line 22: should we use the `TagMapper.persistenceToDto` function here?

src/server/modules/user/index.ts
```diff
@@ -49,7 +55,10 @@ export type GoUploadedFile = {
 }
 
 export type UpdateUrlOptions = Partial<
-  Pick<StorableUrl, 'state' | 'longUrl' | 'description' | 'contactEmail'> & {
+  Pick<
+    StorableUrl,
+    'state' | 'longUrl' | 'description' | 'contactEmail' | `tags`
```
gweiying, line 60: tiny nit, can we standardise on using the single quotation mark here?

src/server/modules/user/UserController.ts
```diff
@@ -174,6 +181,29 @@ export class UserController {
...
+      if (error instanceof NotFoundError) {
+        res.notFound(error.message)
+        return
+      }
+      res.serverError(jsonMessage('Error retrieving URLs for user'))
+      return
+    }
+  }
+
+  private static extractUrlQueryConditions(req: Express.Request) {
```
gweiying, line 206: nice touch with pulling this out into its own function!

src/server/modules/user/UserController.ts
```diff
@@ -184,7 +214,9 @@ export class UserController {
       sortDirection = 'desc',
       isFile,
       state,
+      tags = '',
     } = req.query
+    const tagList = tags ? tags.toString().toLowerCase().split(';') : []
```
gweiying, line 219: should we add `trim()` to preempt any trailing whitespace?

src/server/modules/user/interfaces/UrlManagementService.ts
```diff
@@ -25,6 +27,9 @@ export interface UrlManagementService {
   getUrlsWithConditions: (
     conditions: UserUrlsQueryConditions,
   ) => Promise<UrlsPaginated>
+  getTagsWithConditions: (
```
gweiying, line 30: ditto as above comment, should we move this out into a separate `TagManagementService`

src/server/repositories/UserRepository.ts
```diff
@@ -90,39 +93,82 @@ export class UserRepository implements UserRepositoryInterface {
     conditions: UserUrlsQueryConditions,
```
gweiying, line 93: This file looks much neater! Thanks for refactoring the query conditions for retrieving a url, this is much more maintainable.

src/server/repositories/UserRepository.ts
```diff
@@ -90,39 +93,82 @@ export class UserRepository implements UserRepositoryInterface {
...
-    // to handle edge case where count === 1 but user does not have any urls
+    const urls = rows.map((urlType) => this.urlMapper.persistenceToDto(urlType))
     if (urls.length === 0) {
       count = 0
     }
-
     return { urls, count }
   }
+
+  private static buildQueryConditions(conditions: UserUrlsQueryConditions) {
```
gweiying, line 142: Just one small suggestion, can we add an expected output type here so that it's easy to see at one glance what the built query condition object looks like, since there's quite a lot of optional conditions that can go in

src/server/repositories/UrlRepository.ts
```diff
@@ -54,28 +58,57 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+          properties.tags.map(async (tag) => {
+            return Tag.findOrCreate({
+              where: {
+                tagString: tag,
+                tagKey: tag.toLowerCase(),
+              },
+              transaction: t,
+            })
+          }),
+        )
```
gweiying, line 95: Would it be better to abstract this function into the `TagRepository`? Logically speaking, it might be more consistent and intuitive if the creation of tags happens in the `TagRepository` instead.

It could be a really thin method `tagRepository.findOrCreateTag(tag: string, t: transaction)` that just abstracts over the `Tag.findOrCreate` method, like the one we use in `UserRepository`:

```
public findOrCreateWithEmail: (email: string) => Promise<StorableUser> = (
    email,
  ) => {
    return User.findOrCreate({ where: { email } }).then(([user, _]) => user)
  }
```

src/server/repositories/UrlRepository.ts
```diff
@@ -93,47 +126,108 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+      })
     })
-
     this.invalidateCache(shortUrl)
-
+    if (!newUrl) throw new Error('Newly-updated url is null')
     return this.urlMapper.persistenceToDto(newUrl)
   }
 
+  private async upsertTags(
```
gweiying, line 190: See above comment about `findOrCreateTags`. I think `upsertTags` can also be moved into the `TagRepository` layer for organisation

src/server/util/array.ts
```diff
@@ -0,0 +1,9 @@
+const arraysContainSame = (a: any[] | undefined, b: any[] | undefined) => {
```
gweiying, line 1: Can use the lodash library [_.isEqual](https://lodash.com/docs/#isEqual) instead of building a custom comparator

src/server/repositories/UrlRepository.ts
```diff
@@ -93,47 +126,108 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+    changes: Partial<
+      Pick<
```
gweiying, line 192: Should we move this `changes` type definition into the types file?

src/server/repositories/UrlRepository.ts
```diff
@@ -54,28 +58,57 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+              transaction: t,
+            })
+          }),
+        )
+        await Promise.all(
+          tagCreationResponses.map(async (response) => {
+            const [tag, _] = response
+            if (tag) {
+              // @ts-ignore, addTag is provided by Sequelize during run time.
+              await url.addTag(tag, { transaction: t })
```
gweiying, line 101: I'm not very familiar with this Sequelize special methods/mixins concept, but I'm wondering if we can just call `setTags` instead of calling `addTag` one by one?

gweiying: Suggestion: can we also add the [url to the documentation](https://sequelize.org/docs/v6/core-concepts/assocs/#special-methodsmixins-added-to-instances) for the special methods inline where we use `addTag` and `setTags` so that a future developer looking at this can look directly at the documentation as well?

gweiying (changes requested):
Thanks for working on this feature! 

Once I sat down properly to review, it was actually quite a breeze to go through everything at once, which probably means that everything was well-organised and very readable.

I've just left some comments in line for review, mostly to do with file organisation, repo consistency and other Javascript nits, but overall the functionality is great.

src/server/modules/user/services/UrlManagementService.ts
thanhdatle: I discussed with @halfwhole on this, we were considering the option to create the tagmanagementservice too, but this will be the only function in the tagmanagement service and tags belong to url as well. So I think this should be fine. We can consider refactoring if there are more operations on tags

src/server/api/user/validators.ts
thanhdatle: good idea, will look into this.

src/server/mappers/UrlMapper.ts
thanhdatle: it will just change tags[i].tagString to tagMapper.persistenceToDTO(tags[i]). Do we really want to do that?

src/server/modules/user/index.ts
thanhdatle: yes, will do

thanhdatle: interestingly, my IDE doesn't highlight this as good as Github

src/server/modules/user/UserController.ts
thanhdatle: I thought about it too but that can't happen since we already have the Joi validation

src/server/modules/user/interfaces/UrlManagementService.ts
thanhdatle: let's discuss offline

src/server/repositories/UserRepository.ts
thanhdatle: the function returns [any, any] so it doesn't really add much more information even with the return type. I will do some more research on this but making this optional for now

src/server/repositories/UrlRepository.ts
thanhdatle: agreed, will do this

src/server/repositories/UrlRepository.ts
thanhdatle: agreed, will do this

src/server/util/array.ts
thanhdatle: I didn't know this exist, will try it out

src/server/repositories/UrlRepository.ts
thanhdatle: good idea, will do

src/server/repositories/UrlRepository.ts
thanhdatle: addTags is actually better, will do this as well

src/server/mappers/UrlMapper.ts
gweiying: I was thinking about `tags.map((tag) => tagMapper.persistenceToDTO(tag))`

src/server/models/tag.ts
```diff
@@ -0,0 +1,65 @@
...
+// For sequelize define
+type TagTypeStatic = typeof Sequelize.Model & {
+  new (values?: object, options?: Sequelize.BuildOptions): TagType
+}
+
+export const Tag = <TagTypeStatic>sequelize.define(
+  'tag',
+  {
+    tagString: {
+      type: Sequelize.STRING(25),
```
gweiying, line 20: @halfwhole raised a good point, wanted to suggest not constraining this here so that we don't have to perform future migrations if we want to support longer tag lengths

src/server/models/tag.ts
```diff
@@ -0,0 +1,65 @@
...
+      allowNull: false,
+      unique: true,
+      validate: {
+        is: TAG_STRING_REGEX,
+      },
+    },
+    tagKey: {
+      type: Sequelize.STRING(25),
```
gweiying, line 28: @halfwhole raised a good point, wanted to suggest not constraining this here so that we don't have to perform future migrations if we want to support longer tag lengths

src/server/api/user/validators.ts
thanhdatle: done

src/server/mappers/UrlMapper.ts
thanhdatle: makes sense, done

src/server/modules/user/index.ts
thanhdatle: done

src/server/util/array.ts
thanhdatle: done

src/server/repositories/UrlRepository.ts
thanhdatle: done

src/server/repositories/UrlRepository.ts
thanhdatle: done

src/server/models/tag.ts
thanhdatle: 🤔 okay, will change this to 255

gweiying:
Btw @thanhdatle just wanted to make sure that you saw @halfwhole's comments here too https://github.com/opengovsg/GoGovSG/pull/1917

src/server/modules/user/services/UrlManagementService.ts
thanhdatle: done

src/server/repositories/UrlRepository.ts
```diff
@@ -93,15 +93,16 @@ export class UrlRepository implements UrlRepositoryInterface {
...
-            }
-          }),
-        )
+        const tags: TagType[] = []
+        tagCreationResponses.forEach((response) => {
+          const [tag, _] = response
+          if (tag) {
+            tags.push(tag)
+          }
+        })
```
gweiying, line 107: Just wondering - is there ever a scenario in which `if (tag)` returns false? My understanding is if there was any tag that failed to be created in `Tag.findOrCreate`, then the whole transaction rolls back. 

If so, we can simplify the above code to:

```
        const tags: TagType[] = tagCreationResponses.map(([tag, _]) => tag)
```

src/server/modules/user/interfaces/UrlManagementService.ts
```diff
@@ -11,6 +11,7 @@ export interface UrlManagementService {
     shortUrl: string,
     longUrl?: string,
     file?: GoUploadedFile,
+    tags?: string[],
```
gweiying, line 14: small nit, can we clarify if this is meant to be `tags (type TagType)` or `tagStrings (type string)`?

src/server/modules/user/index.ts
```diff
@@ -20,6 +20,10 @@ type LinkInformationProperties = {
 
 type ShortUrlOperationProperty = UserIdProperty & ShortUrlProperty
 
+type OptionalTagsProperty = {
+  tags?: string[] | undefined
```
gweiying, line 24: nit, `tags?: string[] | undefined` is equivalent to `tags?: string[]`

src/server/repositories/UrlRepository.ts
```diff
@@ -23,9 +24,13 @@ import { SearchResultsSortOrder } from '../../shared/search'
 import { urlSearchVector } from '../models/search'
 import { DirectoryQueryConditions } from '../modules/directory'
 import { extractShortUrl, sanitiseQuery } from '../util/parse'
+import { Tag, TagType } from '../models/tag'
+import { TagRepositoryInterface } from './interfaces/TagRepositoryInterface'
 
 const { Public, Private } = FileVisibility
 
+export const tagSeparator = ';'
```
gweiying, line 32: nit, can we use capital letters for the constant

export const TAG_SEPARATOR = ';'

src/server/repositories/TagRepository.ts
```diff
@@ -0,0 +1,68 @@
...
+          )
+        : []
+      const newTags: TagType[] = []
+      tagCreationResponses.forEach((response) => {
+        const [tag, _] = response
+        if (tag) {
+          newTags.push(tag)
+        }
+      })
+      return newTags
```
gweiying, line 64: Just wondering - is there ever a scenario in which `if (tag)` returns false? My understanding is if there was any tag that failed to be created in `Tag.findOrCreate`, then the whole transaction rolls back. 

If so, we can skip the `if (tag)` check and simplify the above code to:

```
        const newTags: TagType[] = tagCreationResponses.map(([tag, _]) => tag)
```

src/server/repositories/UrlRepository.ts
```diff
@@ -54,28 +63,58 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+          ? this.fileBucket.buildFileLongUrl(file.key)
+          : properties.longUrl,
+        isFile: !!file,
+        tagStrings,
+      }
+      const url = await Url.create(urlStaticDTO, {
+        transaction: t,
+      })
+      if (properties.tags) {
+        const tagCreationResponses = await Promise.all(
```
gweiying, line 90: Should we use the new upsert function here?

src/server/repositories/UrlRepository.ts
```diff
@@ -93,44 +132,67 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+      // Do a fresh read which eagerly loads the associated tags field.
+      return Url.scope(['defaultScope', 'getClicks', 'getTags']).findOne({
+        where: { shortUrl },
+        transaction: t,
+      })
     })
-
     this.invalidateCache(shortUrl)
-
+    if (!newUrl) throw new Error('Newly-updated url is null')
```
gweiying, line 195: how would this come about? sounds like something we should catch earlier?

gweiying (changes requested):

src/server/repositories/UrlRepository.ts
thanhdatle: done

src/server/modules/user/index.ts
thanhdatle: done

src/server/repositories/UrlRepository.ts
thanhdatle: agreed, done

src/server/repositories/UrlRepository.ts
thanhdatle: There are some validations at the DB side, so imagine we have another service calling the upsertTags without having the input validated by Joi object, the function will return empty tag object

src/server/repositories/TagRepository.ts
thanhdatle: same as above

src/server/repositories/UrlRepository.ts
thanhdatle: There could be a failed transaction and the cause this to be null.

src/server/repositories/UrlRepository.ts
gweiying: ok sounds good

src/server/repositories/UrlRepository.ts
gweiying: got it - how is this handled downstream? maybe the error should be more informative like "Url update failed, transaction `t` rolled back" or something?

thanhdatle: I am following the pattern in createURL right now, I don't think it's necessary to show to end user the message like `transaction failed, etc`. We should already have some logs in RDS to tell us what exactly has happened.

gweiying (commented):
Looks good to me.

Just wanted to tag @halfwhole for another review since you left comments on the earlier PR #1919 too

src/server/repositories/UrlRepository.ts
gweiying: sure sounds good for now

src/server/models/tag.ts
```diff
@@ -0,0 +1,65 @@
...
+      unique: true,
+      validate: {
+        is: TAG_STRING_REGEX,
+      },
+    },
+    tagKey: {
+      type: Sequelize.STRING(255),
+      allowNull: false,
+      validate: {
+        is: TAG_KEY_REGEX,
```
halfwhole, line 31: should we validate the lengths for `tagString` and `tagKey` as well? i'm also wondering if it helps (if it's even possible) to have a custom function to validate that `tagKey` is indeed the lowercase version of `tagString`

test/end-to-end/DrawerLogin.test.ts
```diff
@@ -157,10 +157,6 @@ test.before(async (t) => {
   // login to the transfer email to activate it, then logout
   await LoginProcedure(t, transferEmail)
 
-  if (await userModal.exists) {
-    await t.click(userModalCloseButton)
-  }
-
```
halfwhole, line 163: nit: would you mind removing the imports for these as well! 🙇

halfwhole (commented):
am only partially done with reviewing rn sorry :( but have some comments below for now. have checked in with the earlier PR #1917 and all comments from there have been resolved!

also I'd like to ask sometime (perhaps offline) to understand the role of the `tagStrings` column on the `url` table, and why denormalisation is preferred there

src/shared/util/validation.ts
```diff
@@ -7,6 +7,12 @@ export const WHITELIST = [new RegExp('^http://localhost:4566')]
 
 export const SHORT_URL_REGEX = /^[a-z0-9-]+$/
 
+export const TAG_KEY_REGEX = /^[a-z0-9-]+$/
+
+export const TAG_STRING_REGEX = /^[A-Za-z0-9-_]+$/
```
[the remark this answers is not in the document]
thanhdatle: done

thanhdatle:
> am only partially done with reviewing rn sorry :( but have some comments below for now. have checked in with the earlier PR #1917 and all comments from there have been resolved!
> 
> also I'd like to ask sometime (perhaps offline) to understand the role of the `tagStrings` column on the `url` table, and why denormalisation is preferred there

The `TagStrings` is in the url table so that the tag changes can be reflected in `url_history` table

src/server/repositories/UrlRepository.ts
```diff
@@ -93,44 +114,67 @@ export class UrlRepository implements UrlRepositoryInterface {
...
     }
-
-    const newUrl: UrlType = await sequelize.transaction(async (t) => {
+    const urlDto = this.urlMapper.persistenceToDto(url)
+    const newUrl = await sequelize.transaction(async (t) => {
+      if (
+        changes.tags &&
+        !_.isEqual(_.sortBy(urlDto.tags), _.sortBy(changes.tags))
+      ) {
+        const newTags = await this.tagRepository.upsertTags(changes.tags, t)
```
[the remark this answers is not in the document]
thanhdatle: the reason why we don't delete is that, user might want to see tags created before.
Making this optional for now, we can discuss with PM further.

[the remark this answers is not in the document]
halfwhole: is it possible to retrieve the list of previous tags from link history? that could remove the need from storing them directly in the `tags` table

src/server/modules/user/services/UrlManagementService.ts
```diff
@@ -58,26 +59,24 @@ export class UrlManagementService implements interfaces.UrlManagementService {
           mimetype: file.mimetype,
         }
       : undefined
-
     // Success
-    const result = await this.urlRepository.create(
+    return this.urlRepository.create(
```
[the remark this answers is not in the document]
thanhdatle: this function returns a promise and get awaited in the controller.

src/server/modules/user/services/UrlManagementService.ts
```diff
@@ -128,11 +127,9 @@ export class UrlManagementService implements interfaces.UrlManagementService {
     }
 
     // Success
-    const result = await this.urlRepository.update(url, {
+    return this.urlRepository.update(url, {
```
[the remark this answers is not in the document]
thanhdatle: sam here

src/server/modules/user/services/UrlManagementService.ts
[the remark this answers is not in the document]
halfwhole: just checked the docs for async functions, and apparently both methods are equivalent - if it's not a promise, async wraps it in one, and if it's already a promise it doesn't do anything extra. TIL :O

sorry then pls ignore my change, it's not necessary!

halfwhole:
> The `TagStrings` is in the url table so that the tag changes can be reflected in `url_history` table

would it be possible to still get `tagStrings` for `url_histories` without mirroring it under `urls` (e.g. by modifying the create/update hooks to pull it from another source)? though understand that it can have downsides too - it could make the implementation more complicated, possibly incur extra reads, or make the columns in `url_histories` different from the columns in `urls`, which might not be great

also wondering what you think about representing `tagStrings` as a Postgres array, rather than serializing it with `;` - seems more semantically correct but it's not standard SQL sadly!

src/server/models/tag.ts
thanhdatle: I think we don't have to do these now

test/end-to-end/DrawerLogin.test.ts
thanhdatle: done

test/server/repositories/UrlRepository.test.ts
```diff
@@ -236,21 +356,62 @@ describe('UrlRepository', () => {
...
+          { shortUrl: baseShortUrl },
+          { description, tags: newTags },
+        ),
+      ).resolves.toStrictEqual(expect.objectContaining({ isFile: false }))
+      expect(findOne).toHaveBeenCalledWith({
+        where: { shortUrl: baseShortUrl },
+      })
+      expect(putObject).not.toHaveBeenCalled()
+      expect(putObjectAcl).not.toHaveBeenCalled()
+      expect(tagRepository.upsertTags).toHaveBeenCalledTimes(1)
```
halfwhole, line 384: perhaps a bit more detail on what `upsertTags` is called with? is it also appropriate to check on `setTags` here?

src/server/modules/user/interfaces/TagManagementService.ts
```diff
@@ -0,0 +1,9 @@
+import { UserTagsQueryConditions } from '../../../repositories/types'
+
+interface TagManagementServiceInterface {
```
halfwhole, line 3: slight note on naming here: should we name this `TagManagementService`, in line with both the file name, and how the interface for the similar URL management service is called `UrlManagementService` too? (and yet the repository interfaces are called `<Repository>Interface`... how confusing @.@)

test/server/repositories/UserRepository.test.ts
```diff
@@ -235,59 +242,27 @@ describe('UserRepository', () => {
...
-          urls: [],
-          count: 0,
-        },
-      )
-      expect(scope).toHaveBeenCalledWith([
-        'defaultScope',
-        {
-          method: ['urlsWithQueryConditions', conditions],
-        },
-      ])
```
halfwhole, line 273: was this test removed?
also thinking if we can test for some functionalities in `findUrlsForUser`, like the tags being fetched and returned along with the URLs!

halfwhole (commented):
basically looks good to me! just a few more comments, but i'm no good at tests so forgive me if I leave weird/mistaken test-related comments ><

src/server/repositories/UrlRepository.ts
```diff
@@ -54,28 +62,41 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+          ? this.fileBucket.buildFileLongUrl(file.key)
+          : properties.longUrl,
+        isFile: !!file,
+        tagStrings,
+      }
+      const url = await Url.create(urlStaticDTO, {
+        transaction: t,
+      })
+      if (properties.tags) {
+        const tags = this.tagRepository.upsertTags(properties.tags, t)
```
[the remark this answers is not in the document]
thanhdatle: sorry, I missed this after some refactoring 👍

src/server/models/tag.ts
halfwhole: ok will just KIV this then!

src/server/repositories/UrlRepository.ts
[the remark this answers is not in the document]
thanhdatle: would be expensive to query from that table actually, since tags are stored as joined string there.

src/server/modules/user/UserController.ts
```diff
@@ -184,7 +220,9 @@ export class UserController {
...
```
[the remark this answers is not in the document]
thanhdatle: pending for a discussion with PM on searching link by tags

src/server/modules/user/__tests__/UserController.test.ts
```diff
@@ -399,6 +488,70 @@ describe('UserController', () => {
...
+      const res: any = httpMocks.createResponse()
+      res.ok = jest.fn()
+      res.badRequest = jest.fn()
+      const result = { urls: [], count: 0 }
+      urlManagementService.getUrlsWithConditions.mockResolvedValue(result)
+
+      await controller.getUrlsWithConditions(req, res)
+      expect(res.badRequest).toHaveBeenCalledTimes(1)
+    })
+
```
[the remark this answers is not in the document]
thanhdatle: we can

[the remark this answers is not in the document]
thanhdatle: done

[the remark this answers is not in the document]
thanhdatle: done

test/server/repositories/TagRepository.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  const userId = 2
+  const searchText = 'tag1'
+  const scope = jest.spyOn(tagModelMock, 'scope')
+  const findAll = jest.fn()
+
+  beforeEach(() => {
+    findAll.mockReset()
+    scope.mockReset()
+  })
+
```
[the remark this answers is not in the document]
thanhdatle: done

test/server/repositories/UrlRepository.test.ts
```diff
@@ -136,14 +160,46 @@ describe('UrlRepository', () => {
...
         },
         expect.anything(),
       )
-      expect(scope).toHaveBeenCalledWith(['defaultScope', 'getClicks'])
+      expect(scope).toHaveBeenCalledWith([
+        'defaultScope',
+        'getClicks',
+        'getTags',
+      ])
       expect(putObject).not.toHaveBeenCalled()
```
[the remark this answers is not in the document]
thanhdatle: done

test/server/repositories/UrlRepository.test.ts
```diff
@@ -136,14 +160,46 @@ describe('UrlRepository', () => {
...
+        },
+        expect.anything(),
+      )
+      expect(scope).toHaveBeenCalledWith([
+        'defaultScope',
+        'getClicks',
+        'getTags',
+      ])
+      expect(putObject).not.toHaveBeenCalled()
+      expect(baseUrlWithTags.addTags).toHaveBeenCalledTimes(1)
```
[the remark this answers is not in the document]
thanhdatle: this can be handled by end to end test

test/server/repositories/UrlRepository.test.ts
thanhdatle: done

src/server/modules/user/interfaces/TagManagementService.ts
thanhdatle: Can be named either IService or ServiceInterface.

test/server/repositories/UserRepository.test.ts
thanhdatle: done

src/server/modules/user/interfaces/TagManagementService.ts
halfwhole: looks good for now, will note down for future review!

halfwhole (approved):
lgtm! 🎉 

we're still pending input for some of the things above (e.g. how exactly searching for links by multiple tags works), but these can be addressed later on

also: is it safe to merge into `develop` directly? i created a separate feature branch `feat/tags`, but don't have to use it if we don't need to

gweiying (approved):
lgtm!

PR #1962, opened 2022-09-13
feat: bulk - backend validation

jimvae:
## Problem

This is the backend portion for determining if the csv file upload is valid. More details can be found [here](https://www.notion.so/opengov/c622b311c6ae4ef0b416ce47decd0874?v=8df8c11db2944cd08ffc774e3fbae392&p=577c91546dcc4f349e1e6927d5476fdc&pm=s).

## Solution

Created an endpoint for csv validation.
<h2>Request</h2>
<pre><code class="language-jsx">POST /api/user/url/bulk-link

BODY: 
	form-data:
		key: file
		value: &lt;filename.csv&gt;
</code></pre>
<h2>Response</h2>
<pre><code class="language-jsx">{
  isValid: boolean,
}
</code></pre>

<h2>Validations</h2>
<h3>File Size (frontend and backend)</h3>
<p>The max .csv file size is 0.2 MB. This takes into account the current limit of 1000 links</p>
<ul>
<li>Calculation
<ul>
<li>Average size of one URL: 62-120 characters (we take the ceiling for this)
<ul>
<li>
<p>Supported characters by browsers</p>

</li>
</ul>
</li>
<li>Row header: “Original URL that needs to be shortened” ~ 39 characters</li>
<li>Maximum number of characters: 1000 * 120 + 39 = 120039 characters ~ 0.2 MB</li>
</ul>
</li>
</ul>
<h3>File Type  (frontend and backend)</h3>
<li>Only .csv files are allowed. Otherwise the server will throw a 415 Status Code</p>
<h3>File Count  (frontend and backend)</h3>
<li>Only one file is allowed. Otherwise the server will throw a 422 Status Code</p>
<h3>Virus Detection  (backend)</h3>
<li>The VirusScanService scans the file for any malicious entities. If there is a virus, the server will throw a 400 Status Code</p>
<h3>Link Count (backend)</h3>
<li>Uploaded csv file cannot have more than 1000 links</p>
<h3>Template Format (backend)</h3>
<li><a href="https://www.papaparse.com/">papaparse</a> library  was used for parsing the csv file.
<ul>
<li>No additional headers</li>
<li>longURL should be valid
<ul>
<li>notblacklisted</li>
<li>isHttps</li>
<li>isPrintableAscii</li>
</ul>
</li>
<li>Input rows should not be empty</li>
<li>Only one column per row</li>
</ul>
<h3>URL Threat Validation (backend)</h3>
<ul>
<li>Check if the link is malicious</li>
</ul>
<h3>Circular Redirects(backend)</h3>
<ul>
<li>Check if the link points back to the same domain</li>
</ul>

**New dependencies**:

- `papaparse": "^5.3.2`

**New dev dependencies**:

- `@types/papaparse`


src/server/api/user/index.ts
```diff
@@ -65,12 +83,31 @@ router.post(
...
   userController.createUrl,
 )
 
+/**
+ * Endpoint for a user to check validity of csv uploaded
+ * In order to create a file URL, provide the `isFile` parameter set to `true`, and supply
+ * the binary in a multipart/form-data under the field titled `file`. The longUrl
+ * parameter in the request body must point to the s3 bucket with the correct key. For
+ * example, a file upload request for the shortUrl named `test` should have `longUrl` set
+ * to `https://file[-staging].go.gov.sg/test`.
```
gweiying, line 99: Not sure if this comment is outdated?

src/server/modules/threat/FileCheckController.ts
```diff
@@ -35,13 +35,54 @@ export class FileCheckController {
...
       if (!(await this.fileTypeFilterService.hasAllowedType(file))) {
         res.unsupportedMediaType(jsonMessage('File type disallowed.'))
         return
       }
+    }
+
+    next()
+  }
+
+  public csvExtensionCheck: (
```
gweiying, line 48: The method seems to have quite abit of reused code. Instead of creating a special function to validate csv, what about [passing a parameter to the middleware](https://tsmx.net/express-middleware-function-with-custom-parameters/) e.g. `FileCheckController.fileExtensionCheck('csv')`

src/server/api/user/index.ts
```diff
@@ -65,12 +83,31 @@ router.post(
...
+ */
+
+router.post(
+  '/url/bulk-link',
+  bulkCSVUploadMiddleware,
+  preprocessPotentialIncomingFile,
+  fileCheckController.csvExtensionCheck,
```
gweiying, line 106: See comment below about middleware with params

src/server/modules/threat/services/FileTypeFilterService.ts
```diff
@@ -50,12 +50,23 @@ export class FileTypeFilterService implements interfaces.FileTypeFilterService {
...
-    async ({ name, data }) => {
-      const fileType = await FileType.fromBuffer(data)
-      const extension = fileType?.ext || `${`${name}`.split('.').pop()}`
-      return this.allowedFileExtensions.includes(extension)
+  hasAllowedType: (
+    file: {
+      name: string
+      data: Buffer
+    },
+    inputExtension?: string,
```
gweiying, line 58: Nit/ naming suggestion, can we rename this to `allowedType(s)/ allowedExtensions` so that it's consistent with the function name? 

Also a suggestion, maybe you can generalise this too so that somebody can pass in an array of allowed input extensions, instead of just a single one.

src/shared/constants.ts
```diff
@@ -1,2 +1,3 @@
+export const MAX_CSV_UPLOAD_SIZE = 200000 // 0.2 MB
```
gweiying, line 1: nit, can we use standardise on using the binary size 1024 instead of 1000? would be good to rewrite this as either `200*1024 // 200KB` or `0.2*1024*1024 // 0.2MB`

gweiying (changes requested):
In general the PR looks good! Thanks for refactoring the FileCheckController. 

Just a small suggestion about file organisation. Currently, `CsvCheckController` is in  `modules/threat` - logically there's no threat-checking in the csv, it's simply a validation controller for our bulk endpoint. I'm thinking that we can create a new modules folder, `modules/bulk` and move the controller there. What do you think of this?

Assuming we do move the controller there, I'm also wondering if whether we should think of it as a `CsvCheckController` with a single method to validate the csv, or a `BulkController` with multiple methods, one of which is to validate the csv, another to start the Lambda, etc. What do you think? Both have their merits. Keen to hear your thoughts

src/server/api/user/index.ts
jimvae: my bad  and updated the comment!

src/server/modules/threat/FileCheckController.ts
```diff
@@ -22,7 +22,33 @@ export class FileCheckController {
     this.virusScanService = virusScanService
   }
 
-  public checkFile: (
+  public fileExtensionCheck =
+    (allowedExtensions?: string[]) =>
+    async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+      const file = req.files?.file
+
+      if (Array.isArray(file)) {
```
gweiying, line 30: This logic seems to be duplicated multiple times, can we abstract this out?

gweiying (commented):
overall looks good to me, just one nit about duplicated logic for checking file arrays!

src/server/modules/threat/FileCheckController.ts
jimvae: abstracted it out as another middleware

Issue #1927, opened 2022-08-18
Introduce E2E test automation for Directory section of Release Checklist


yong-jie:
**Is your feature request related to a problem? Please describe.**
The Directory tests in the release checklist is very tedious and requires a lot of click ops. It also looks quite easily automatable via our existing e2e test framework.

**Describe the solution you'd like**
Consider extending our testcafe e2e test suite to help with directory tests.

PR #1963, opened 2022-09-14
Feat/e2e/directory page test

KoalaTasha:
## Problem

Directory section of Release Checklist 

Closes [#1927 ](https://github.com/opengovsg/GoGovSG/issues/1927)

## Solution

Introduce E2E test automation for the Directory section of the Release Checklist


gweiying (approved):
In general the PR looks good to me, the functionality is good and the coding style is very clean! Thanks for all the detailed inline comments.

Just braindumping some concerns I have about this PR (and tests in general) below, but I think they are non-blocking and require closer thought. Would love to hear what you think!

1. Populating links within the Directory tests couples the success of the Directory tests to the success of link creation and retrieval. This has the unintended side effect that a Directory test could fail for reasons outside of the Directory flow. This would also mean more maintenance work in the future, say if the link creation flow changes. One alternative to consider is, instead of adding links and retrieving them within the web console, to run a script to seed the database before running the tests. I think this isn't a high priority task since the functionality here is complete, so I will just open it as an issue for now.

2. `LoginProcedure` runs on every test, which makes the tests much longer. This seems to be a quirk of TestCafe to not allow nested tests, otherwise we could reuse the same testController across the whole test suite, and not have to log in every time. Something we could also explore is having a global hook to log us in by default (such that we stay logged in), but disable it for the Login suite. What do you think? I'll open this as an issue as I also think this is not blocking for now - it probably doesn't hurt if our E2E tests are 20 seconds longer since they already take 8 minutes to complete. 

Thanks for working on this feature

KoalaTasha:
> In general the PR looks good to me, the functionality is good and the coding style is very clean! Thanks for all the detailed inline comments.
> 
> Just braindumping some concerns I have about this PR (and tests in general) below, but I think they are non-blocking and require closer thought. Would love to hear what you think!
> 
> 1. Populating links within the Directory tests couples the success of the Directory tests to the success of link creation and retrieval. This has the unintended side effect that a Directory test could fail for reasons outside of the Directory flow. This would also mean more maintenance work in the future, say if the link creation flow changes. One alternative to consider is, instead of adding links and retrieving them within the web console, to run a script to seed the database before running the tests. I think this isn't a high priority task since the functionality here is complete, so I will just open it as an issue for now.
> 2. `LoginProcedure` runs on every test, which makes the tests much longer. This seems to be a quirk of TestCafe to not allow nested tests, otherwise we could reuse the same testController across the whole test suite, and not have to log in every time. Something we could also explore is having a global hook to log us in by default (such that we stay logged in), but disable it for the Login suite. What do you think? I'll open this as an issue as I also think this is not blocking for now - it probably doesn't hurt if our E2E tests are 20 seconds longer since they already take 8 minutes to complete.
> 
> Thanks for working on this feature!!

Wow, thanks for the quick review! 😄 

Regarding your concerns
1. I think it makes a lot of sense to directly seed the database with the required links, to prevent extra maintenance work should the FE link creation flow change in the future. Thanks for creating the issue for this!
2. Hmm the extra time logging in will probably become a problem as the testing suite grows. But I think it's important to ensure that each test starts at the main landing page with the default settings (e.g. filters with default values). Logging in ensures all of this. Not sure if extra work will be required to ensure this when implementing a global hook 🤔 

Thanks for letting me work on this~

PR #1965, opened 2022-09-15
chore(datadog): refactor custom metric names

halfwhole:
## Problem

Follow-up from #1937: names for custom metrics should be extracted and put into a separate file

## Solution

Made a small refactoring: moved custom metric names into `dogstatsd.ts`

### DIFF
--- a/src/server/modules/user/services/UrlManagementService.ts
+++ b/src/server/modules/user/services/UrlManagementService.ts
@@ -6,7 +6,10 @@ import {
   UrlsPaginated,
   UserUrlsQueryConditions,
 } from '../../../repositories/types'
-import dogstatsd from '../../../util/dogstatsd'
+import dogstatsd, {
+  SHORTLINK_CREATE,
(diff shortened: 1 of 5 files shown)


thanhdatle (approved):

src/server/modules/user/services/UrlManagementService.ts
```diff
@@ -60,9 +63,9 @@ export class UrlManagementService implements interfaces.UrlManagementService {
           mimetype: file.mimetype,
         }
       : undefined
+
     // Success
-    dogstatsd.increment('shortlink.create', 1, 1, [`isfile:${!!file}`])
-    return this.urlRepository.create(
+    const result = await this.urlRepository.create(
```
halfwhole, line 68: just rebased on `develop` to resolve some merge conflicts. reordered this bit to put `dogstatsd.increment` only after a URL was successfully created, hope it's okay!

PR #1972, opened 2022-09-19
feat(tagging): add tags to links on user page

halfwhole:
Note: this PR contains the API integrations for searching by tags as well

## Problem

We want users to be able to do these on the main user page:
1. See the tags for each URL
2. Search URLs by tags when clicking on a tag

## Solution

**Features**:

Added buttons for URL tags on the user page
- URL tags will only render on large screens, but not on small ones (see desktop/mobile screenshots below)

Added the API integrations for searching URLs by tags
- Clicking on a tag will amend the search bar to search by tags, instead of by link. If there's already pre-existing tags in the search bar, clicking on a new tag will append `;` plus the new tag to the end

**Improvements**:

- Refactored `setUrlTableConfig` to take in a *partial* URL table config, instead of the full thing. IMO this matches the intention better, and prevents the need for inappropriate/awkward type casting with `as UrlTableConfig`
- Removed the local state `query` in the search input component, and instead peg the search input text directly to global redux state `state.user.tableConfig.searchText`

## Before & After Screenshots

**AFTER**:

Desktop:
Screenshot 2022-09-20 at 1 08 57 PM

Mobile:
Screenshot 2022-09-20 at 1 09 31 PM

## Tests

- Still pending E2E tests


src/client/user/components/UserLinkTable/ToolBar/SearchInput/index.tsx
```diff
@@ -138,11 +140,10 @@ const SearchInput = () => {
       applySearch()
     }, 500)
     return () => clearTimeout(timeoutId)
-  }, [query])
+  }, [searchText])
```
halfwhole, line 143: note: not entirely sure that this effect should be placed here, but idk how we can do better - any thoughts? `searchText` belongs to the global redux store, and yet this effect for applying the search resides within this local component. this setup may be confusing because changes made to `searchText` anywhere else from any other component, e.g. `EnhancedTableBody`, will trigger this effect here as well (so long as this component exists).

src/client/user/components/UserLinkTable/ToolBar/SearchInput/index.tsx
```diff
@@ -110,26 +110,28 @@ const textFieldHeight = useSearchInputHeight()
...
   const tableConfig = useSelector(
     (state: GoGovReduxState) => state.user.tableConfig,
   )
+  const { isTag, searchText } = tableConfig
   const searchInputHeight = useSearchInputHeight()
   const classes = useStyles({ textFieldHeight, searchInputHeight })
 
-  const changeSearchTextHandler = (e: string) => {
-    setQuery(e)
+  const setSearchText = (searchText: string) => {
```
gweiying, line 121: Just leaving a general comment here, not specific to this line but the effects in general throughout the file.

It seems a bit untidy that we are dispatching the urlTable config from multiple places throughout the code. Is this necessary?

Seems like the goal is to dispatch if something changes, but implement debouncing so that it's not constantly fetching. Is that right? If so, can we abstract that out somehow?

halfwhole: > It seems a bit untidy that we are dispatching the urlTable config from multiple places throughout the code. Is this necessary?

can't see a cleaner way to do this sadly :( any suggestions on how this can be improved?

> Seems like the goal is to dispatch if something changes, but implement debouncing so that it's not constantly fetching. Is that right? If so, can we abstract that out somehow?

indeed, though I'm not aware of any other way to do debouncing apart from using `useEffect` and cleanup functions, and I can't think of where else to place `useEffect` apart from at the component level (e.g. here in this component) - would there be other approaches we can consider?

thanhdatle (approved):

PR #1973, opened 2022-09-19
feat: search for tags in lower cap

thanhdatle:
## Problem

The backend should return tag suggestions in both upper and lower cap

Closes [insert issue #]

## Solution

SQL query to perform LIKE operation on tagKey field and searchText input needs to be in lower cap.


halfwhole (approved):
lgtm

PR #1979, opened 2022-09-22
fix: return successful response when no URLs are found

### LINKED ISSUE #1978: Backend should not return 404 errors when no URLs are found
**Describe the bug**

When querying `/api/user/url`, if zero URLs match the query, the backend responds with a 404 error with body `Urls not found`. This should not be the response returned.

When the frontend tries to parse `Urls not found` as JSON, it results in an uncaught error:

Screenshot 2022-09-22 at 12 05 40 PM

This breaks the user page whenever the user makes queries that return 0 URLs.

**Expected behaviour**

Backend should return 200 with body `{"urls":[],"count":0}` - this is the behaviour in our latest release `v1.69.0`

On the frontend, when the user has no links, we should see this:
Screenshot 2022-09-22 at 12 11 10 PM

When the user has links and makes a query returning no results, we should see this:
Screenshot 2022-09-22 at 12 23 56 PM

**Actual behaviour**

Backend returns 404 error with body `Urls not found` - this is the current behaviour in `develop`

On the frontend, when the user has no links:
Screenshot 2022-09-22 at 12 13 17 PM

When the user has links and makes a query returning no results, the page simply does not update at all.

**Additional thoughts**

- Having backend integration tests could be useful in helping to catch regressions like these
- E2E tests should test for having no links or queries returning no results

halfwhole:
## Problem

Closes #1978, see problem description in issue 

## Solution

Backend returns a successful 200 response with body `{ urls: [], count: 0 }` when there's no URLs found, instead of throwing a 404 error with `Urls not found`.

**Improvements**:

- Handle errors better on the frontend, in cases where the JSON it receives from the backend cannot be parsed

## Tests

- Updated backend tests to reflect the new changes
- Removed the test case for `throws NotFoundError on findAndCountAll without user`, because it looks to be a duplicate/didn't seem to add much value

### DIFF
--- a/src/client/user/actions/index.ts
+++ b/src/client/user/actions/index.ts
@@ -244,13 +244,12 @@ const getLinkHistory: (queryObj: ParsedUrlQueryInput) => Promise<{
     message?: string
   }
   isOk: boolean
-}> = (queryObj) => {
+}> = async (queryObj) => {
   const query = querystring.stringify(queryObj)
-
-  return get(`/api/link-audit?${query}`).then((response) => {
-    const isOk = response.ok
-    return response.json().then((json) => ({ json, isOk }))
-  })
+  const response = await get(`/api/link-audit?${query}`)
+  const isOk = response.ok
+  const json = await response.json()
+  return { json, isOk }
 }
 
 const isGetLinkHistoryForUserSuccess: (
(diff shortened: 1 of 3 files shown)


jimvae (approved):
LGTM!  Success response is sent back to the client when no URLS are found.

Screen Shot 2022-09-26 at 1 31 02 PM

Just a small nit! Perhaps we can consider HTTP 204 when we are returning an empty list of urls.

halfwhole:
just checked again, and hope it's ok to stick with a 200 response code! from what I've looked up, 204 seems to be reseved for situations where no body is returned in the response, but here we're returning `[]` for URLs and `0` for count - 204 may lead to issues where the browser may discard the response body ([mdn docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204))

PR #1981, opened 2022-09-26
fix: revise findUrlsForUser to return urls in correct order

thanhdatle:
## Problem

findUrlsForUser function doesn't return urls according to `ORDER BY` correctly
This is due to a common problem found in Sequelize while querying many-to-many rel, this is stil unresolved by now. 
[sequelize-issue](https://github.com/sequelize/sequelize/issues/7344)
The common practice is to write raw query when encountering this bug.
I am keeping the `url_tag` table, in case we need to re-write the query with raw query.

## Solution

Revised the findUrlsForUser to search for tag by looking for subString in tagStrings column.

### DIFF
--- a/src/server/repositories/UserRepository.ts
+++ b/src/server/repositories/UserRepository.ts
@@ -140,13 +140,11 @@ export class UserRepository implements UserRepositoryInterface {
         },
       ],
     }
-    let whereConditions: any =
-      conditions.searchText.length > 0
-        ? {
-            ...searchTextCondition,
-            userId: conditions.userId,
-          }
-        : { userId: conditions.userId }
+    let whereConditions: any = { userId: conditions.userId }
+    if (conditions.searchText.length > 0) {
+      whereConditions = { ...whereConditions, ...searchTextCondition }
+    }
+
     if (conditions.tags && conditions.tags.length > 0) {
       const searchTagConditions = {
         [Op.or]: conditions.tags.map((tag) => {
--- a/src/server/repositories/UserRepository.ts
+++ b/src/server/repositories/UserRepository.ts
@@ -118,9 +118,6 @@ export class UserRepository implements UserRepositoryInterface {
       throw new NotFoundError(notFoundMessage)
     }
     const { rows, count } = urlsAndCount
-    if (!rows || count === 0) {
-      throw new NotFoundError(notFoundMessage)
-    }
     const urls = rows.map((urlType) => this.urlMapper.persistenceToDto(urlType))
     return { urls, count }
   }
--- a/src/server/repositories/UserRepository.ts
+++ b/src/server/repositories/UserRepository.ts
@@ -154,7 +154,7 @@ export class UserRepository implements UserRepositoryInterface {
           return { tagStrings: { [Op.substring]: `${tag}` } }
         }),
       }
-      whereConditions = { ...whereConditions, searchTagConditions }
+      whereConditions = { ...whereConditions, ...searchTagConditions }
     }
     if (conditions.state) {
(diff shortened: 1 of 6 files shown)


src/server/repositories/UserRepository.ts
```diff
@@ -160,30 +141,28 @@ export class UserRepository implements UserRepositoryInterface {
...
             userId: conditions.userId,
           }
         : { userId: conditions.userId }
+    if (conditions.tags && conditions.tags.length > 0) {
+      const searchTagConditions = {
+        [Op.or]: conditions.tags.map((tag) => {
+          return { tagStrings: { [Op.substring]: `${tag}` } }
+        }),
+      }
+      whereConditions = { ...whereConditions, searchTagConditions }
```
halfwhole, line 157: was just testing it locally, searching with tags breaks without the addition of the spread operator `...` below:
```suggestion
      whereConditions = { ...whereConditions, ...searchTagConditions }
```
even with the above in place though, I'm worried that the `[Op.or]` from `searchTagConditions` will override `[Op.or]` from `searchTextCondition` defined above, if the user searches by both `searchText` and `tags`

src/server/repositories/UserRepository.ts
```diff
@@ -160,30 +141,28 @@ export class UserRepository implements UserRepositoryInterface {
...
         ? {
             ...searchTextCondition,
             userId: conditions.userId,
           }
         : { userId: conditions.userId }
+    if (conditions.tags && conditions.tags.length > 0) {
+      const searchTagConditions = {
+        [Op.or]: conditions.tags.map((tag) => {
+          return { tagStrings: { [Op.substring]: `${tag}` } }
+        }),
```
halfwhole, line 155: v small nit:
```suggestion
        [Op.or]: conditions.tags.map((tag) => (
          { tagStrings: { [Op.substring]: tag } }
        )),
```
also have a question: since we're searching based on `tagStrings` now (from urls) instead of `tagKey` from before (from tags), will it mean that our search is now case-sensitive?

thanhdatle: sorry, didn't push the latest change  😆

src/server/repositories/UserRepository.ts
```diff
@@ -116,32 +113,16 @@ export class UserRepository implements UserRepositoryInterface {
...
       ],
-      include: [includeConditions],
     })
     if (!urlsAndCount) {
       throw new NotFoundError(notFoundMessage)
     }
-    let { rows } = urlsAndCount
-    let { count } = urlsAndCount
+    const { rows } = urlsAndCount
+    const { count } = urlsAndCount
```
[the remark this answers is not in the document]
thanhdatle: done

src/server/repositories/UserRepository.ts
thanhdatle: the Op.iLike should be able to handle case insensitive search

src/server/repositories/UserRepository.ts
thanhdatle: we actually don't support searchText and Tags at the same time.

src/server/repositories/UserRepository.ts
```diff
@@ -160,30 +141,28 @@ export class UserRepository implements UserRepositoryInterface {
...
       ],
     }
-    const whereConditions: any =
+    let whereConditions: any =
       conditions.searchText.length > 0
         ? {
             ...searchTextCondition,
             userId: conditions.userId,
           }
         : { userId: conditions.userId }
```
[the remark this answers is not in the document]
thanhdatle: Sounds good, done

PR #1982, opened 2022-09-26
fix(tagging): reset tags state after link creation

halfwhole:
## Problem

After a user successfully creates a link with some tags, those same tags will show up again when the user tries to create another new link. This is probably not the behaviour that we want

## Solution

Set `tags: []` when the user state is reset. Its action is dispatched after a link creation is successful


jimvae (approved):
LGTM!

PR #1985, opened 2022-09-27
chore: remove disallowed file extensions

halfwhole:
## Problem

`FileTypeFilterService.ts` contains a whitelist of allowed file extensions. However, some of these are actually disallowed in practice, because our `file-type` library is incapable of recognising them correctly. ([here's the list](https://github.com/sindresorhus/file-type#supported-file-types) of supported file extensions by the library)

More details are on our [Notion page](https://www.notion.so/opengov/Confirm-which-file-type-uploads-are-accepted-f42b0e198f0a4086a7d610e99d71667d)

## Solution

Remove these unused and disallowed file extensions.


jimvae (approved):
LGTM!

PR #1986, opened 2022-09-27
Fix/link tagging for file upload

jimvae:
## Problem

Link tagging for file upload

## Solution

Add logic to add tags when creating link from file

## Before & After Screenshots

**BEFORE**: Tags were not included in the HTTP request
image
image

**AFTER**: 
Screen Shot 2022-09-27 at 6 15 24 PM
Screen Shot 2022-09-27 at 6 16 18 PM
Screen Shot 2022-09-27 at 6 19 08 PM

### DIFF
--- a/src/client/user/actions/index.ts
+++ b/src/client/user/actions/index.ts
@@ -768,7 +768,7 @@ const uploadFile =
     getState: GetReduxState,
   ) => {
     const {
-      user: { shortUrl },
+      user: { shortUrl, tags },
     } = getState()
     if (file === null) {
       // Sentry analytics: create link with file fail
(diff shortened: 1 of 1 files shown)


jimvae:
Also added a fix to reset the tag state every time we create a url / close the `create url modal`.

halfwhole:
for resetting the tag state after link creation, i believe #1982 should take care of it! as for resetting on close modal lgtm, though iirc shortlinks/long urls are not reset either when the modal closes

jimvae:
> for resetting the tag state after link creation, i believe #1982 should take care of it! as for resetting on close modal lgtm, though iirc shortlinks/long urls are not reset either when the modal closes

alright, reverted the commit that does this!

halfwhole (approved):
LGTM, thanks for the catch!

(note for future: we might want to refactor the actions sometime, the current layout looks a bit messy sadly)

PR #1988, opened 2022-09-28
fix(tagging): remove empty tag when tagStrings is empty

halfwhole:
## Problem

Empty tag shows up when `tagStrings` is empty.

## Solution

Ensure that there are no tags when `tagStrings` is empty.

TODO: write tests to ensure that empty `tagStrings` doesn't result in an empty tag.


halfwhole:
Closed - Dat is working on this already

PR #1990, opened 2022-09-28
fix(tagging): serialize single tags for file uploads

halfwhole:
## Problem

Currently, creating files fails when creating single tags (as opposed to 0 or many). This is due to the behaviour of formdata uploads, which does not interpret a single key-value pair as an array.

## Solution

- Serialize as JSON in the frontend, and deserialize JSON in the backend.
- Return a bad request if the JSON cannot be parsed in the backend.

(Done w pair programming with @thanhdatle)

### DIFF
--- a/src/server/api/user/index.ts
+++ b/src/server/api/user/index.ts
@@ -50,6 +50,7 @@ function preprocessPotentialIncomingFile(
   if (req.files) {
     req.body.files = req.files
     if (req.body.tags) {
+      // Tags for files sent as FormData should be deserialised from JSON format
       try {
         req.body.tags = JSON.parse(req.body.tags)
       } catch (e) {
--- a/src/server/api/user/index.ts
+++ b/src/server/api/user/index.ts
@@ -3,6 +3,7 @@ import fileUpload from 'express-fileupload'
 import { createValidator } from 'express-joi-validation'
 import { DependencyIds } from '../../constants'
 import { container } from '../../util/inversify'
+import jsonMessage from '../../util/json'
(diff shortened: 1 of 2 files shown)


halfwhole:
thanks @jimvae for the review!

src/server/api/user/index.ts
```diff
@@ -43,11 +44,19 @@ const validator = createValidator({ passError: true })
  */
 function preprocessPotentialIncomingFile(
   req: Express.Request,
-  _: Express.Response,
+  res: Express.Response,
   next: Express.NextFunction,
 ) {
   if (req.files) {
     req.body.files = req.files
+    if (req.body.tags) {
```
jimvae, line 52: nit: this will still be parsed successfully even if the  tag is not an array of string! The error will be caught by the validators (one of the subsequent middlewares)

jimvae (commented):
LGTM! Just a small nit on the preprocessing middleware

src/server/api/user/index.ts
halfwhole: yup, that sounds right to me! the validating of the exact format should probably be left to the Joi schema later on, as I understood this to be just for the preprocessing step - would you have any thoughts on how to improve this? or perhaps adding comments would help to clarify the intention here?

thanhdatle (approved):
LGTM

src/server/api/user/index.ts
jimvae: alright makes sense!

jimvae (approved):

PR #1991, opened 2022-09-28
fix: wrap backend errors in json messages

halfwhole:
## Problem

The frontend is unable to parse backend error messages correctly if they do not come in the standard JSON message format, i.e. `{ message: "..." }`

## Solution

Use `jsonMessage()` on error messages returned in the backend to wrap it in a JSON message.
- This is consistent with other instances of `res.[badRequest/notFound/serverError](...)` in the backend codebase, which all use `jsonMessage()` as well.

### DIFF
--- a/src/server/modules/user/UserController.ts
+++ b/src/server/modules/user/UserController.ts
@@ -190,7 +190,7 @@ export class UserController {
     const queryConditions = UserController.extractUrlQueryConditions(req)
     const validationResult = userUrlsQueryConditions.validate(queryConditions)
     if (validationResult.error) {
-      res.badRequest(validationResult.error.message)
+      res.badRequest(jsonMessage(validationResult.error.message))
       return
     }
     // Find user and paginated urls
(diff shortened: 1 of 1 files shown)


jimvae (commented):
nit: was trying to simulate the errors and just wanted to confirm that this is the intended behaviour if we call this  api without `limit`

https://user-images.githubusercontent.com/65295129/192831374-08c1b772-a4f9-432b-b1a4-47d8596f539b.mov


This is because under the UserController,`extractUrlQueryConditions` changes the limit to 1000 as the default

halfwhole:
from git blame, I see that the limit of 1000 has been there for a while now (last modified 2 years ago in [this diff](https://github.com/opengovsg/GoGovSG/commit/178d1632bde3f820c47f5d96d1151cf24ff11551#diff-6810b3beae6ed0f7cf02736895665caf0dd9086128d3f46a1298794e764bb18eR153-R154)) - seems like the intended default behaviour to me!

thanhdatle (approved):

jimvae:
> from git blame, I see that the limit of 1000 has been there for a while now (last modified 2 years ago in [this diff](https://github.com/opengovsg/GoGovSG/commit/178d1632bde3f820c47f5d96d1151cf24ff11551#diff-6810b3beae6ed0f7cf02736895665caf0dd9086128d3f46a1298794e764bb18eR153-R154)) - seems like the intended default behaviour to me!

Alright! other than that LGTM!

jimvae (approved):

PR #1999, opened 2022-10-03
fix(tagging): escape searching by tags with underscores

### LINKED ISSUE #1994: Search by tags does not work with underscores
**Describe the bug**

Searching for tags with underscores doesn't filter for the underscores correctly.

**To Reproduce**

1. Create some links with tags with underscores in them, and some links with tags without underscores in them
2. Go the user page and search by tags for `_`
3. See the incorrect behaviour that all links are returned, not just the links that have tags with `_` in them

**Expected behavior**

Only links with tags with `_` in them should be returned.

**Screenshot**

Screenshot 2022-09-30 at 10 52 49 AM

First row `/sniok0` shouldn't be in the search results.

**Additional context**

When searching by tags for `_`, we see the following SQL query executed in the console:

`SELECT count(*) AS "count" FROM "urls" AS "url" LEFT OUTER JOIN "url_clicks" AS "UrlClicks" ON "url"."shortUrl" = "UrlClicks"."shortUrl" WHERE ("url"."tagStrings" ILIKE '%_%') AND "url"."userId" = 2;`

This query is incorrect because the underscore `_` in the search string should be escaped: see https://stackoverflow.com/questions/38084864/how-to-escape-underscores-in-postgresql

halfwhole:
## Problem

Closes #1994, see problem description in issue

## Solution

Created function to escape wildcard characters like `_`, `%`, `\`, and used it to escape underscores in tags at the repo layer before searching. Credits to [this StackOverflow Q&A](https://stackoverflow.com/questions/42519805/how-to-escape-like-wildcard-characters-and-in-sequelize)

## Tests

Added a test to ensure that underscores are escaped correctly in the where conditions passed to Sequelize

### DIFF
--- a/src/server/repositories/UserRepository.ts
+++ b/src/server/repositories/UserRepository.ts
@@ -10,10 +10,11 @@ import { UserRepositoryInterface } from './interfaces/UserRepositoryInterface'
 import { User, UserType } from '../models/user'
 import { Mapper } from '../mappers/Mapper'
 import { DependencyIds } from '../constants'
+import { UrlClicks } from '../models/statistics/clicks'
 import { Url, UrlType } from '../models/url'
 import dogstatsd, { USER_NEW } from '../util/dogstatsd'
 import { NotFoundError } from '../util/error'
-import { UrlClicks } from '../models/statistics/clicks'
+import { escapeWildcard } from '../util/sequelize'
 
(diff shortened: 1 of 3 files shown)


gweiying (approved):
lgtm

PR #2000, opened 2022-10-03
fix(tagging): separate searchText from searchTextInput for link search

halfwhole:
## Problem

On the frontend user page, we currently do not distinguish between the *search text* (what the frontend sends out to the backend via requests) and the *search text input* (the text input on the search bar - this is what the user interacts with). 

This is problematic because there's a 500ms delay between when the user changes the *search text input*, to when the search request is actually sent out to the backend. If we store only a single variable for both the search text and its input, then in the midst of that 500ms delay, it's hard to determine if the request has been sent out or not.

That leads to some problems on the frontend. For example, on the current `develop` branch:
1. Enter a search query that returns 0 results
2. Delete the entire search query
3. Observe that the top bar (search input, link count, etc.) is immediately gone, when it shouldn't be

**Before**:

https://user-images.githubusercontent.com/41856541/193543700-2663c4ff-c25f-4a37-833a-390030733bbd.mov

## Solution

Create a new state variable for the *search text input* in the redux store, separate from the *search text*. We allow it to be manipulated via either a change in the search input text field, or by clicking a tag on a link row. 

**After**:

https://user-images.githubusercontent.com/41856541/193544418-3fb7d344-6367-4865-9b20-677603e83839.mov

(For context: the search text input was originally available under the local state variable `query`, but I mistakenly removed it in #1972. Now I've brought it back as `searchTextInput`, but put it in the global redux store, because it needs to be changed when clicking on the tag buttons as well.)
https://github.com/opengovsg/GoGovSG/blob/02672d7b7dc052b50aff13e3b6a730afd8163041/src/client/user/components/UserLinkTable/ToolBar/SearchInput/index.tsx#L67-L70

(also: I'm still not entirely happy with the way debouncing/this whole thing works, lmk if there's any better suggestions!)


gweiying (approved):
lgtm

PR #2014, opened 2022-10-07
chore(lambda): update sql scripts for transferring links with `tagStrings`

halfwhole:
## Problem

When using lambda functions to transfer links, the url history records created should contain the `tagStrings` as well

## Solution

Add `tagStrings` as a column to be inserted into `url_histories` in the SQL scripts

## Tests

- [x] Test both lambda functions for transferring links on staging

## Deploy Notes

- Post deploy: the SQL functions need to be updated


gweiying (approved):
lgtm

PR #2016, opened 2022-10-07
feat: backend/bulk tags

gweiying:
## Problem

Currently, when we bulk create links, we do not accept tags as input.

## Solution

This PR adds backend functionality for tagging on bulk creation. Specifically, this PR adds functionality to accept tags as input on bulk creation, upsert tags, and add the tag-url mapping. 

This PR also refactors SafeBrowsingService to clean up duplicated logic.


halfwhole (approved):
lgtm!

PR #1983, opened 2022-09-26
chore(tagging): add e2e tests for link tags

halfwhole:
~Note: one E2E test is still broken because of #1994 (searching by underscores) - otherwise, it's ready for review~

## Problem

We need E2E tests for the link tagging feature

## Solution

Added E2E tests for the link tagging feature
- E2E tests for drawer
- E2E tests for URL creation
- E2E tests for link history
- E2E tests for user page

PR #2021, opened 2022-10-10
chore: clean up e2e tests

halfwhole:
Should've gotten rid of some stray comments and things from #1983, sorry about that!


gweiying (approved):
lgtm

PR #2022, opened 2022-10-10
chore: add e2e tests for bulk creation

gweiying:


test/end-to-end/UrlCreation.test.ts
```diff
@@ -218,3 +229,56 @@ test('The URL searching test.', async (t) => {
...
+  await t
+    // It should show an success snackbar when a new file link has been added
+    .expect(successBulkCreation.exists)
+    .ok()
+    // The number of links should increase by numLongUrls
+    .expect(await getLinkCount())
+    .eql(expectedLinkCount)
+    // // The new short url should be highlighted on the user's links table when a new file link is created
+    // .expect(urlTable.child(0).getStyleProperty('background-color'))
+    // .eql('rgb(249, 249, 249)') // #f9f9f9 in rgb
```
halfwhole, line 260:
```suggestion
```

test/end-to-end/UrlCreation.test.ts
```diff
@@ -218,3 +229,56 @@ test('The URL searching test.', async (t) => {
...
+    .pressKey('enter')
+    .click(createLinkButton.nth(2))
+
+  await t
+    // It should show an success snackbar when a new file link has been added
+    .expect(successBulkCreation.exists)
+    .ok()
+    // The number of links should increase by numLongUrls
+    .expect(await getLinkCount())
+    .eql(expectedLinkCount)
```
halfwhole, line 257: for some reason, on my local environment `await getLinkCount()` still returns the old value before the 100 new links were created :/

it doesn't appear to be waiting for the number of links to update - but if I add e.g. an extra second of wait time, then it passes on local dev

halfwhole (approved):
LGTM! just a small concern about getting the link count, which appears to fail on local dev

test/end-to-end/UrlCreation.test.ts
gweiying: Interesting! Thanks for finding this bug - it works on my computer but I think we can add a small wait time anyway just preemptively to reduce flakiness

gweiying: added here [6987340](https://github.com/opengovsg/GoGovSG/pull/2022/commits/698734090f0c532493751b9023e7bbf71ec50754)

PR #2025, opened 2022-10-12
fix(tagging): allow tags to be added on space, comma, clickaway

halfwhole:
## Problem

Currently, adding tags requires the user to press enter, which can be non-obvious and unintuitive. 

## Solution

Allow tags to be added when the user presses enter, space, comma, or clicks away from the tags input.

~But a slight ordering problem happens when the user clicks away from the tags input to the "create link" or "save" button at one go:~
~- When the user clicks "create link" on the create link form, first the tag input is added, then the link is created;~
~- When the user clicks "save" on the edit link form, first the link is saved, then the tag input is added.~

edit: above problem is fixed, always adds tags before saving. but the solution might be a little unclean though

Some extra fixes: slight amendments to logic for disabling save/create link buttons

### DIFF
--- a/src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
+++ b/src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
@@ -19,6 +19,7 @@ import {
   isValidLongUrl,
   isValidShortUrl,
   isValidTag,
+  isValidTags,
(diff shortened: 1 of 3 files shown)


thanhdatle (approved):
LGTM

PR #2028, opened 2022-10-13
fix(tagging): clear tag input when tag is removed

### LINKED ISSUE #2027: Link tagging: removing tag while tag input exists clears tag input
**Describe the bug**
On the create or the edit link form, removing a tag while some tag input exists clears the tag input

**To Reproduce**
1. Go to the create new link form
2. Add a tag
3. Enter some text on the tag input field
4. Click 'x' on the previously added tag
4. See that the tag input field has been cleared

**Expected behavior**
The tag input should not be cleared

_Alternatively, we could just live with this and clear the input whenever the user deletes an existing tag - this makes sense to me as well. What do we think?_

**Screenshots**
https://user-images.githubusercontent.com/41856541/195317110-c80f5f51-3c1c-4975-953c-f297b2331870.mov

**Additional context**
It's probably a result of `freeSolo` from MUI autocomplete behaving weirdly, see the comment for `event.stopPropagation()`:

https://github.com/opengovsg/GoGovSG/blob/d9a38baa4969cbf4d4bcb2c46b3806be71575825/src/client/user/widgets/TagsAutocomplete.tsx#L126-L134

halfwhole:
## Problem

The decision is to just stick with clearing the tag input when a tag is removed. However, the underlying React state for the tag input is not cleared even when the input looks to be cleared, so we need to clear the state as well.

Closes #2027

## Solution

Clear tag input state (and tag suggestions) when the user removes a tag.

### DIFF
--- a/src/client/user/widgets/TagsAutocomplete.tsx
+++ b/src/client/user/widgets/TagsAutocomplete.tsx
@@ -114,7 +114,11 @@ export default function TagsAutocomplete({
               <FormTag
                 key={tag}
                 tag={tag}
-                onClose={() => setTags(tags.filter((t) => t !== tag))}
+                onClose={() => {
+                  setTags(tags.filter((t) => t !== tag))
+                  setTagInput('')
+                  resetTagSuggestions()
+                }}
               />
             ))
           }


gweiying (approved):
lgtm

PR #2030, opened 2022-10-13
Handle malformed JSON with 400 error

### LINKED ISSUE #1961: Handle 500 error with invalid JSON input
**Describe the bug**

When making a POST request with malformed JSON input, the server gives a 500 error.

**To Reproduce**

Log in, then make a POST request to an endpoint like `/api/user/url` with some random invalid JSON input like `{ _12o3i }`

**Expected behavior**

Server should handle it gracefully and return 400 bad request instead of 500 internal server error.

developerdenesh:
## Problem

Closes  #1961

## Solution

Check the incoming error code. If 400 is received, return 400 back

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
Tested using: `curl -XPOST -s -H 'Content-Type: application/json' -d '{ _12o3i }' http://localhost:3000/api/url/user`

The following result was received:
Screenshot 2022-10-13 at 2 37 33 PM

## Tests

### DIFF
--- a/src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
+++ b/src/client/user/components/CreateUrlModal/CreateLinkForm.tsx
@@ -329,6 +329,7 @@ const CreateLinkForm: FunctionComponent<CreateLinkFormProps> = ({
               tagInput={tagInput}
               setTagInput={setTagInput}
               disabled={isUploading || tags.length >= MAX_NUM_TAGS_PER_LINK}
+              fixHelperTextPosition={false}
             />
           </div>
           <Button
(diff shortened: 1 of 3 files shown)


src/server/index.ts
```diff
@@ -223,6 +223,12 @@ initDb()
         res.badRequest(jsonMessage(err.error.toString()))
         return
       }
+
+      if (err.status === 400) {
+        res.badRequest(jsonMessage("Bad Request. JSON is malformed"))
+        return;
+      }
```
halfwhole, line 230: hey @developerdenesh, thanks for contributing! this works well, but a possible concern is that errors with status 400 that are not necessarily JSON parsing errors will be caught and handled under this block as well.

perhaps we could strengthen the if condition to narrow it down to JSON errors? I've just checked and found an answer [here](https://stackoverflow.com/a/53049009) that tries to do just that, as well as another possible candidate `type: 'entity.parse.failed'` described [here](https://expressjs.com/en/resources/middleware/body-parser.html#errors) that we may want to check on as well - hope it helps!

developerdenesh:
Pull request has been shifted to: https://github.com/opengovsg/GoGovSG/pull/2042

src/server/index.ts
developerdenesh: That's true. Have made the changes in #2042

PR #1993, opened 2022-09-29
Feat/bulk backend/functionality

gweiying:
Tasks
- [x] safe browsing in bulk @jimvae 
  - [x] create SafeBrowsingService.lookupBulk (if more than 500, you need to paginate)
  - [x] hook up SafeBrowsingService.lookupBulk to pipeline
- [x] url enum column and code changes @gweiying 
- [x] test dd logs/ metrics (for hash collision - track both success and failure) - can test failure by reducing hash length @gweiying 
- [x] debug why only 7700~ links each time - solved, our `express-file-upload` middleware was truncating our file to the specified size silently instead of throwing an error
- [ ] fix frontend download - might not fix this before resolving issue with disordered links when changing number of links on page 
- [x] fix tests

Environment variables
- `BULK_UPLOAD_MAX_NUM` | Required? - No | Maximum number of links that can be bulk uploaded at once. Defaults to 1000|
- `BULK_UPLOAD_RANDOM_STR_LENGTH` | Required? - No | String length of randomly generated shortUrl in bulk upload. Defaults to 8|

Migration Notes
- `20220929_add_source_column.sql` should be **run pre-deploy** since the deployment contains code changes that will attempt to write into the new column
- `20220929_backfill_source_column.sql` should be **run post-deploy** since the backfill migration backfills and adds a non-null constraint on the new column -> our deployment should be running so that newly created links will have source set
  - lets run this outside of working hours as this might be a table locking operation
- Migrate Serverless functions on DB after deployment so that user link transfers can still happen (may need to increase Lambda timeout)

PR #2031, opened 2022-10-13
chore(datadog): add metrics for malicious file or link activity

halfwhole:
## Problem

Currently, the metrics for malicious activity are defined under metric filters for CloudWatch log groups. (See [AWS console](https://ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#logsV2:log-groups/log-group/$252Faws$252Felasticbeanstalk$252Fgo-production$252Fvar$252Flog$252Feb-docker$252Fcontainers$252Feb-current-app$252Fstdouterr.log$23metric-filters) for more details, requires AWS permissions to access)

The current approach presents several problems:
1. We cannot distinguish between malicious links and malicious files being uploaded
2. We cannot distinguish between staging/prod and gov/edu/health services without giving each metric a different name
3. The metrics are not defined in the codebase, so it's not under version control, and it adds some 'hidden' technical complexity on AWS infra that's not obvious to developers

## Solution

Add two new malicious activity metrics that will be sent to Datadog: one for malicious files (`go.malicious_activity.file`), and one for malicious links (`go.malicious_activity.link`). The env and service tags for staging/prod and gov/edu/health respectively should be added on Datadog automatically. This will allow us to remove metric filters on CloudWatch in the future as well, if we want to.

**Improvements**:

Resolved a TODO comment from before (https://github.com/opengovsg/GoGovSG/pull/1993#discussion_r984335841) in `UrlManagementService.ts`. For the shortlink creation metric, a new tag was also added to label the source (console or bulk, but not yet API).

## Tests

- [x] Test on staging to make sure that new metrics are being sent with tags

## Deploy notes

Post-deployment: can remove CloudWatch metric filters for malicious activity (optional, but would be nice to do)

### DIFF
--- a/src/server/modules/threat/FileCheckController.ts
+++ b/src/server/modules/threat/FileCheckController.ts
@@ -2,6 +2,7 @@ import { NextFunction, Request, Response } from 'express'
 import { inject, injectable } from 'inversify'
 
 import fileUpload from 'express-fileupload'
+import dogstatsd, { MALICIOUS_ACTIVITY_FILE } from '../../util/dogstatsd'
 import jsonMessage from '../../util/json'
 import { DependencyIds } from '../../constants'
 import { FileTypeFilterService, VirusScanService } from './interfaces'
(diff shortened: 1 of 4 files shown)


gweiying (approved):
lgtm!

PR #2032, opened 2022-10-14
Feat/qrcode infra lambda

jimvae:
- [x] add qr code generation
- [x] add upload to s3
- [x] add send message to sqs
- [ ] optimise qr code generation


src/server/serverless/bulk-qrcode-generation/qrCode.js
```diff
@@ -0,0 +1,204 @@
...
+  PNG: 'png',
+  JPEG: 'jpeg',
+}
+
+const envVariant = process.env.ASSET_VARIANT
+
+const domain = process.env.DOMAIN_URL
+
+const assetVariants = ['gov', 'edu', 'health']
+if (!assetVariants.includes(envVariant)) throw Error('Invalid variant name!')
```
gweiying, line 32: should also throw error if domain is missing

src/server/serverless/bulk-qrcode-generation/index.js
```diff
@@ -0,0 +1,42 @@
...
+      bufferData,
+      'application/zip',
+      `${filePath}/generated_svg.zip`,
+    )
+
+    // bufferData = await createQRCodesBuffer(
+    //   mappings.map((mapping) => mapping.shortUrl),
+    //   ImageFormat.PNG,
+    // )
+    // await uploadToS3(bufferData, 'application/zip', `${filePath}/generated_png.zip`)
```
gweiying, line 28: is there a reason this is commented out?

serverless.yml
```diff
@@ -46,5 +46,13 @@ functions:
           - 'src/server/serverless/capture-ses-events/**'
           - 'package-lock.json'
           - 'package.json'
-
-  
+  bulk-qrcode-generation:
+    handler: src/server/serverless/bulk-qrcode-generation/index.handler
+    memorySize: 512
```
gweiying, line 51: is this memory size sufficient for 1000 QR codes

src/server/serverless/bulk-qrcode-generation/index.js
```diff
@@ -0,0 +1,42 @@
+const { ImageFormat, createQRCodesBuffer } = require('./qrCode')
+const { createCsv } = require('./createCsv')
+const { uploadToS3 } = require('./s3')
+const { sendSQSMessage } = require('./sqs')
+
+async function handler(event) {
+  const { body } = event.Records[0]
```
gweiying, line 7: Nit, can add a short inline comment that this currently assumes Records batch size of 1?

gweiying (commented):
Overall looks good to me! 

Would be also good to add a small documentation writeup in the PR of any considerations you had to take into account for the Lambda function. For example, you were having issues with the size of the Lambda function and the memory usage, how did you resolve this? Would be good to leave these notes here so that if we need to rework this in the future we know what not to do.

Edit 
Just clarified with @jimvae offline, the current QR code writing process does not support up to 1000 QR codes and will need to be fine-tuned further. Key issue with current approach is that we are storing QR codes in buffer, taking up space in the Lambda RAM.

PR #2037, opened 2022-10-18
fix(tagging): fix downloading links with tags

halfwhole:
## Problem

To reproduce:
1. Search by tags on the user page
2. Click on "Download links"

Observed behaviour: downloaded links do not reflect the URLs searched by tags in the links table

Expected behaviour: downloaded links should reflect the URLs searched by tags in the links table

The bug happens because the function `getUrls` (used by the downloader) is unable to differentiate between searching by links vs searching by tags, and incorrectly always searches by links. This is unlike `getUrlsForUser` (used by the search bar), which has some extra added logic to differentiate between the two.

See [Notion issue](https://www.notion.so/opengov/Report-Empty-download-link-csv-84b79c4ea3b9402796204cdc538d701e)

## Solution

Instead of mimicking the extra logic from `getUrlsForUser` in `getUrls` to differentiate between searching by links vs. by tags, this PR refactors the redux store such that this logic is no longer necessary. For context, `searchText` in the redux store is currently being overloaded for use in both searching by links and searching by tags. This PR refactors it by splitting it out into separate fields `searchText` and `tags`, which is better because it is consistent with the backend API.

I've also renamed `searchTextInput` to `searchInput` to disambiguate it from `searchText`. Now, `searchInput` can either be dispatched to `searchText` or `tags`, depending on the `isTag` boolean variable.

## Tests

- We don't have E2E tests for downloading links as CSVs. Might want to set that up sometime


thanhdatle (approved):
It's a pretty elegant fix to me.

PR #2039, opened 2022-10-18
feat(tagging): allow backspaces to remove tags

halfwhole:
## Problem

Currently, clicking on the 'x' on a tag on the create/edit link forms is the only way to remove the tag. Backspace doesn't do this, which may be unintuitive or non-keyboard friendly.

See [Notion issue](https://www.notion.so/opengov/Update-Delete-should-delete-the-tag-ae73e761415e4ac1874965316a7e7444)

## Solution

Allow backspaces to remove tags.


gweiying (approved):
lgtm

PR #2040, opened 2022-10-18
Feat/bulk/frontend/status bar

gweiying:
### Background

This PR adds the frontend status bar to display the status of QR code generation. Designs in [Notion here](https://www.figma.com/file/Syvs1wxfOiRGJVXGGiHMiz/Go-2022-Design-Master?node-id=1181%3A29084)

Specifically, the PR adds:
- [x] StatusBar component to display status (with 3 variants, Success, Info and Error)
- [x] Actions and reducers to set message title, body and variant for StatusBar

Remaining work to be done in separate PRs:
- [x] Connect bulk creation upload to trigger Info StatusBar https://github.com/opengovsg/GoGovSG/pull/2041
- [ ] Add action to retrieve QR code status state and update StatusBar (with Success or Failure) 
- [x] Add Download button for Success StatusBar #2047

### Screenshot
web view
Screenshot 2022-10-18 at 14 39 52
Screenshot 2022-10-18 at 14 38 53
Screenshot 2022-10-18 at 14 37 49

mobile view
Screenshot 2022-10-18 at 14 48 41


src/client/user/components/UserLinkTable/StatusBar/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+    content: {
+      paddingTop: theme.spacing(2),
+      paddingBottom: theme.spacing(2),
+      [theme.breakpoints.up('md')]: {
+        paddingRight: (props: StyleProps) => props.appMargins,
+        paddingLeft: (props: StyleProps) => props.appMargins,
+      },
+      [theme.breakpoints.down('sm')]: {},
+    },
+    message: {},
```
halfwhole, line 43:
```suggestion
```

src/client/user/components/UserLinkTable/StatusBar/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const dispatchCloseStatusBar = () => {
+    dispatch(userActions.closeStatusBar())
+  }
+
+  return (
+    <>
+      {hasStatusBarAlert && (
+        <Alert
+          icon={icon}
+          className={`${colorClass} ${classes.root} ${classes.content}`}
```
halfwhole, line 102: nit: maybe we could merge `classes.root` and `classes.content`, since they only seem to be used together?

src/client/user/components/UserLinkTable/UrlTable/EnhancedTableBody/index.tsx
```diff
@@ -129,6 +129,7 @@ const useStyles = makeStyles((theme) => {
     },
     clicksCellTooltipContent: {
       display: 'inline-block',
+      float: 'right',
```
halfwhole, line 132: wondering why we're going with right-aligning the cells here? seems to add a lot of space between the last two columns

.eslintrc.json
```diff
@@ -51,5 +51,13 @@
...
-  }
+  },
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx"],
+      "rules": {
+          "no-undef": "off"
+      }
+    }
+  ]
```
halfwhole, line 62: am curious what overriding `no-undef` would do :o is it risky to disable the rule for all typescript files?

halfwhole (approved):
lgtm! have questions about eslint and clicks column alignment though

src/client/user/components/UserLinkTable/StatusBar/index.tsx
gweiying: thought to leave it there for ease of styling next time, what do you think?

src/client/user/components/UserLinkTable/UrlTable/EnhancedTableBody/index.tsx
gweiying: To be honest, it was for ease of alignment with the status bar close button, so that the justification on the right from the close button to the end can be matched with the justification on the click count to the end. Had some issues with alignment otherwise

Will check with Feli if this is acceptable!

.eslintrc.json
gweiying: Thanks for checking in about this. 

I ran into an issue with eslint when using `JSX` type from React. This is because `eslint` does not check globally defined variables (unlike Typescript) when checking for definitions, so it throws a no-defined error. I disabled the `eslint` check based on the [recommendations here](https://github.com/typescript-eslint/typescript-eslint/blob/main/docs/linting/TROUBLESHOOTING.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors), quoted below:

> We strongly recommend that you do not use the no-undef lint rule on TypeScript projects. The checks it provides are already provided by TypeScript without the need for configuration - TypeScript just does this significantly better.

Since TypeScript is likely to provide more comprehensive and accurate checks on no-def, I think it'd be safe to disable eslint no-def checks. I'll also add the link inline for future reference. Thanks for clarifying!

src/client/user/components/UserLinkTable/StatusBar/index.tsx
gweiying: good idea [b7559d9](https://github.com/opengovsg/GoGovSG/pull/2040/commits/b7559d91c8663c78ed7333f72f265aefd956d326)

.eslintrc.json
gweiying: [ee5aac2](https://github.com/opengovsg/GoGovSG/pull/2040/commits/ee5aac22d6e49ee2ff215a9efe309ae5724543ad)

src/client/user/components/UserLinkTable/StatusBar/index.tsx
halfwhole: sure, would be good too!

PR #2041, opened 2022-10-18
chore: display QR code inprogress after bulk upload

gweiying:
Adds action to trigger status bar In progress display after bulk creation has completed


halfwhole (approved):
lgtm!

PR #2042, opened 2022-10-18
feat: json parsing error

### LINKED ISSUE #1961: Handle 500 error with invalid JSON input

developerdenesh:
## Problem

Closes #1961

## Solution

Check the incoming error code, type and syntax error. If 400 is received, return 400 back

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -225,7 +225,6 @@ initDb()
       }
 
       if (
-        err instanceof SyntaxError &&
         err.statusCode === 400 &&
         'body' in err &&
         err.type === 'entity.parse.failed'
@@ -233,6 +232,7 @@ initDb()
         // This catches body-parser errors and returns a 400 error message
         console.error(err)
         res.badRequest(jsonMessage('Bad Request. JSON is malformed'))
+        return
       }
 
       res.status(500).render('500.error.ejs')
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -230,6 +230,7 @@ initDb()
         'body' in err &&
         err.type === 'entity.parse.failed'
       ) {
+        // This catches body-parser errors and returns a 400 error message
         console.error(err)
         res.badRequest(jsonMessage('Bad Request. JSON is malformed'))
       }
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -223,6 +223,17 @@ initDb()
         res.badRequest(jsonMessage(err.error.toString()))
         return
       }
+
+      if (
+        err instanceof SyntaxError &&
+        err.statusCode === 400 &&
+        'body' in err &&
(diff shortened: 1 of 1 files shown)


developerdenesh:
This takes over from https://github.com/opengovsg/GoGovSG/pull/2030

src/server/index.ts
```diff
@@ -223,6 +223,17 @@ initDb()
         res.badRequest(jsonMessage(err.error.toString()))
         return
       }
+
+      if (
+        err instanceof SyntaxError &&
+        err.statusCode === 400 &&
+        'body' in err &&
+        err.type === 'entity.parse.failed'
```
gweiying, line 231: Thanks for contributing to our code! 

This error message is specific to the `body-parser` library we use. I suggest that it'd be nice for future reference either to 1) leave an inline comment above to note that this catches body-parser errors and returns a 400 error message (similar like how we did it with Joi errors above), or 2) wrap an error handler around the bodyParser.json() parsing in apiMiddleware like so, so that it's clear what this piece of code is meant to be catching.

```
(req, res, next) => {
    bodyParser.json()(req, res, err => {
        if (err) {
            console.error(err);
            return res.sendStatus(400); // Bad request
        }

        next();
    });
```
Taken from https://stackoverflow.com/questions/53048642/node-js-handle-body-parser-invalid-json-error

developerdenesh: Hi @gweiying, have added a comment similar to the Joi errors to reflect that this catches `body-parser` json errors and returns error 400 accordingly

gweiying: Thanks for the quick fix.

Pulled your code locally to run - just a couple of small issues. 
1) missing return statement 
2) small issue with Typescript - since `status` and `type` does not exist on `SyntaxError`, the code cannot be compiled. A quick solution for this is to just drop the `err instanceof SyntaxError` check since the other errors should be specific enough to catch the body parser error, or possibly to [create our own error type](https://github.com/expressjs/body-parser/issues/122#issuecomment-290918495) - the latter might be more involved.

gweiying (approved):
lgtm, thanks for your contribution

Issue #2045, opened 2022-10-19
Inaccurate status for AWS lambda migrate url to user


clementcht:
**Describe the bug**
The lambda for GoGovSG production "migrate url to user" shows “execution result: succeeded” even when the link migration has failed. Can this be edited to show “execution result: failure” if the link migration fails? eg. when short url is not found, or if user is not found.

**Expected behavior**
For lambda to show “execution result: failure” if the link migration fails, eg. when short url is not found, or if user is not found.

**Screenshots**
<img width="1589" alt="Screenshot 2022-10-19 at 1 33 34 PM" src="https://user-images.githubusercontent.com/108923650/196607052-8ae9336b-3f79-4169-9dd5-86798c6c0564.png">

PR #2046, opened 2022-10-20
chore: return error result in migrate-user-links and migrate-url-to-user lambdas

jimvae:
## Problem

- Inaccurate status for AWS lambda migrate url to user
- Inaccurate status for AWS lambda migrate user links

Closes [#2045 ](https://github.com/opengovsg/GoGovSG/issues/2045)

## Before
image

## After
image


gweiying (approved):
lgtm

PR #2050, opened 2022-10-26
fix: restrict state to active and inactive in url search

halfwhole:
## Problem

The query conditions for `state` when searching for URLs in the user page are overly permissive, and should be restricted to either active or inactive states.

See [Snyk dashboard - Improper Type Validation](https://app.snyk.io/org/gogovsg/project/195ed33d-81b4-40d4-9918-45609ce999fb#issue-dbdfdd86-c468-4c17-aef0-d0c9b0bad893)

## Solution

Use Joi validator to allow the optional `state` field to be only one of `ACTIVE`, `INACTIVE` ~or `''`.~

~I'm perhaps less comfortable with allowing `state` to be `''`, but the frontend inserts `&state=` by default so disallowing it will break the frontend search~ edit: removed `''` from `state` following the comments below!

**Tests**:

- Added tests for an invalid and empty `state` field in `UserController.test.ts`
- (Improvement) Renamed test cases that return a bad request response code to begin with "reports bad request", in line with other test cases in the file that do this

### DIFF
--- a/src/client/user/actions/index.ts
+++ b/src/client/user/actions/index.ts
@@ -366,7 +366,7 @@ const getUrlsForUser =
       offset,
       orderBy,
       sortDirection,
-      state: urlState,
+      ...(urlState && { state: urlState }),
       isFile,
       searchText,
       tags,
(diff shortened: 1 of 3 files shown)


yong-jie:
>but the frontend inserts &state= by default

random passing thought, but would changing the default frontend behavior to insert `&state=ACTIVE` instead, help? then our backend api's accepted inputs can be cleaner

halfwhole:
it'd be good to preserve the functionality of searching for both active and inactive links in some way though (currently done through `&state=`), and changing the default behaviour to `&state=ACTIVE` may not be able to help with that :(

agree that the inputs aren't really clean though - maybe we can try dropping `&state=` altogether if the user is searching for both active and inactive links, then the backend won't need to accept `''` as an input?

yong-jie:
oh yes good idea!

thanhdatle (approved):

PR #2063, opened 2022-11-01
feat: allow html for announcement modal message

halfwhole:
## Problem

The announcement modal currently does not allow HTML inside its message, preventing us from using clickable hyperlinks.

## Solution

Allow HTML for the announcement modal message. Since we control the announcement message (through the `ANNOUNCEMENT_MESSAGE` env variable), this should be safe.
- Sanitize message before setting it in inner HTML
- Changed from `display: flex` to `display: block` to avoid new HTML elements from being added in new columns
- Fixed small typo by renaming `hmtlSanitizer` to `htmlSanitizer`

## Before & After Screenshots

**AFTER**:

`ANNOUNCEMENT_MESSAGE` here is set to `You can now tag your links and manage your dashboards effectively. Learn more about this on <a href="https://guide.go.gov.sg/guide-1/link-tagging">our guide</a>.\nFurthermore, our team is constantly looking for new problems to solve for our users. If you would like us to build a solution for you, please provide your thoughts <a href="https://go.gov.sg/go-pain-point-survey">here</a>!`

Screenshot 2022-11-01 at 6 59 01 PM

### DIFF
--- a/src/client/app/util/format.ts
+++ b/src/client/app/util/format.ts
@@ -66,7 +66,7 @@ export function formatBytes(bytes: number) {
   return `${parseFloat((bytes / k ** i).toFixed(1))}${sizes[i]}`
 }
 
-export function hmtlSanitizer(dirtyHtmlString: string) {
+export function htmlSanitizer(dirtyHtmlString: string) {
   return sanitizeHtml(dirtyHtmlString, {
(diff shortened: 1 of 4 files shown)


thanhdatle (approved):
LGTM

PR #2068, opened 2022-11-03
fix(tagging): fix filtering for no results when searching by tags

halfwhole:
## Problem

To reproduce:
- Login and go to the user page, ensure that you have some links
- Search for a non-existent tag
- Search bar disappears and page says that you have no short links

Regression likely introduced by #2037, by separating `searchText` out into `searchText` and `tags`

Closes [Notion issue](https://www.notion.so/opengov/Address-bug-when-searching-for-tag-that-does-not-exist-b6c77f13eac149d796881c93f45921c0)

## Solution

Change the condition for determining if the user has searched/filtered for URLs, by also checking for the existence of `tags` in `tableConfig`

## Before & After Screenshots

**BEFORE**:
When searching for a non-existent tag:
Screenshot 2022-11-03 at 2 54 39 PM

**AFTER**:
When searching for a non-existent tag:
Screenshot 2022-11-03 at 2 55 31 PM

## Tests

- [x] Add E2E test for searching for non-existent tag
- [x] Add E2E test for searching for non-existent link

Also shortened the wait time for searching for tags from 3s -> 2s, which still seems more than long enough. Wait time for searching for links remains at 1s

### DIFF
--- a/src/client/user/components/EmptyState/isFiltered.ts
+++ b/src/client/user/components/EmptyState/isFiltered.ts
@@ -5,7 +5,7 @@ export default function useIsFiltered() {
   const tableConfig = useSelector(
     (state: GoGovReduxState) => state.user.tableConfig,
   )
-  const { searchText } = tableConfig
+  const { searchText, tags } = tableConfig
   const filtered = Object.values(tableConfig.filter).some((value) => !!value)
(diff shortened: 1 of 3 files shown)


thanhdatle (approved):
LGTM

PR #2074, opened 2022-11-10
fix: add return after making server response

halfwhole:
## Problem

In two places in the codebase, the server does not return immediately after making a response, leading to `res.send()` being called _twice_, which causes an unhandled promise rejection. This is still okay for node <14 because it just gives a warning, but not from node 15 onwards (see [node 15 release docs](https://nodejs.org/en/blog/release/v15.0.0/#throw-on-unhandled-rejections-33021)) as it throws an error and terminates the entire node process, which is very scary.

Example error message:
Screenshot 2022-11-10 at 6 26 35 PM

## Solution

- In `UserController`: simply add `return`
- In `LinkAuditController`: had to invert the logic a little to do the same functionality without calling `res.send()` twice


gweiying (approved):
lgtm

PR #2077, opened 2022-11-14
feat(api): add feature flag for external API

halfwhole:
## Problem

We want to have a feature flag to toggle between enabling/disabling the external API

## Solution

Introduce a new optional env variable `FF_EXTERNAL_API`, a boolean that defaults to false unless explicitly set to `true`. This variable controls whether the external API v1 routes are enabled.

## Deploy Notes

- `FF_EXTERNAL_API` (optional): feature flag for external API that defaults to false

Pre-deploy or post-deploy: set `FF_EXTERNAL_API` to `true` to enable it when we're ready


halfwhole:
^ ignore the above gitguardian comment, it relates to `API_KEY_SALT` in `docker-compose.yml` that's safe and wasn't even touched by this PR

thanhdatle (approved):

PR #2080, opened 2022-11-14
fix: handle unhandled rejections

halfwhole:
## Problem

Unhandled rejections can arise due to situations like in #2074. In the event that such a situation occurs and we've upgraded to node 16, unhandled rejections will cause node to crash.

## Solution

Handle unhandled rejections in the node process, preventing the process from crashing.

Screenshot (after) for a sample logger error message, when there's an unhandled rejection:
Screenshot 2022-11-14 at 5 18 28 PM


src/server/index.ts
```diff
@@ -242,6 +242,10 @@ initDb()
 
     const port = 8080
     app.listen(port, () => logger.info(`Listening on port ${port}!`))
+
+    process.on('unhandledRejection', (error) => {
+      logger.error(`Unhandled rejection:\t${error}`)
```
gweiying, line 247: what do you think about sending this event to DD too as an alert of sorts? seems like we want to be more explicitly notified of unhandled rejections

halfwhole: sounds a good idea to have DD alerts for this! it looks to me that there's 2 options for that (that don't take too much extra effort):
1. no changes to the above code, create a DD monitor based on logs filtering with the term "Unhandled rejection"
2. add a new metric here for unhandled rejections, create a DD monitor based on the metric

(2) feels a bit more proper, so have a slight preference for that but am ok with whichever approach, what do you think?

thanhdatle: I would prefer (2) too

halfwhole: have updated, please take a look

src/server/index.ts
```diff
@@ -242,6 +243,11 @@ initDb()
 
     const port = 8080
     app.listen(port, () => logger.info(`Listening on port ${port}!`))
+
+    process.on('unhandledRejection', (error) => {
+      dogstatsd.increment(ERROR_UNHANLDED_REJECTION, 1, 1)
```
gweiying, line 248: nit, spelling error but otherwise lgtm!

gweiying (approved):

src/server/index.ts
halfwhole: whoops, fixed in [f878e46](https://github.com/opengovsg/GoGovSG/pull/2080/commits/f878e46d01435526750ef1aee7277de119eb8b51)

PR #2084, opened 2022-11-16
fix: disallow updating files for links and links for files

halfwhole:
## Problem

One can manually submit a request (`PATCH /api/user/url`) to update the `longUrl` for a file, or the `file` for a link. This should not be allowed.

Screenshot 2022-11-16 at 11 25 32 AM

## Solution

Disallow updating files for links and links for files at the service layer, depending on the `isFile` field of the URL.
- Return a 400 bad request on the user and API controller when encountering this error
- Created a new error type `InvalidUrlUpdateError` for this, because it didn't seem to fall under any existing error type - not sure of my approach though, happy to modify where necessary


thanhdatle (approved):
LGTM

PR #2086, opened 2022-11-18
chore: change default session cookie max age to 1 day

halfwhole:
## Problem

It's really tiring for devs to have to log in again every 30 mins

## Solution

Change the default `COOKIE_MAX_AGE` from 1800000 (30 mins) to 86400000 (1 day) in `docker-compose.yml`.
- `COOKIE_MAX_AGE` is already set to 86400000 on gov/edu/health prod, so there will be no change at all to production
- I could not find any history as to why `COOKIE_MAX_AGE` was set at 30 mins 3 years ago, but since prod already overrides this to 1 day anyway, I figure that bumping everything to 1 day by default is fine


gweiying (approved):
GREAT! hahaha I always changed it locally but this is much better

PR #2087, opened 2022-11-22
feat(api): disable closing api modal on click out

halfwhole:
## Problem

The user may accidentally click out of the API key modal, which closes it before they have confirmed saving the API key. See notion issue description [here](https://www.notion.so/opengov/Bug-Staging-key-starts-with-live-fba9622a73cd49e687363e383c9a637a)

## Solution

Disable closing the API key modal when the user clicks out from the modal.

### DIFF
--- a/src/client/apiintegration/components/ApiKeyModal/index.tsx
+++ b/src/client/apiintegration/components/ApiKeyModal/index.tsx
@@ -82,7 +82,6 @@ const ApiKeyModal: FunctionComponent = () => {
       fullScreen={false}
       fullWidth
       maxWidth="sm"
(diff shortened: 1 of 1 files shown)


gweiying:
Have we confirmed that this is the behavior we desire on the frontend?

halfwhole:
Nope, am still waiting on a reply @.@ lemme bug someone for an answer soon

halfwhole:
Closing for now as there are no plans to go ahead with this anytime in the near future, unless we identify a need to do so (e.g. user complaints). If so, we can always reopen this PR

PR #2089, opened 2022-11-22
fix: revise Download Links to only download currently visible URLs fr…

thanhdatle:


src/client/app/util/download.ts
```diff
@@ -17,67 +19,49 @@ export const downloadUrls = async (urlCount: number, tableConfig: any) => {
...
-      ...tableConfig,
-      limit,
-      offset: i * limit,
-    }
-    getUrlsPromiseArray.push(userActions.getUrls(queryObj))
-  }
-
-  await Promise.all(getUrlsPromiseArray).then((promises) => {
+  const queryObj = queryObjFromTableConfig(tableConfig)
+  await userActions.getUrls(queryObj).then((promise) => {
```
halfwhole, line 23: nit: would it be preferable to do the following? it looks possible to eliminate the `then` callback entirely by taking the result of `await`

```suggestion
  const { json, isOk } = await userActions.getUrls(queryObj)
```

halfwhole (approved):
LGTM, looks a lot cleaner too!

src/client/app/util/download.ts
thanhdatle: good idea, done!

PR #2090, opened 2022-11-22
fix: set apiEnv according to DD_ENV instead of NODE_ENV

thanhdatle:
## Problem

NODE_ENV is set to production in both Staging and Production environment, causing apiEnv to have `live_` prefix in both staging and prd

## Solution

To make apiEnv to rely DD_ENV instead.


halfwhole (approved):
lgtm - feels a little hackish to be relying on `DD_ENV`, but should be the simplest option for now

PR #2093, opened 2022-11-22
Feat/bulk/all get job client

gweiying:
Adds frontend logic to retrieve user's latest jobs and updates to their job status.

### Details
**Desired status bar behavior**
- Always show the last update, unless user closes it manually
- Status bars are updated automatically without user having to refresh them
- Status bars across different browsers and tabs should display the same state
- Can download QR codes if successful
- In progress status bars cannot be closed

### Implementation details
**Always show the last update, unless user closes it manually**
This is achieved by always fetching the user's last job. If a user has acknowledged it and closed it, the stringified representation of the job is stored in a user's localStorage. When the job is fetched and it is equal to the job stored in local storage, we do not display it, otherwise we display it.

Note that currently, we are also storing the jobItemUrls for failed jobs in the status message even though it is not displayed to the user. This is a way to ensure that the localStorage message is unique, i.e. if a user has closed a failure message before, and a new failure message appears, the localStorage message needs to be unique, otherwise the message will not re-appear. The alternative would have been to clear the localStorage every time a new job is created, but for now I've stored the jobItemUrls because I was thinking about eventually reusing the callbacks field to retrigger the failed jobs.

**Status bars are updated automatically without user having to refresh them**
This is achieved by exposing a long polling endpoint on the backend. The client opens a request to the backend to poll for job updates for a given jobId. When the jobId status updates from in progress to completed/ failed, the updated job and related information is sent to the client, and the client changes the status bar based on the response. If the request to the backend times out with a 408 error, the client reopens the connection.

**Status bars across different browsers and tabs should display the same state**
Status bar statuses are fetched from the backend, so status bars across different browsers and tabs should always behave the same, **with one exception**: We use localStorage to keep track of whether a user has previously closed a status bar. Since localStorage is not shared between browsers, if a user has acknowledged and closed a status bar in one browser, the status bar will still appear on another browser since we are unable to track if a user has previously acknowledged it. However, if a user starts a new job, the backend will fetch the new job instead, and the two browsers will synchronise in display again.

**Can download QR codes if successful**
The backend sends the bucket urls to download from. If the job is successful, the download button is displayed and allows the user to start downloads from the bucket.

**In progress status bars cannot be closed**
No close button for the in progress status bar


src/client/user/components/UserLinkTable/StatusBar/index.tsx
```diff
@@ -92,18 +88,45 @@ const StatusBar = () => {
...
+  useEffect(() => {
+    if (header && body) {
+      const localStatusBarMessage = localStorage.getItem('statusBarMessage')
+      const statusBarMessageString = JSON.stringify(statusBarMessage)
+      if (localStatusBarMessage === statusBarMessageString) {
+        // if current message is same as last acknowledged message, hide
+        setShowStatusBar(false)
+      } else {
+        // current message has not been acknowledged before
+        setShowStatusBar(true)
```
halfwhole, line 110: i'm wondering what would happen if we decide to add in S3 bucket lifecycle rules? say a user does the following:
- User bulk uploads links, everything is successful and zipped QR codes are uploaded to S3
- The expiry time passes (e.g. 1 month), zip files are deleted from S3
- User clears local storage and logs into the website, sees the success modal to download zip files
- User tries to download the now non-existent zip files, gets error

is there any way to prevent showing the latest job after a certain amount of time has passed (e.g. a week)?

src/client/user/actions/index.ts
```diff
@@ -934,6 +925,90 @@ const updateTags =
...
+    >,
+  ) => {
+    const response = await postJson(`/api/user/job/status`, { jobId })
+    const isOk = response.ok
+    if (!isOk) {
+      // retry
+      dispatch(getJobInformationForPendingJobId(jobId))
+    } else {
+      const resp = await response.json()
+      console.log(resp)
```
halfwhole, line 978:
```suggestion
```

src/client/user/actions/index.ts
```diff
@@ -934,6 +925,90 @@ const updateTags =
     })
   }
 
+const setQRCodeGenerationMessage =
+  (status: string, callbacks: string[]) =>
```
halfwhole, line 929: does the term `callbacks` have any specific meaning? it looks like it's only ever used in the context of `jobItemIds` in the frontend, would that name hopefully be more descriptive?

src/client/user/actions/index.ts
```diff
@@ -248,9 +248,14 @@ const closeStatusBar: () => CloseStatusBarAction = () => ({
...
+  callbacks: string[],
+) => SetStatusBarErrorMessageAction = (
+  header: string,
+  body: string,
+  callbacks: string[],
+) => ({
   type: UserAction.SET_STATUS_BAR_ERROR_MESSAGE,
-  payload: { header, body },
+  payload: { header, body, callbacks },
 })
```
halfwhole, line 259: will error messages need to have `callbacks` for the job item URLs? if i'm understanding correctly, it seems strange to allow users to access their S3 links if the latest job failed as a whole

src/client/user/components/UserLinkTable/StatusBar/index.tsx
gweiying: Sure, Hena and I discussed that we could do something like this in the future (i.e. cut off the display after a certain time period e.g. one week) but we're not implementing this for now to give users full control of their experience. We can implement the lifecycle rules after implementing the time period cutoff. 

The technical implementation should be quite straightforward because we can just modify the call to retrieve the latest job for a user to cut off after a certain period.

src/client/user/actions/index.ts
gweiying: For now I'm using it to just ensure that the localStorage message is unique, i.e. if a user has closed a failure message before, and a new failure message appears, the localStorage message needs to be unique, otherwise the message will not re-appear. I could also do it by clearing the localStorage every time a new job is created, but I thought that this was a neater solution because we can probably eventually reuse the callbacks field to retry the failed jobs.

src/client/user/actions/index.ts
gweiying: Was thinking about reusing it eventually to allow users to retrigger failed jobs, which is why I left the name pretty generic. What do you think?

src/client/user/components/UserLinkTable/StatusBar/index.tsx
halfwhole: i see, thanks for clarifying!

src/client/user/actions/index.ts
halfwhole: AH i didn't think of that at all, sounds neat!

src/client/user/actions/index.ts
halfwhole: ahh sounds good to leave it as this then since we'll revisit it eventually!

halfwhole (approved):

src/client/user/actions/index.ts
gweiying: Added it into the PR description!

PR #2100, opened 2022-11-28
feat: improve validations and errors for shortUrl and longUrl

halfwhole:
## Problem

The server errors for invalid short and long URLs are not always caught up-front at the Joi validator level, leaving them to be caught at the Sequelize level or via safe browsing checks (which is not that great). Plus, the error messages thrown are not very informative or well-formatted.

Closes [this Notion issue](https://www.notion.so/opengov/Improve-error-messages-for-external-APIs-b41fc3755f0d46958ae62536b0830e27)

## Solution

- Shift Joi validator checks for URLs *before* safe browsing checks (in `urlCheckController`) and various file checks (in `fileCheckController`)
- Add long URL Joi validations for invalid URLs and circular redirect checks
- Update error messages to be more informative
- Standardise error messages, including between Joi validations and Sequelize models - e.g. all end with period, capitalize "URL", "not allowed" vs. "prohibited", etc.

There's a bit of error duplication across the Joi schemas and Seqeulize validations, but I can't think of an easy fix - it seems ok for now, so will leave it as it is

## Tests

- (Eventually) Add integration tests for invalid short and long URL formats


thanhdatle (approved):
LGTM, thanks for working on this!

gweiying (approved):
lgtm!

PR #2101, opened 2022-11-28
fix: update safe browsing error messages

halfwhole:
## Problem

Safe browsing errors are not being logged and returned properly. 

The returned error message looks like this, because `response` and `response.body` are objects:
Screenshot 2022-11-28 at 4 52 10 PM

The logged error message looks like this, because it's trying to log `error` instead of `error.message`:
Screenshot 2022-11-28 at 3 25 04 PM

## Solution

- Change the error message contents from `response` and `response.body` to `body.error.message`
- Change the logged error message from `error` to `error.message`

Now, both the logged and returned error messages look like this:
Screenshot 2022-11-28 at 5 18 28 PM

For context, when calling the Safe Browsing API, the reponse JSON `body` looks something like this when there's an error:
Screenshot 2022-11-28 at 3 23 54 PM

The indentation changes from line 83/84 onwards in this PR are caused by prettier

Also, apologies to the team for unintentionally exposing the now-removed API key 🙇

### DIFF
--- a/src/server/modules/threat/UrlCheckController.ts
+++ b/src/server/modules/threat/UrlCheckController.ts
@@ -45,7 +45,7 @@ export class UrlCheckController {
           return
         }
       } catch (error) {
-        logger.error(error)
+        logger.error(error.message)
         res.serverError(jsonMessage(error.message))
         return
       }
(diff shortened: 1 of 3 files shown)


gweiying (approved):
lgtm! thanks for the catch

PR #2102, opened 2022-11-28
Fix/bulk/all errors

gweiying:
Fixes the following errors in bulk upload:

- [x] Bulk validation errors - root error was because we did not promisify the parsing so errors were timing out
  - [x] File with num links > `BULK_UPLOAD_MAX`
  - [x] Incorrect header
  - [x] File with 0 links only header (previously uncaught, added additional check)
- [x] Error styling and display for bulk upload
- [x] Dogstatsd variables
- [ ] Please let me know if I've missed any other fixes

Pending changes (will be in separate PR after rebase):
- [ ] Lambda guard refactor to use API
  - [ ] remove Lambda env variables
  - [ ] change to `/callback/qr`

### DIFF
--- a/src/server/modules/bulk/services/BulkService.ts
+++ b/src/server/modules/bulk/services/BulkService.ts
@@ -18,91 +18,91 @@ const BULK_UPLOAD_MAX_NUM = bulkUploadMaxNum
 
 @injectable()
 export class BulkService implements interfaces.BulkService {
-  parseCsv: (file: UploadedFile) => interfaces.CSVSchema = (file) => {
+  parseCsv: (file: UploadedFile) => Promise<string[]> = async (file) => {
     const dataString = file.data?.toString()
 
-    const schema = {
-      rows: 0,
-      isValid: true,
-      longUrls: [],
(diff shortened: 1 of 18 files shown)


src/server/modules/bulk/services/BulkService.ts
```diff
@@ -18,91 +18,91 @@ const BULK_UPLOAD_MAX_NUM = bulkUploadMaxNum
...
-              case isNotBlacklisted:
-                schema.errorMessage = `Row ${updatedRow}: ${stringData} is blacklisted`
-                break
-              case isNotEmpty:
-                schema.errorMessage = `Row ${updatedRow} is empty`
-                break
-              case isNotCircularRedirect:
-                schema.errorMessage = `Row ${updatedRow}: ${stringData} redirects back to ${ogHostname}`
-                break
+            switch (true) {
```
halfwhole, line 68: TIL about `switch (true)`, never seen it before but honestly am loving it

halfwhole (approved):
LGTM

PR #2104, opened 2022-11-29
Feat/bulk/all api guard

gweiying:
Refactor existing lambda guard to use API admin user token

Before merging:
- [x] Add `ADMIN_API_EMAIL` to EB and generate API key for email
- [x] Modify `EB_CALLBACK_SECRET` on Lambda to new API key


halfwhole (approved):
lgtm!

PR #2105, opened 2022-11-30
chore(datadog): add metric for generating API keys

halfwhole:
## Problem

We want metrics to track API key generation

## Solution

Add datadog metrics to track API key generation, and whether these API keys are new or not. This will incur an extra database read, but the cost is probably quite small in the grand scheme of things


thanhdatle (approved):
I like your use of words

PR #2106, opened 2022-12-01
chore: refactor validations for getting urls

halfwhole:
## Problem

We want to clean up validations for getting URLs on the internal API, as noted in #2073 

## Solution

Improved the validations for getting URLs on the internal API (user controller).

**Improvements**:

Validations
- Joi validations are clearer in terms of what they expect from each of the params
- Removed `updatedAt` as an option for `orderBy`, since it's unused
- Added optional XOR between `searchText` and `tags` to ensure that both are not used at the same time
- Add `lowercase()` to `searchText`, since we want to convert it to lowercase anyway

User controller
- Set default `orderBy` option to `createdAt` rather than the unused `updatedAt`
- Make `searchText` and `tags` optional params to pass down to the URL management service and URL repository

### DIFF
--- a/src/server/api/user/validators.ts
+++ b/src/server/api/user/validators.ts
@@ -43,15 +43,16 @@ const tagSchema = Joi.array()
 
 export const userUrlsQueryConditions = Joi.object({
   userId: Joi.number().required(),
-  limit: Joi.number().required(),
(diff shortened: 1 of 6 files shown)


gweiying (approved):
lgtm!

PR #2109, opened 2022-12-02
chore: add gitguardian precommit checks

halfwhole:
## Problem

Devs like me may accidentally commit and push secrets to GitHub

## Solution

Add GitGuardian pre-commit check to detect secrets and stop commits with secrets from being made. Devs are still able to manually override these checks if it's a false positive.

To enable secrets detection, the user must:
- Install `ggshield` locally (no npm package is available)
- Set `GITGUARDIAN_API_KEY` in `.env`

If either of those are not set up, the secrets detection check will be skipped.

## Screenshots

Example of what happens when a secret is exposed (this test credential has already been revoked):

Screenshot 2022-12-02 at 10 47 51 AM


halfwhole:
Would any of these approaches be a good idea?
1. Make it compulsory for all devs, i.e. pre-commit check fails if gitguardian is not set up locally
2. Make it optional, but change the echo'ed message to a warning that strongly recommends setting up gitguardian if not already done so
3. Make it optional, just discuss internally within our team to ensure that we set this up

PR #2114, opened 2022-12-05
chore: disable otp rate limit on local dev

halfwhole:
## Problem

Integration tests on local dev (e.g. #2091) can make many user logins within a short time period, and if the OTP rate limit is too low, the logins will fail

## Solution

Disable OTP rate limiting on local dev. Setting the max rate limit to 0 disables it entirely (see [`express-rate-limit` documentation](https://www.npmjs.com/package/express-rate-limit#configuration))

### DIFF
--- a/src/server/config.ts
+++ b/src/server/config.ts
@@ -121,7 +121,7 @@ if (DEV_ENV) {
     maxAge,
   }
   proxy = false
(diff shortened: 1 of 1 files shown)


gweiying (approved):
lgtm!

PR #2154, opened 2023-02-16
fix(LoginForm): validate email on initial blur, then on change

mantariksh:
## Problem
On the login page, the error message for the email input appears the moment you start typing. This is quite jarring:

https://user-images.githubusercontent.com/29480346/219290831-ca3deb77-2cd8-4146-aa03-cbbf23f66085.mov

## Solution

Use an `onTouched` strategy, as offered by [`react-hook-form`](https://react-hook-form.com/api/useform); in other words, validate on blur initially, then on change subsequently.

https://user-images.githubusercontent.com/29480346/219291002-0ef82e74-1154-4e88-9dd2-ff6fb17561dd.mov

### DIFF
--- a/src/client/login/components/LoginForm.tsx
+++ b/src/client/login/components/LoginForm.tsx
@@ -10,8 +10,9 @@ type LoginFormProps = {
   variant: VariantType
   autoComplete: string
   onChange: (email: string) => void
-  textError: () => boolean
-  textErrorMessage: () => string
+  onBlur?: () => void
+  isError: boolean
+  textErrorMessage: string
   onSubmit: (e: React.FormEvent<HTMLFormElement>) => void
   value: string
 }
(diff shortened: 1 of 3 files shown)


halfwhole (commented):
thanks for helping with the fix, and sorry we took so long to get back to this!!

lgtm for the `onTouched` logic, though when submitting the login form by hitting enter with an invalid email initially (before validation on initial blur), the error message below the email field doesn't appear, and instead an "Unauthorized" error toast shows up above -- is there a way we could also trigger the validations and prevent invalid submissions upon form submission, either by adding logic to `onSubmit` or by some other method?

https://user-images.githubusercontent.com/41856541/224600987-7f41076b-0396-406a-ae5c-ff243f33a314.mov

PR #2176, opened 2023-03-14
[Test] More E2E tests

oversparkling:
## Problem

Closes [insert issue #]

## Solution

**Features**:

- Details ...

**Improvements**:

- Details ...

**Bug Fixes**:

- Details ...

## Before & After Screenshots

**BEFORE**:
[insert screenshot here]

**AFTER**:
[insert screenshot here]

## Tests

## Deploy Notes

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

**New dev dependencies**:


gweiying:
Noting that the above tests addresses the following test cases:

Redirection
- [ ] It should redirect to the long url when short url is active
- [ ] It should redirect to a 404 page when short url is inactive
Drawer, Editing and Toaster logic
- [ ] It should redirect to the amended file immediately when the file is replaced (any caching for that short url is cleared)

PR #2179, opened 2023-03-14
feat(link-transfer): add request ownership button to directory table

halfwhole:
## Problem

Part of the link transfer feature (see [Notion doc](https://www.notion.so/opengov/20230309-Link-Transfers-4450fc62cba84a91862b130055d5abeb?pvs=4))

## Solution

This PR adds a "Request ownership" button to the directory table. On mobile, the button will not be shown on the table but only appears in the bottom panel (after the user clicks on the table row).

Clicking on "Request ownership" will take the user to Go's [Contact Us form](https://form.gov.sg/62f0747c7da3e1001240f8bb). In the future, clicking it should instead create an actual link transfer request and send an email out to the link owner, but as a first step we've agreed with product-side to redirect them to the contact form first.

**Notes**:

Ideally, the "Request ownership" button should be an actual `Button` component instead of a `Link`, but the CSS on the directory table is such a mess that it's a bit difficult to figure out a proper styling -- it should still be alright to have it as a link though, as that would be consistent with the other clickable links on the table too (e.g. contact email).

On the mobile panel though, the "Request ownership" button is actually a `Button`.

**Improvements**:

- Removed some unused CSS
- Removed some unnecessary padding/whitespace (to also make room for the "Request ownership" link)

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-03-14 at 5 52 28 PM
Screenshot 2023-03-14 at 5 52 58 PM

**AFTER**:
Screenshot 2023-03-14 at 5 53 22 PM
Screenshot 2023-03-14 at 5 53 46 PM

https://user-images.githubusercontent.com/41856541/224963954-5dade8bc-a7b4-4df8-997b-b29ade2c057a.mov

### DIFF
--- a/src/client/directory/components/DirectoryResults/DirectoryTable/DirectoryTableRow/index.tsx
+++ b/src/client/directory/components/DirectoryResults/DirectoryTable/DirectoryTableRow/index.tsx
@@ -3,6 +3,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import copy from 'copy-to-clipboard'
 import {
   Hidden,
+  Link,
   TableCell,
   TableRow,
   Typography,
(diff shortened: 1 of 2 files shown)


gweiying (approved):
lgtm!

PR #2181, opened 2023-03-15
chore: redirect user to existing link in directory

disKeith:
## Problem

User Story 1 of [Link Transfer Feature](https://github.com/opengovsg/GoGovSG/pull/2179#:~:text=transfer%20feature%20(see-,Notion%20doc,-))

## Solution

This PR changes the following:

1. Changed `CreateLinkForm` component's error message for short URLs to redirect to `/#/directory?query=${shortUrl}` to query the short URL that already exists
2. Updated error message if short link already exists

**Note:**
With the change, the address bar will not reflect the query as the `useEffect` hook in `SearchPage` component for Directory (which handles the keyword/email search) would set the query to empty as the logic is run once upon initialisation - to be worked on a separate refactoring effort.

**Improvements**:

- When user attempts to create a short link that already exists, and clicks on the error message, they will be redirected to the directory page with the existing short link queried to show the owner of the link

## Before & After Screenshots

**BEFORE**:
before_link_redir

**AFTER**:
after_link_redir

### DIFF
--- a/src/server/modules/user/UserController.ts
+++ b/src/server/modules/user/UserController.ts
@@ -98,7 +98,7 @@ export class UserController {
         return
       }
       if (error instanceof AlreadyExistsError) {
-        const newErrorMessage = `${error.message} Click here to find out more`
+        const newErrorMessage = `${error.message} Click here to find out who is the link owner & request for ownership.`
         res.badRequest(jsonMessage(newErrorMessage, MessageType.ShortUrlError))
         return
       }
--- a/src/client/directory/index.tsx
+++ b/src/client/directory/index.tsx
@@ -48,39 +48,6 @@ const useStyles = makeStyles(() =>
 type SearchPageProps = {}
 
 /**
- * Search query default parameters.
- */
-const defaultParams: GoSearchParams = {
-  query: '',
-  sortOrder: defaultSortOption,
-  rowsPerPage: 10,
-  currentPage: 0,
-  state: '',
-  isFile: '',
-  isEmail: 'false',
-}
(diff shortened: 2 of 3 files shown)


halfwhole (commented):
the code changes look very nice! when testing out the directory page though, typing on the search box gives kinda glitchy results:

Before:
https://user-images.githubusercontent.com/41856541/225202439-8392dfeb-4c1d-4543-8a73-891d8544d8b9.mov

After:
https://user-images.githubusercontent.com/41856541/225213804-9b9eff54-2e2b-4326-9170-82c3689c7cc6.mov

have a guess that it's due to the redirection/debouncing being shifted into the search component? :(

should we also change the error message shown to the user on the frontend? following the screenshot from the figma design:

Screenshot 2023-03-15 at 11 52 13 AM

src/server/modules/user/UserController.ts
```diff
@@ -98,7 +98,7 @@ export class UserController {
         return
       }
       if (error instanceof AlreadyExistsError) {
-        const newErrorMessage = `${error.message} Click here to find out more`
+        const newErrorMessage = `${error.message} Click here to find out who is the link owner & request for ownership.`
```
halfwhole, line 101: btw sometime ago dat and I were looking through the codebase and saw this line and agreed that the backend shouldn't be sending back this kind of error message lol (why is the backend telling the client to "Click here"?? that should be the frontend's responsibility)
but this change alone is good!

halfwhole (approved):
lgtm!

PR #2185, opened 2023-03-16
chore: fix error messages

gweiying:
The error messages on the Lambda were not sending successfully to the server as the error object was in the wrong shape (object instead of string).

This PR is a quick fix to change the shape of the error object.


halfwhole (approved):
lgtm!

PR #2195, opened 2023-04-10
feat: add env var to configure announcement modal button text

halfwhole:
## Problem

The button text on the announcement modal is fixed to "Try me now", but we want it to be configurable

## Solution

Add a new env var `ANNOUNCEMENT_BUTTON_TEXT` that allows configuration of the announcement modal's button text. If not set, the button's text will default back to "Try me now".

Also changed `width` to `minWidth` on the button styling, since the button text could overshoot the width limit and be wrapped across multiple lines, which we probably don't want?

## Before & After Screenshots

**BEFORE**:
Screenshot 2023-04-10 at 1 46 31 PM

**AFTER**:
Screenshot 2023-04-10 at 1 45 29 PM

## Tests

- [ ] Test it on staging, after adding `ANNOUNCEMENT_BUTTON_TEXT`

## Deploy Notes

- `ANNOUNCEMENT_BUTTON_TEXT` : button text on the announcement modal


thanhdatle (approved):
LGTM!

PR #2203, opened 2023-04-19
feat: allow zip files and block password-protected files

halfwhole:
## Problem

We want to allow users to upload zip files, but block users from uploading password-protected files (zip, pdf or otherwise) as Cloudmersive cannot properly scan them for viruses.

## Solution

1. Allow zip files to be uploaded
2. Switch from using basic file scanning to advanced file scanning on Cloudmersive. This gives us additional options to tell if the file is password protected or not, and block it from being uploaded if it is.

## Tests

On staging, we should test the following:
- [x] Regular non-password-protected files are allowed to be uploaded, including zip files
- [x] Password-protected files (zip, pdf) are blocked from being uploaded
- [x] 20mb files can still be scanned and uploaded within a reasonable amount of time

Should also update the tests on the release checklist to include password-protected zip/pdf files, if applicable?

## Deploy Notes

Unfortunately, this change doesn't do anything to address the already existing password-protected pdf files on Go, which have not been properly scanned for viruses :(

### DIFF
--- a/src/server/modules/threat/FileCheckController.ts
+++ b/src/server/modules/threat/FileCheckController.ts
@@ -67,11 +67,22 @@ export class FileCheckController {
     next: NextFunction,
   ) => Promise<void> = async (req, res, next) => {
     const file = req.files?.file as fileUpload.UploadedFile | undefined
+    const user = req.session?.user
     if (file) {
       try {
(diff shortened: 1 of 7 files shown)


thanhdatle (approved):
Works well in stg!

PR #2247, opened 2023-07-27
Experiment go <> sgid integration 

PikkaPikkachu:
## Problem

We are trying to integrate SGID with GO and will be testing it with OGP officers. If the integration is successful, we will open this up to other public officers as well. 

## Solution

Integration doc with the detailed solution [here](https://www.notion.so/opengov/Go-SGID-Integration-Working-Doc-9a5428038168411ba877a25827d867df), for summary refer below:

Followed developer docs [here](https://docs.id.gov.sg/introduction/overview) 

1. Add a new login page for SGID login `/ogp-login`
2. Upon user click we redirect to SGID authentication url (opens up in the same window)
3. Once user allows sign in using GO SGID client, we fetch their relevant data i.e. officer's work email.
4. There's a validation check on valid officer's email ending in `.gov.sg`
5. If the validation succeeds we create or find the user in the DB and set the user session
6. Else we redirect the back to sgid login page and show the relevant error message.

**Features**:
Login feature with SGID

## Before & After Screenshots

**AFTER**:

https://github.com/opengovsg/GoGovSG/assets/25703976/ad4cf77c-f29e-4242-b601-ed58fbcaf555

Tested on health and edu domains, we get a 400 response, with this error message on the UI: 

Screenshot 2023-08-03 at 5 59 32 PM

## Tests

1. Check login using SGID manually
2. Check remaining functionality remains intact [regression testing]
3. Check invalid authentication or invalid email does not succeed login
4. Check that edu and health domains are still intact, without the SGID client being initialised

## Deploy Notes
Need to create the relevant environment variables before deploying. 
There might be a need to use the AWS key manager, in case env variable store in EB does not work. 

- SGID Client 
- SGID Private key 
- SGID API hostname
- SGID Client Secret


src/client/sgidLogin/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+        opacity: 0.5,
+      },
+    },
+  }),
+)
+
+const SgidLoginPage = (): JSX.Element => {
+  const dispatch = useDispatch()
+  const classes = useStyles()
+  const queryParams = new URLSearchParams(
```
raynerljm, line 113: may i know why you decided to pass it via the hash instead of using query params? personally it took me a while to understand how this extracted the query params

on that note - it might help to defined 
`const URL_PREFIX_LENGTH = "#/ogp-login".length `
as it might be clearer as 11 is a magic number

src/client/sgidLogin/SgidLoginForm.tsx
```diff
@@ -0,0 +1,97 @@
...
+        >
+          {buttonMessage}
+        </Button>
+        {children}
+      </section>
+    </form>
+  )
+}
+
+export default SgidLoginForm
```
raynerljm, line 97: i don't seem to see whether this component is being imported - could you point me in the right direction if it is?

src/client/sgidLogin/actions/index.ts
```diff
@@ -0,0 +1,15 @@
+import { get } from '../../app/util/requests'
+import { GetReduxState } from '../../app/actions/types'
+
+const getAuthRedirectionUrl = () => (_getState: GetReduxState) => {
```
raynerljm, line 4:
```suggestion
const getAuthUrl = () => (_getState: GetReduxState) => {
```
having `redirection` in the function name might be misleading as the `redirect URL` or `callback URL` has a different meaning in OAuth 2.0.

this is an authorization URL as it redirects to the `/oauth/authorize` sgID QR code page.

src/server/services/sgid.ts
```diff
@@ -0,0 +1,98 @@
...
+  clientSecret: string
+  privateKey: string
+  redirectUri: string
+  hostname: string
+}
+
+export const SGID_LOGIN_OAUTH_STATE = 'login'
+
+class SgidService {
+  // SGID client will be null for other domains like edu, health
```
raynerljm, line 14: could you explain what you mean by the sgID Client will be null for other domains?

it is true that there will not be work email present in the `userinfo` requested for end-users logging in without a `.gov.sg` email but how does that make the `sgidClient` null? thanks!

src/server/services/sgid.ts
```diff
@@ -0,0 +1,98 @@
...
+      throw new Error('Error retrieving url via sgid-client')
+    }
+  }
+
+  /**
+   * Fetches the token via sgid SDK.
+   */
+  async callback(code: string, nonce: string, codeVerifier: string) {
+    if (!code || !this.sgidClient)
+      throw Error(`code cannot be empty or sgid client not initialised`)
```
raynerljm, line 64:
```suggestion
      throw new Error(`code cannot be empty or sgid client not initialised`)
```

may I know why you decided to use it without the `new` keyword? functionally they [seem to be the same](https://www.geeksforgeeks.org/difference-between-throw-errormsg-and-throw-new-errormsg/) but I think it would be good to standardize

src/server/services/sgid.ts
```diff
@@ -0,0 +1,98 @@
...
+      throw new Error('Error retrieving access token via sgid-client')
+    }
+  }
+
+  /**
+   * Fetches the user info via sgid SDK.
+   */
+  async userinfo(accessToken: string, sub: string) {
+    if (!accessToken || !this.sgidClient)
+      throw Error(`accessToken cannot be empty`)
```
raynerljm, line 82:
```suggestion
      throw new Error(`accessToken cannot be empty`)
```

same for this

src/server/services/sgid.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+// Initialised the sgidService object with the different environments
+export const SgidAuthService = new SgidService({
+  clientId: process.env.CLIENT_ID as string,
+  clientSecret: process.env.CLIENT_SECRET as string,
+  privateKey: process.env.PRIVATE_KEY as string,
+  redirectUri: 'http://localhost:3000/api/sgidLogin/authenticate',
+  hostname: process.env.SGID_API_HOSTNAME as string,
```
raynerljm: may i know why you used `as string`? the type coercion might result in runtime errors that TypeScript cannot catch

src/server/services/sgid.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+}
+
+// Initialised the sgidService object with the different environments
+export const SgidAuthService = new SgidService({
+  clientId: process.env.CLIENT_ID as string,
+  clientSecret: process.env.CLIENT_SECRET as string,
+  privateKey: process.env.PRIVATE_KEY as string,
+  redirectUri: 'http://localhost:3000/api/sgidLogin/authenticate',
```
raynerljm: the redirect URI should be defined via an environment variable as it will differ for your local/staging/dev environment

raynerljm (commented):
the sgID implementation looks like it should work! left some clarifications and some questions - thanks Prakriti! 🎉

src/server/services/sgid.ts
gweiying: Go is deployed to 3 environments, go.gov.sg, for.edu.sg (education link shortener) and for.sg (health link shortener). I believe Prakriti has it set up such that the SGID client is not initialized for the education and health environments!

.platform/hooks/postdeploy/01_fetch_ssm_parameters.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "Set AWS region"
+aws configure set default.region ap-southeast-1
+
+for ENV_VAR in "${ENV_VARS[@]}"; do
+  echo "Running for this ${ENV_TYPE}"
+  echo "Fetching ${ENV_VAR} from SSM"
+  VALUE=$(aws ssm get-parameter --name "${ENV_TYPE}_${ENV_VAR}" --with-decryption --query "Parameter.Value" --output text)
+  echo "${ENV_VAR}=${VALUE}" >> /opt/elasticbeanstalk/deployment/env
+  echo "Saved ${ENV_VAR}"
+done
```
PrawiraGenestonlia, line 15: Good stuff! Kudos to @PikkaPikkachu

src/client/sgidLogin/actions/index.ts
PrawiraGenestonlia: rayner brought up a good nit point.

src/client/sgidLogin/actions/types.ts
```diff
@@ -0,0 +1,9 @@
+import { ReduxPayloadAction } from '../../app/actions/types'
+
+export type SgidAuthRedireectionUrl = (url: string) => boolean
```
PrawiraGenestonlia, line 3: nit: typo

src/server/modules/auth/SgidLoginController.ts
```diff
@@ -0,0 +1,89 @@
...
+        !officerEmail.length ||
+        !officerEmail.match(
+          new RegExp(validEmailDomainGlobExpression.substring(1)),
+        )
+      ) {
+        // redirect back to sgid login page, if authentication fails,
+        // or officer email is not valid
+        res.redirect(`/#/ogp-login?statusCode=403&officerEmail=${officerEmail}`)
+        return
+      }
```
PrawiraGenestonlia, line 72: Great implementation of oauth endpoint. As for error handling when !officerEmail, I'll let go team comment on the implementation.

PrawiraGenestonlia (commented):
Left minor nit comments. Overall, good job @PikkaPikkachu with the implementation! Awesome work!

src/client/sgidLogin/SgidLoginForm.tsx
PikkaPikkachu: Oops copy pasta stuff — will remove!

src/client/sgidLogin/actions/index.ts
PikkaPikkachu: Ahh makes sense, thanks!

src/server/services/sgid.ts
PikkaPikkachu: Thanks for the catch!

src/server/services/sgid.ts
PikkaPikkachu: Ahh yup this was a test commit 😅  — we have resolved it now :)

src/server/services/sgid.ts
PikkaPikkachu: Yup resolved, thanks!

src/client/sgidLogin/index.tsx
PikkaPikkachu: For some reason the query params were not being extracted properly, hence had to go with this approach.

src/server/modules/auth/SgidLoginController.ts
```diff
@@ -0,0 +1,89 @@
...
+    return
+  }
+
+  public handleLogin = async (req: Express.Request, res: Express.Response) => {
+    try {
+      const { code, state } = req.query
+      const sessionData = req.cookies[SGID_STATE_COOKIE_NAME]
+
+      if (state !== SGID_LOGIN_OAUTH_STATE) {
+        throw new Error('Error logging in with SGID: state is incorrect')
```
halfwhole, line 49: what might you think about handling this similarly to line 70, i.e. something like `res.redirect('/#/ogp-login?statusCode=400'); return`, which at least sends the user back to the frontend where they can get some indication that something went wrong?

this failure mode with sgid state also seems to indicate more of a client fault with a bad request; meanwhile, throwing an `Error` probably gives an internal server error (?) indicating a server fault, which might not be our intention

.ebextensions/25-load-sgid-env.config
halfwhole, line 1: 🤯 https://github.com/opengovsg/FormSG/pull/2194

halfwhole (approved):
lgtm, thanks so much for this 🥳
the overall flow looks great, left a bunch of minor comments below but they're very much non-blocking so feel free to merge!

src/server/modules/auth/SgidLoginController.ts
PikkaPikkachu: Hmm makes sense! We can do that!

gweiying:
One additional comment from me, let's check the behavior on staging.for.edu.sg and staging.for.sg for the login request endpoint!

src/server/services/sgid.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+// Initialised the sgidService object with the different environments
+export const SgidAuthService = new SgidService({
+  clientId: process.env.SGID_CLIENT_ID || '',
+  clientSecret: process.env.SGID_CLIENT_SECRET || '',
+  privateKey: process.env.SGID_PRIVATE_KEY || '',
+  redirectUri: `${process.env.OG_URL}/api/sgidLogin/authenticate`,
+  hostname: process.env.SGID_API_HOSTNAME || '',
```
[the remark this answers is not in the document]
PikkaPikkachu: Ohh the hostname is required just, in case we would like the staging hostname from SGID, which is `staging.api.gov.sg` ? But I think makes sense to probably just have it as a hardcoded constant.

src/client/sgidLogin/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+      setLoginStatusCode(
+        `${officerEmail} doesn't look like a valid ${i18next.t(
+          'general.emailDomain',
+        )} email.`,
+      )
+    } else if (statusCode) {
+      setLoginStatusCode(
+        `Unable to fetch a valid work email for authentication.`,
+      )
+    }
```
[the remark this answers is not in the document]
PikkaPikkachu: Okiee, yeah we can think more about it, when we do a public launch, for now it's just for OGP officers :)

src/server/modules/auth/SgidLoginController.ts
```diff
@@ -0,0 +1,89 @@
...
+
+      const { data } = await this.sgidService.userinfo(accessToken, sub)
+      const officerEmail = data[OFFICER_EMAIL_SCOPE]
+
+      if (
+        !officerEmail ||
+        !officerEmail.length ||
+        !officerEmail.match(
+          new RegExp(validEmailDomainGlobExpression.substring(1)),
+        )
```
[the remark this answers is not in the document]
PikkaPikkachu: Thanks!

src/client/sgidLogin/actions/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { GetReduxState } from '../../app/actions/types'
+
+const getAuthUrl = () => (_getState: GetReduxState) => {
+  get('/api/sgidLogin/authurl').then(async (response) => {
+    if (response.ok) {
+      const text = await response.text()
+      window.open(text, '_self')
+    }
+  })
+}
```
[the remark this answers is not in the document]
PikkaPikkachu: Okiee changed it!

src/client/sgidLogin/index.tsx
PikkaPikkachu: hmm still didn't work :/

halfwhole (approved):
lgtm, apart from the first comment below!

src/server/modules/auth/SgidLoginController.ts
```diff
@@ -46,7 +51,8 @@ export class SgidLoginController {
       const sessionData = req.cookies[SGID_STATE_COOKIE_NAME]
 
       if (state !== SGID_LOGIN_OAUTH_STATE) {
-        throw new Error('Error logging in with SGID: state is incorrect')
+        res.redirect(`/#/ogp-login?statusCode=400`)
```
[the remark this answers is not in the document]
PikkaPikkachu: Hmm makes sense, added in!

PR #2290, opened 2024-02-06
chore: fix serverless plugin include dependencies to v5

halfwhole:
## Problem

Serverless deployment is failing recently, because we're using the most recent version of the `serverless-plugin-include-dependencies` plugin (v6.0.0, released 2 weeks ago). This version requires node >=18, but serverless is currently on node 14.

## Solution

Fix `serverless-plugin-include-dependencies` at v5.0.0, which is its one and only v5 release. v5 is compatible with node 14.


gweiying (approved):
lgtm!

PR #2296, opened 2024-03-18
Chore/fix user query

gweiying:
## Problem
Optimising our current query to check if short link exists. Previous query was doing an unnecessary join on the users table, causing the query duration to spike.

Screenshot 2024-03-18 at 5 34 26 PM

## Solution
Replaced query to only check in urls table instead of joining urls with user table.

## Before & After Screenshots

**BEFORE**:
Screenshot 2024-03-18 at 5 44 04 PM

**AFTER**:
Screenshot 2024-03-18 at 5 35 38 PM

## Tests
- [x] Ran load tests on staging to check if query duration reduced.

**E2E Manual tests**
- [x] If short link is already used, should not be able to create short link
- [x] When creating new links via APIs, should be able to create shortened link without specifying short link
- [x] When creating new links via APIs, if short link is already used, should not be able to create short link

### DIFF
--- a/src/server/repositories/UrlRepository.ts
+++ b/src/server/repositories/UrlRepository.ts
@@ -184,9 +184,10 @@ export class UrlRepository implements UrlRepositoryInterface {
   ) => {
     try {
       // Cache lookup
+      await this.getLongUrlFromCache(shortUrl)
       // if long url does not exist, throws error
-      const longUrl = await this.getLongUrlFromCache(shortUrl)
-      return !longUrl
+      // return false if no error since long url exists
+      return false
     } catch {
       // Cache failed, look in database
       const url = await Url.findOne({
--- a/src/server/repositories/UrlRepository.ts
+++ b/src/server/repositories/UrlRepository.ts
@@ -175,6 +175,27 @@ export class UrlRepository implements UrlRepositoryInterface {
     return this.urlMapper.persistenceToDto(newUrl)
   }
 
+  /**
+   * @param  {string} shortUrl Short url.
+   * @returns {Promise<boolean>} Returns true if shortUrl is available and false otherwise.
+   */
+  public isShortUrlAvailable: (shortUrl: string) => Promise<boolean> = async (
+    shortUrl,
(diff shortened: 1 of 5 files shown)


halfwhole (approved):
lgtm!

src/server/repositories/UrlRepository.ts
```diff
@@ -175,6 +175,27 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+   * @returns {Promise<boolean>} Returns true if shortUrl is available and false otherwise.
+   */
+  public isShortUrlAvailable: (shortUrl: string) => Promise<boolean> = async (
+    shortUrl,
+  ) => {
+    try {
+      // Cache lookup
+      // if long url does not exist, throws error
+      const longUrl = await this.getLongUrlFromCache(shortUrl)
+      return !longUrl
```
halfwhole: nit: small optimisation but unnecessary
```suggestion
      return true
```
edit should be `false` OOPS

src/server/repositories/UrlRepository.ts
```diff
@@ -175,6 +175,28 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+      // Cache lookup
+      await this.getLongUrlFromCache(shortUrl)
+      // if long url does not exist, throws error
+      // return false if no error since long url exists
+      return false
+    } catch {
+      // Cache failed, look in database
+      const url = await Url.findOne({
+        where: { shortUrl },
+      })
```
timotheeg, line 195: /nit/micro-optimization: since we only care for an `exists`, you could reduce the selection to just the `shortUrl` field. Otherwise the query seems to be sent like this atm (according to the traces):
```
SELECT shortUrl, longUrl, state, isFile, contactEmail, description, source, tagStrings, createdAt, updatedAt, userId FROM urls WHERE url . shortUrl = ?
```

```suggestion
      const url = await Url.findOne({
        attributes: ['shortUrl'],
        where: { shortUrl },
      })
```

PR #2305, opened 2024-03-21
chore: replace safe browsing with web risk

gweiying:
## Context

Replaces Google Safe Browsing with Google Web Risk API, which has no daily quotas on number of links scanned. 

**Notes**
- There seems to be minor differences in the threat types detected by Web Risk and Safe Browsing.
  -  Web Risk: `MALWARE` `SOCIAL_ENGINEERING` `UNWANTED_SOFTWARE` `SOCIAL_ENGINEERING_EXTENDED_COVERAGE`
  - Safe Browsing: `MALWARE` `SOCIAL_ENGINEERING` `UNWANTED_SOFTWARE` `POTENTIALLY_HARMFUL_APPLICATION`
  - Checking w the Google team right now whether this will cause any adverse effect /regression
- Required abit of refactoring as the return types for Safe Browsing is different from Web Risk
  Safe Browsing return:
  ```
  // pass in list of urls, 
  // if threats found, returns list of matched threats
  {"matches": [{  "threatType":  "MALWARE", "platformType":  "WINDOWS", "threatEntryType": "URL", "threat": {"url": "http://www.urltocheck1.org/"}, "threatEntryMetadata": { "entries": [{ "key": "malware_threat_type", "value": "landing" }]}, "cacheDuration": "300.000s"}]}
  // if no threats 
  {}
  ```
  Web Risk return:
  ```
  // only single lookup endpoint available
  // if threat found, returns threat as object
  { "threat": { "threatTypes": ["MALWARE"], "expireTime": "2024-03-20T05:29:41.898456500Z"} }
  // if no threats
  {}
  ```

## Tests
Tests on staging
- [x] Try creating short link to malicious url, should see error "Link is likely malicious, please contact us for help"
- [x] Create short link to safe url, should be able to create short link
- [x] [API] Create short link to safe url, should be able to create short link
- [x] [API] Try creating short link to malicious url, should see error "Link is likely malicious, please contact us for help" 
- [x] [Bulk] Try creating short links with bulk upload of 500 links, all safe, should be able to bulk create
- [x] [Bulk] Try creating short links with bulk upload of 500 links, at least one malicious link, should not be able to bulk create
- [x] See usage on Web Risk API on Google console

## Deploy Notes
Before deploying,
- [ ] Activate Web Risk API on edu prod
- [ ] Activate Web Risk API on go prod
- [ ] Activate Web Risk API on health prod

After deployment, tests on production, 3 environments
- [ ] Try creating short link to malicious url, should see error "Link is likely malicious, please contact us for help"
- [ ] Create short link to safe url, should be able to create short link
- [ ] [API] Create short link to safe url, should be able to create short link
- [ ] [API] Try creating short link to malicious url, should see error "Link is likely malicious, please contact us for help" 
- [ ] [Bulk] Try creating short links with bulk upload of 500 links, all safe, should be able to bulk create
- [ ] [Bulk] Try creating short links with bulk upload of 500 links, at least one malicious link, should not be able to bulk 

None, reusing existing Safe Browsing API keys

### DIFF
--- a/src/server/modules/threat/interfaces/SafeBrowsingRepository.ts
+++ b/src/server/modules/threat/interfaces/SafeBrowsingRepository.ts
@@ -1,17 +1,17 @@
-import { HasCacheDuration } from '../../../repositories/types'
+import { WebRiskThreat } from '../../../repositories/types'
 
 export interface SafeBrowsingRepository {
   /**
    * Sets or replaces the Safe Browsing threat matches associated with a URL.
    * @param  {string} url The URL.
-   * @param  {Array<HasCacheDuration>} matches The threat matches returned by Safe Browsing for the URL.
+   * @param  {WebRiskThreat} threat The threat match returned by Safe Browsing for the URL.
    */
-  set(url: string, matches: HasCacheDuration[]): Promise<void>
+  set(url: string, threat: WebRiskThreat): Promise<void>
 
   /**
    * Retrieves Safe Browsing threat matches for a URL, if any.
    * @param  {string} url The URL.
-   * @returns {Promise<Array<HasCacheDuration> | null>} An array of threat matches, or null if none.
+   * @returns {Promise<WebRiskThreat | null>} Threat match or null if none.
    */
-  get(url: string): Promise<HasCacheDuration[] | null>
+  get(url: string): Promise<WebRiskThreat | null>
 }
--- a/src/server/modules/threat/mappers/SafeBrowsingMapper.ts
+++ b/src/server/modules/threat/mappers/SafeBrowsingMapper.ts
@@ -1,27 +1,25 @@
 /* eslint-disable class-methods-use-this, lines-between-class-members, no-dupe-class-members */
 import { injectable } from 'inversify'
 import { TwoWayMapper } from '../../../mappers/TwoWayMapper'
-import { HasCacheDuration } from '../../../repositories/types'
+import { WebRiskThreat } from '../../../repositories/types'
 
 @injectable()
-export class SafeBrowsingMapper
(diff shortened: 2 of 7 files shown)


halfwhole (approved):
LGTM!
- agree that it looks ok to just set the cache time to 300s, no real need to follow the given `expireTime` so closely
- minor question, non-blocking: `SOCIAL_ENGINEERING_EXTENDED_COVERAGE` wasn't there before and so it looks like we don't need to add it right away, but i'm wondering if it's good to add in eventually?
- previously `isThreatBulk` could handle bulk requests all at once (up to 500), but now it sends individual requests for each URL, should we test that bulk uploads are okay too?
- in the future, refactoring to use "web risk" instead of "safe browsing" would be nice!

gweiying:
> LGTM!
> 
>     * agree that it looks ok to just set the cache time to 300s, no real need to follow the given `expireTime` so closely
> 
>     * minor question, non-blocking: `SOCIAL_ENGINEERING_EXTENDED_COVERAGE` wasn't there before and so it looks like we don't need to add it right away, but i'm wondering if it's good to add in eventually?
> 
>     * previously `isThreatBulk` could handle bulk requests all at once (up to 500), but now it sends individual requests for each URL, should we test that bulk uploads are okay too?
> 
>     * in the future, refactoring to use "web risk" instead of "safe browsing" would be nice!

Good points! Thanks for the quick review.

Added in `SOCIAL_ENGINEERING_EXTENDED_COVERAGE` as part of the threat types covered, I agree that it's good to add and err on the safe side!

Also ran tests on bulk upload and updated the test list above, thanks for flagging out.

halfwhole (approved):

PR #2341, opened 2024-08-27
build: fix link to ibm plex sans font

halfwhole:
This PR fixes some build issues:
1. `npm run dev` fails because the link to retrieve the IBM plex sans font is broken
   - Solution: replace the broken link (https://github.com/IBM/plex/blob/master/IBM-Plex-Sans/fonts/complete/otf/IBMPlexSans-Regular.otf?raw=true) with the updated link (https://github.com/IBM/plex/blob/master/packages/plex-sans/fonts/complete/otf/IBMPlexSans-Regular.otf?raw=true)
   - There could be better solutions (e.g. download a copy of the OTF file into this repo, use a fixed commit hash instead of referencing the `master` branch, use Google fonts API), but refraining from making further more extensive changes for now
2. GitHub actions are failing because `docker-compose` no longer exists, see [this discussion](https://github.com/orgs/community/discussions/116610)
   - Solution: replaced it with `docker compose` (without the dash)

## Tests

- [x] Test on staging that QR code generation works (png + svg)

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -9,8 +9,8 @@ RUN sudo apt-get update \
 RUN sudo mkdir -p /docker-entrypoint-initaws.d
 RUN sudo chown gitpod /docker-entrypoint-initaws.d
 
-# Installs IBMPlexSans-Regular.ttf for QRCodeService.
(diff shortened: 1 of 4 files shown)


thanhdatle (approved):
LGTM

PR #2381, opened 2025-05-27
fix: remove statistics repo

seaerchin:
## Problem
- the statistics repository causes thundering herd problems because it invalidates the query to our db on timeout then repeatedly requeries. 
- we also don't need an immediately "live" view of the stats, can just update every quarter, so this has been shifted out into env vars
 
## Solution
- shift the stats out into an env var, which is to be injected at runtime
- mass deletion of `StatisticsRepository` and related files since we no longer use it

## Notes
- This PR **has not been tested**
- The numbers are gibberish -> i tried to look at the numbers on `go.gov.sg` but it showed 0 for all so...


src/server/config.ts
halfwhole, line 1: should be able to remove `STATISTICS_EXPIRY` and `statisticsExpiry` from this file too (should be unused)

halfwhole (approved):
lgtm! shall we test it out on staging?

edit: a bit out of date 🫠, but the statistics on go.gov.sg (as of 24 mar) were these:
- 77,288 public officers onboard
- 28,151,439 short links created
- 666,820,545 clicks

src/server/config.ts
seaerchin: oki, done!

seaerchin:
pushed to staging, **didn't add env var yet**

Issue #2153, opened 2023-02-16
Remove trailing empty rows in uploaded CSV file for bulk link creation


halfwhole:
A number of users have been complaining that their bulk link uploads have been failing because the last row of their uploaded CSV file is empty, and they seem to be unable to debug and fix the issue on their own. It would be nice if Go's backend could help to remove or ignore any extra empty rows when parsing the CSV file, so users are less likely to encounter upload errors

ketanwani:
Hello, I have attempted a fix for this issue. Can you pls review: https://github.com/opengovsg/GoGovSG/pull/2389

PR #2389, opened 2025-07-08
fix: handle empty rows in bulk upload CSV files and refactor generate…

ketanwani:
## Problem

A number of users have been complaining of failed bulk link uploads because the last row of their uploaded CSV file is empty, and they seem to be unable to debug and fix the issue on their own. It would be nice if Go's backend could help to remove or ignore any extra empty rows when parsing the CSV file, so users are less likely to encounter upload errors

Closes [insert issue #]
https://github.com/opengovsg/GoGovSG/issues/2153

## Solution
Handle logic to ignore excel/csv rows which are empty in the end.

## Before & After Screenshots

**BEFORE**:

https://github.com/user-attachments/assets/157a6127-534d-48e8-a8d9-5aef54bdc592

**AFTER**:

https://github.com/user-attachments/assets/d92e607a-4a0b-4303-9397-eb1de9ebfe04

### DIFF
--- a/src/server/modules/bulk/services/BulkService.ts
+++ b/src/server/modules/bulk/services/BulkService.ts
@@ -35,7 +35,7 @@ export class BulkService implements interfaces.BulkService {
 
     return new Promise((resolve, reject) => {
       Papa.parse(dataString, {
-        skipEmptyLines: false,
+        skipEmptyLines: true,
         delimiter: ',',
         complete: () => {
           // check for empty file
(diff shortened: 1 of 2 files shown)


src/server/modules/bulk/services/BulkService.ts
```diff
@@ -75,60 +75,38 @@ export class BulkService implements interfaces.BulkService {
...
+                errorMessage += 'contains invalid url'
+              } else if (isNotCircularRedirect) {
+                errorMessage += 'contains circular redirect'
+              } else if (!noParsingError) {
+                errorMessage += 'has parsing error'
+              }
+              dogstatsd.increment(BULK_VALIDATION_ERROR, 1, 1, [
+                `${BULK_VALIDATION_ERROR_TAGS.isValidUrl}`,
+              ])
+              throw new Error(errorMessage)
```
seaerchin, line 108: not entirely sure - can there only be 1 error per line? if there's possibly >1, then the formatting might be abit awkward

seaerchin (commented):
no issues apart from a minor formatting issue

PR #2395, opened 2025-07-10
chore: escape raw html

seaerchin:
## Problem
Go uses raw html for templating without escaping them. This PR adds that in. 

## Solution
add escaping into the html template (ejs [docs](https://ejs.co/#docs))


adriangohjw:
bugbot run

seaerchin:
when will ai replace me :kekw:

dcshzj (approved):

PR #2396, opened 2025-07-11
chore: fix compute issue from tags

seaerchin:
## Problem
https://opengovproducts.slack.com/archives/C0949RH4NTA/p1752216333458559

## Solution
- don't use `helpers.message` because it's compute intensive 
- return a hardcoded message cos it won't be seen by frontend
- update initial regex for creation of url to follow frontend's requirements


adriangohjw:
bugbot run

harishv7 (approved):
lgtm, pls test on staging first

PR #2400, opened 2025-07-21
isom-2003 feat(otp): update OTP verification to include IP address

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-2003/vapt-active-denial-of-service-of-login-to-gogovsg

## Solution

**Features**:

- use `email:ip` as redis key instead of just `email`

**Improvements**:

- add test case for this

## Tests

- refer to Linear ticket for steps to reproduce

### DIFF
--- a/src/server/modules/auth/LoginController.ts
+++ b/src/server/modules/auth/LoginController.ts
@@ -68,7 +68,7 @@ export class LoginController {
     const { email, otp }: VerifyOtpRequest = req.body
 
     try {
-      const user = await this.authService.verifyOtp(email, otp)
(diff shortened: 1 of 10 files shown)


adriangohjw:
bugbot run

dcshzj (approved):
I think ok, but feels a bit like the blind leading the blind lol, need to really test on staging to make sure everything is still working as expected.

adriangohjw:
> I think ok, but feels a bit like the blind leading the blind lol, need to really test on staging to make sure everything is still working as expected.

@dcshzj yeah i agree :sadge:

PR #2401, opened 2025-07-21
isom-2008 feat(redirect): enhance checks for bypassing transition page

adriangohjw:
## Problem

Closes https://linear.app/ogp/issue/ISOM-2008/vapt-insufficient-referrer-validation-bypasses-transition-page

## Solution

**Improvements**:

- use built-in `URL` function to check instead of `startsWith`
- extract into private method for readability 
- add tests

## Tests

- refer to Linear ticket for steps to reproduce

### DIFF
--- a/src/server/modules/redirect/services/RedirectService.ts
+++ b/src/server/modules/redirect/services/RedirectService.ts
@@ -68,7 +68,7 @@ export class RedirectService {
       }
     }
 
-    const isFromTrustedPage = referrer.startsWith(ogUrl)
+    const isFromTrustedPage = RedirectService.isFromTrustedPage(referrer)
 
     const renderTransitionPage =
       !this.cookieArrayReducerService.userHasVisitedShortlink(
@@ -87,6 +87,23 @@ export class RedirectService {
       redirectType: renderTransitionPage
         ? RedirectType.TransitionPage
         : RedirectType.Direct,
+    }
+  }
+
+  /**
+   * Checks whether the referrer is from a trusted page (same origin as ogUrl).
+   * This prevents malicious sites from bypassing the transition page.
+   * @param {string} referrer - The referrer URL to check.
+   * @returns {boolean} - True if referrer is from trusted origin, false otherwise.
+   */
+  private static isFromTrustedPage(referrer: string): boolean {
+    try {
+      const referrerUrl = new URL(referrer)
+      const trustedUrl = new URL(ogUrl)
+      return referrerUrl.origin === trustedUrl.origin
+    } catch {
+      // If referrer is not a valid URL, treat as untrusted
+      return false
     }
   }
 
--- a/src/server/modules/redirect/services/__tests__/RedirectService.test.ts
+++ b/src/server/modules/redirect/services/__tests__/RedirectService.test.ts
@@ -59,7 +59,7 @@ describe('RedirectService', () => {
     )
   })
 
-  describe('referrer validation', () => {
+  describe('redirectFor', () => {
     it('should allow direct redirect for exact ogUrl match when user has not visited before', async () => {
       const result = await redirectService.redirectFor(
         'test',
--- a/src/server/modules/redirect/services/__tests__/RedirectService.test.ts
+++ b/src/server/modules/redirect/services/__tests__/RedirectService.test.ts
@@ -1,0 +1,193 @@
+import { RedirectService } from '../RedirectService'
+import { UrlRepositoryInterface } from '../../../../repositories/interfaces/UrlRepositoryInterface'
+import { CrawlerCheckService } from '../CrawlerCheckService'
+import { CookieArrayReducerService } from '../CookieArrayReducerService'
+import { LinkStatisticsService } from '../../../analytics/interfaces'
+import { RedirectType } from '../..'
+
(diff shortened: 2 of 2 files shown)


adriangohjw:
bugbot run

adriangohjw:
> ### Bug: Trusted Referrer Tests Expect Incorrect Redirect
> The `RedirectService.test.ts` file contains incorrect expectations for trusted referrer scenarios. Tests for referrers from the same origin as `ogUrl` (e.g., exact match, with path, or with query parameters) currently expect `RedirectType.TransitionPage`. However, the `RedirectService`'s logic dictates `RedirectType.Direct` for these cases, as a trusted referrer should bypass the transition page.

fixed tests in https://github.com/opengovsg/GoGovSG/pull/2401/commits/29b40c8a169dcf2e4babd2f95deb861a557cfb5e

adriangohjw:
> ### Bug: Malformed URL in Security Test
> The test "should NOT allow transition page bypass for domain containing ogUrl" constructs a malformed referrer URL. The template literal `https://staging.${ogUrl}.malicious.com` results in `https://staging.https://go.gov.sg.malicious.com` because `ogUrl` already includes `https://`, creating a double protocol. While the test passes (as invalid URLs are correctly treated as untrusted), it does not test the intended scenario of a malicious domain containing the trusted domain name. The URL should be `https://staging.go.gov.sg.malicious.com`.

fixed in https://github.com/opengovsg/GoGovSG/pull/2401/commits/9206094819e29b9b2b14119186527f63f2cf3c80

src/server/modules/redirect/services/RedirectService.ts
```diff
@@ -68,7 +68,7 @@ export class RedirectService {
       }
     }
 
-    const isFromTrustedPage = referrer.startsWith(ogUrl)
+    const isFromTrustedPage = RedirectService.isFromTrustedPage(referrer)
```
dcshzj, line 71: issue: should use `this` instead of `RedirectService`

src/server/modules/redirect/services/__tests__/RedirectService.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    // Create service instance with mocked dependencies
+    redirectService = new RedirectService(
+      mockUrlRepository as unknown as UrlRepositoryInterface,
+      mockCrawlerCheckService as unknown as CrawlerCheckService,
+      mockCookieArrayReducerService as unknown as CookieArrayReducerService,
+      mockLinkStatisticsService as unknown as LinkStatisticsService,
+    )
+  })
+
+  describe('referrer validation', () => {
```
dcshzj: nit: can name it after the function instead? Easier to identify and debug

src/server/modules/redirect/services/__tests__/RedirectService.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      'test',
+    ])
+
+    // Create service instance with mocked dependencies
+    redirectService = new RedirectService(
+      mockUrlRepository as unknown as UrlRepositoryInterface,
+      mockCrawlerCheckService as unknown as CrawlerCheckService,
+      mockCookieArrayReducerService as unknown as CookieArrayReducerService,
+      mockLinkStatisticsService as unknown as LinkStatisticsService,
+    )
```
dcshzj: suggestion: I think this init can be outside of the `beforeEach` and `describe`, might help in making tests run a bit faster

src/server/modules/redirect/services/RedirectService.ts
adriangohjw: actually im not super familiar with the framework, but i referenced this from an existing code

https://github.com/opengovsg/GoGovSG/blob/2d94333681183713289901ec0cbe11d1ade024d5/src/server/modules/redirect/services/RedirectService.ts#L47-L49

https://github.com/opengovsg/GoGovSG/blob/2d94333681183713289901ec0cbe11d1ade024d5/src/server/modules/redirect/services/RedirectService.ts#L98-L100

curious - what's the issue for this?

src/server/modules/redirect/services/__tests__/RedirectService.test.ts
adriangohjw: updated in https://github.com/opengovsg/GoGovSG/pull/2401/commits/36b91b063ff399ce2783c76f9bcc005a33d39cd7 👍

src/server/modules/redirect/services/__tests__/RedirectService.test.ts
adriangohjw: updated in https://github.com/opengovsg/GoGovSG/pull/2401/commits/1684f2601ba30a977c3f0a0856935042bf40d255 👍

src/server/modules/redirect/services/RedirectService.ts
dcshzj: Ah oops sorry was reviewing on mobile, so didn't see the `static` part. Doing `RedirectService.isFromTrustedPage` is ok since `isFromTrustedPage` is a static function. If otherwise then should use `this`.

dcshzj (approved):

PR #2402, opened 2025-07-22
feat: add safeBrowsingExpiry column to urls table

dcshzj:
## Problem

We need to support the caching of the safe browsing scan results for at least 24 hours.

## Solution

**Features**:

- Add a new `safeBrowsingExpiry` column to the `urls` table to support storing of the safe browsing scan result's expiry date.
    - This date represents the time in which no further safe browsing scanning is required.
    - The duration will be decided later by the producer of this information, so there is flexibility in determining when a scan result expires.

## Tests

- [ ] Run the DB migration to add the `safeBrowsingExpiry` column. Verify that the new column appears in the `urls` table.
- [ ] Create a new link, verify that link creation still works as intended.
- [ ] Edit the new link you have created, verify that link editing still works as intended.

## Deploy Notes

- [ ] Run DB migration to add the `safeBrowsingExpiry` column.


src/server/db_migrations/20250721_add_urls_safebrowsing_expiry.sql
```diff
@@ -0,0 +1,11 @@
...
+-- backwards-compatible with the current codebase as it allows NULL values.
+
+BEGIN TRANSACTION;
+
+ALTER TABLE urls ADD "safeBrowsingExpiry" TIMESTAMP WITH TIME ZONE;
+
+COMMIT;
+
+-- Down migration
+-- ALTER TABLE urls DROP COLUMN "safeBrowsingExpiry";
```
seaerchin: question: how will we run this migration since it looks like we're not using `sequelize-cli`?

seaerchin (commented):
1 comment, could we also run this on `staging` (technically `edge`) to ensure that it's working properly?

i think might be hard to check manually so there's some value in having unit tests here to make sure that we rescan past 24 hours

src/server/db_migrations/20250721_add_urls_safebrowsing_expiry.sql
seaerchin: from sequelize-cli [docs](https://github.com/sequelize/cli): `sequelize migration:generate` -> `sequelize migration:generate --name migration-name`

seaerchin: (it's hidden somewhere [here](https://sequelize.org/docs/v6/other-topics/migrations/) using `sequelize-cli`: `npx sequelize-cli migration:generate --name migration-skeleton`)

dcshzj: Updated in db085469f146dfcc4d40408ef9bdf49c87b61d99.

seaerchin (approved):

PR #2403, opened 2025-07-22
feat: check for malicious link just before redirect

dcshzj:
## Problem

Closes ISOM-2006.

## Solution

**Features**:

- Add a check on the destination URL to ensure that it is not malicious just before redirecting the user.
- Save the malicious scan result into the database with an expiry date.
- Automatically disable the link and email the link owner when the destination URL is detected to be malicious.

## Tests

**TO BE ADDED**

### DIFF
--- a/src/server/modules/redirect/services/RedirectService.ts
+++ b/src/server/modules/redirect/services/RedirectService.ts
@@ -6,6 +6,9 @@ import { RedirectResult, RedirectType } from '..'
 import { LinkStatisticsService } from '../../analytics/interfaces'
 import { logger, ogUrl } from '../../../config'
 import { CookieArrayReducerService, CrawlerCheckService } from '.'
+import { UrlThreatScanService } from '../../threat/interfaces'
+import { getSafeBrowsingExpiryDate } from '../../../util/safeBrowsing'
+import { UrlManagementService } from '../../user/interfaces'
 
 @injectable()
 export class RedirectService {
@@ -17,6 +20,10 @@ export class RedirectService {
 
   private linkStatisticsService: LinkStatisticsService
 
+  private urlThreatScanService: UrlThreatScanService
+
+  private urlManagementService: UrlManagementService
+
   public constructor(
     @inject(DependencyIds.urlRepository) urlRepository: UrlRepositoryInterface,
     @inject(DependencyIds.crawlerCheckService)
@@ -25,11 +32,17 @@ export class RedirectService {
     cookieArrayReducerService: CookieArrayReducerService,
     @inject(DependencyIds.linkStatisticsService)
     linkStatisticsService: LinkStatisticsService,
+    @inject(DependencyIds.urlThreatScanService)
+    urlThreatScanService: UrlThreatScanService,
+    @inject(DependencyIds.urlManagementService)
+    urlManagementService: UrlManagementService,
   ) {
     this.urlRepository = urlRepository
     this.crawlerCheckService = crawlerCheckService
     this.cookieArrayReducerService = cookieArrayReducerService
     this.linkStatisticsService = linkStatisticsService
+    this.urlThreatScanService = urlThreatScanService
+    this.urlManagementService = urlManagementService
   }
 
   public redirectFor: (
@@ -51,7 +64,34 @@ export class RedirectService {
     const shortUrl = rawShortUrl.toLowerCase()
 
     // Find longUrl to redirect to
-    const longUrl = await this.urlRepository.getLongUrl(shortUrl)
+    const { longUrl, isFile, safeBrowsingExpiry } =
+      await this.urlRepository.getLongUrl(shortUrl)
+
+    // Validate that the longUrl is not a malicious link
+    const isSafeBrowsingResultExpired =
+      !isFile &&
+      (!safeBrowsingExpiry ||
+        new Date(safeBrowsingExpiry).getTime() < Date.now())
+
+    if (isSafeBrowsingResultExpired) {
+      const isThreat = await this.urlThreatScanService.isThreat(longUrl)
+      if (isThreat) {
+        logger.warn(
+          `Malicious link attempt: ${longUrl} was detected as malicious for shortUrl ${shortUrl}`,
+        )
+
+        // Deactivate the short link and warn the short link owner
+        await this.urlManagementService.deactivateMaliciousShortUrl(shortUrl)
+
+        // NOTE: We return a 404 error here to make the user experience the same
+        // as if the short link was deactivated/not found for simplicity and to
+        // avoid inducing user panic.
+        throw new NotFoundError('Malicious link detected')
+      }
+      // Store the result of the threat scan in the database
+      const expiry = getSafeBrowsingExpiryDate({ longUrl })
+      await this.urlRepository.updateSafeBrowsingExpiry(shortUrl, expiry)
+    }
 
     // Update clicks and click statistics in database.
     try {
--- a/src/server/repositories/UrlRepository.ts
+++ b/src/server/repositories/UrlRepository.ts
@@ -14,6 +14,7 @@ import { FileVisibility, S3Interface } from '../services/aws'
 import { UrlRepositoryInterface } from './interfaces/UrlRepositoryInterface'
 import {
   BulkUrlMapping,
+  RedirectDestination,
   StorableFile,
   StorableUrl,
   UrlDirectory,
@@ -27,6 +28,7 @@ import { DirectoryQueryConditions } from '../modules/directory'
 import { extractShortUrl, sanitiseQuery } from '../util/parse'
 import { TagRepositoryInterface } from './interfaces/TagRepositoryInterface'
 import { TAG_SEPARATOR } from '../../shared/constants'
+import { getSafeBrowsingExpiryDate } from '../util/safeBrowsing'
 
 const { Public, Private } = FileVisibility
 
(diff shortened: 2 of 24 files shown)


src/server/modules/redirect/services/RedirectService.ts
```diff
@@ -51,7 +64,34 @@ export class RedirectService {
     const shortUrl = rawShortUrl.toLowerCase()
 
     // Find longUrl to redirect to
-    const longUrl = await this.urlRepository.getLongUrl(shortUrl)
+    const { longUrl, isFile, safeBrowsingExpiry } =
+      await this.urlRepository.getLongUrl(shortUrl)
+
+    // Validate that the longUrl is not a malicious link
+    const isSafeBrowsingResultExpired =
```
seaerchin, line 71: question: it looks like we don't scan again for files; i know we scan on initial upload, but what happens if the user changes the initial file (file A) to another file (file B)? 

this is out of scope of this PR, but i'll double check later to make sure we rescan on file updates

src/server/modules/redirect/services/RedirectService.ts
```diff
@@ -51,7 +64,34 @@ export class RedirectService {
...
+      !isFile &&
+      (!safeBrowsingExpiry ||
+        new Date(safeBrowsingExpiry).getTime() < Date.now())
```
seaerchin, line 74: [nit] there might be a gap betewen the `getTime` and `Date.now` if the TZs are different. just double confirming that we're using the same TZ for both?

src/server/modules/redirect/services/RedirectService.ts
```diff
@@ -51,7 +64,34 @@ export class RedirectService {
...
+
+    if (isSafeBrowsingResultExpired) {
+      const isThreat = await this.urlThreatScanService.isThreat(longUrl)
```
seaerchin, line 77: [issue] actually there's an issue here, on `urlThreatScanService.isThreat`. the response out from google contains a `expireTime`, which we should be checking against (https://cloud.google.com/web-risk/docs/reference/rpc/google.cloud.webrisk.v1#threaturi). 

this isn't a super big deal (because we cache for 24 hours only), but i think the fix is simple enough to do

src/server/repositories/UrlRepository.ts
```diff
@@ -521,7 +528,17 @@ export class UrlRepository implements UrlRepositoryInterface {
             )
             return
           }
-          resolve(cacheLongUrl)
+
+          try {
+            const redirectDestination = JSON.parse(cacheLongUrl)
```
seaerchin, line 535: [issue] i think there are 2 issues here:

first, i think we updated the data structure from string -> json, which might cause issues when we're trying to read the new value. i think our new code cannot `reject` outright and we'd need to handle the fallback of parsing a string value rather than our json encoded value. 

secondly, because our deploy window is so long, we need to handle the case where our old code reads a json encoded value. this will cause our existing flow to mess up, so it needs to be handled properly too. 

from our offline discussion, i think we can do a small release so that we wrap the `JSON.parse` method here to try reading it as a JSON value first, then fallback to raw string

src/server/repositories/UrlRepository.ts
```diff
@@ -197,21 +200,20 @@ export class UrlRepository implements UrlRepositoryInterface {
...
-      // Cache lookup
-      return await this.getLongUrlFromCache(shortUrl)
-    } catch {
-      // Cache failed, look in database
-      const longUrl = await this.getLongUrlFromDatabase(shortUrl)
-      this.cacheShortUrl(shortUrl, longUrl).catch((error) =>
-        logger.error(`Unable to cache short URL: ${error}`),
-      )
-      return longUrl
+  public getLongUrl: (shortUrl: string) => Promise<RedirectDestination> =
```
seaerchin, line 203: we need to relax this return condition to `RedirectDestination | string` due to the previously mentioned issue of data format changing

src/server/repositories/UrlRepository.ts
```diff
@@ -604,6 +632,36 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+    const url = await Url.findOne({ where: { shortUrl } })
+    if (!url) {
+      throw new NotFoundError(`Url with shortUrl ${shortUrl} not found`)
+    }
+
+    await this.update(
+      { shortUrl },
+      { safeBrowsingExpiry: expiry.toISOString() },
+      undefined,
+    )
```
seaerchin, line 674: [question] the `update` method invalidates our redis cache, so now new accesses to get the expiry time would also hit our db, ensuring the latest `expiry` right?

seaerchin (commented):
issues wrt the data structure being changed impacting the code, especially because we rely on it for doing our code scanning.

seaerchin:
(some stuff that claude pointed out)

1. our scanning takes place in the hot path, which might delay the redirect, especially because this is `user -> us -> webrisk` 
2. rate limiting for threat scanning API call (i think we don't need this because we cache our results and we also check if we need to scan first)
3. silent failure on `UrlManagementService:237` when we try to lookup users (<- afaict, we don't throw an error but we do catch the `null` case)

adriangohjw:
bugbot run

src/server/modules/redirect/services/RedirectService.ts
dcshzj: Yeah for re-uploads, the file is rescanned. Ref:

https://github.com/opengovsg/GoGovSG/blob/705b25f9e9aec6653eb16ce1e767800eeab91579/src/server/api/user/index.ts#L141-L151

src/server/modules/redirect/services/RedirectService.ts
dcshzj: Hmm the code to set the `safeBrowsingExpiry` is also done using `Date.now()`, so that should cancel out the time zone differences (i.e. they are generated using the same time zone). Ref:

https://github.com/opengovsg/GoGovSG/blob/705b25f9e9aec6653eb16ce1e767800eeab91579/src/server/util/safeBrowsing.ts#L13

src/server/repositories/UrlRepository.ts
dcshzj: Let me do a split of the PRs so they get released separately.

src/server/modules/redirect/services/RedirectService.ts
dcshzj: Good point yeah, let me update.

src/server/repositories/UrlRepository.ts
dcshzj: Yup that's correct, it purges the record from the redis cache so the next retrieval will need to go to DB, then get cached in redis with the new expiry time.

src/server/repositories/UrlRepository.ts
```diff
@@ -521,7 +528,17 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+          try {
+            const redirectDestination = JSON.parse(cacheLongUrl)
+            resolve(redirectDestination)
+          } catch (_) {
+            reject(
+              new NotFoundError(
+                `longUrl not found in cache:\tshortUrl=${shortUrl}`,
+              ),
+            )
+          }
```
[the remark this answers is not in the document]
dcshzj: Wonderful, this was the same point raised by @seaerchin earlier. To mitigate, we will do as a 3-step release:
1. Update app code to handle both cases, using the JSON.parse function optimistically, then falling back to using strings
2. Update app code to start inserting using the new JSON object
3. Update app code to remove the handling of the string case

src/server/modules/redirect/services/RedirectService.ts
dcshzj: Actually hmm no, seems like the change is more complex than I thought. The `isThreat` returns a boolean which is relatively straightforward to do, but we will need to handle for `isThreatBulk` which itself does a `.every()`, which will have quite a bit of changes.

I think 24 hours should be relatively short enough in this case?

src/server/repositories/UrlRepository.ts
dcshzj: Updated in 647fee7becd9a080e4cd8dd8216a7695162bf43f, will push a second PR to revert back the changes

src/server/modules/redirect/services/RedirectService.ts
seaerchin: ok, i think 24 hours is fine ba

src/server/repositories/UrlRepository.ts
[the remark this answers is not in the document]
seaerchin: praise bugbot

src/server/repositories/UrlRepository.ts
```diff
@@ -197,21 +200,22 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+    async (shortUrl) => {
+      try {
+        // Cache lookup
+        return await this.getLongUrlFromCache(shortUrl)
+      } catch {
+        // Cache failed, look in database
+        const redirectDestination = await this.getLongUrlFromDatabase(shortUrl)
+        // FIXME: Cache the entire RedirectDestination object once all app
+        // clients are updated to use the new RedirectDestination type.
+        this.cacheShortUrl(shortUrl, redirectDestination.longUrl).catch(
```
seaerchin, line 213: [question] why are we not `await`ing the cache call? doesn't seem like it has an impact, but just curious

seaerchin (approved):

src/server/repositories/UrlRepository.ts
dcshzj: Hmm I think it's cos the adding to cache part can be done later, so it doesn't impact the person being redirected

PR #2407, opened 2025-08-06
feat: start writing RedirectionDestination in redis cache

dcshzj:
## Problem

Split from #2403 

## Solution

**Features**:

- Start writing RedirectDestination into the redis cache

## Tests

## Deploy Notes

- Must be released after #2403 is merged and released!

### DIFF
--- a/src/server/modules/redirect/__tests__/RedirectController.test.ts
+++ b/src/server/modules/redirect/__tests__/RedirectController.test.ts
@@ -21,7 +21,7 @@ import {
   CrawlerCheckService,
   RedirectService,
 } from '../services'
-// import { RedirectDestination } from '../../../repositories/types'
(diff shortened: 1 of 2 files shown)


seaerchin (commented):
generally no issues, but can i be abit more onerous here, cos the blast radius of this chnage is huge. 

would it be possible to add a test case for back-compat (so now, my values in redis are `string` and not `json`) then check that reads still work gracefully.

understand that we do a `try/catch` then resolve with defaults:
```ts
            resolve({
              longUrl: cacheLongUrl,
              isFile: false,
              safeBrowsingExpiry: null,
            })
```

but just paranoid haha

seaerchin (approved):
discarding my previous review, noticed existing tests use a string already

PR #2418, opened 2025-12-04
fix: prevent IPs from being used as long URL

seaerchin:
## Problem
closes https://linear.app/ogp/issue/ISOM-2009/vapt-low-internal-and-private-ip-addresses-as-long-urls

## Solution
- validate that the domain section of the url is not an IP address
- this validator is used across the dashboard, api and bulk upload 
- also validated that this prevents users from changing existing links to have an IP address as the long url 
    - i'm ok w/ the case where users **cannot** change existing long urls that are IP addresses but i haven't validated teh scale of the problem


dcshzj (approved):
Approving first, but I think Copilot got valid points regarding the test coverage, would be good to add.

PR #2419, opened 2025-12-04
fix: lowercase urls to prevent bypass

seaerchin:
## Problem
go has a blacklist for link shorteners, but this check is a case agnostic check, which means that `Bit.ly` for example, passes the chekc but `bit.ly` does not

## Solution
1. make teh check case insensitive by converting the url to lowercase in the check. opted against converting the url to lowercase outright incase case matters for the actual url


dcshzj (approved):
Approving first, also should add tests for this.

PR #2434, opened 2025-12-18
refactor: add migration for urlShortUrl

seaerchin:
## Problem
https://www.notion.so/opengov/Query-timeouts-for-Go-2c577dbba7888057ae33ce5d2b5475bc?source=copy_link

Closes [insert issue #]

## Solution
- add index on url short url 
- this index is also `concurrent` so that execution won't block other db queries

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -1,0 +1,34 @@
+export default {
+  development: {
+    username: process.env.DB_USERNAME,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME,
+    host: process.env.DB_HOSTNAME,
+    port: process.env.DB_PORT,
+    dialect: 'postgres',
+  },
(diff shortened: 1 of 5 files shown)


config/config.js
```diff
@@ -0,0 +1,34 @@
+export default {
+  development: {
```
harishv7: Shall we extract this out into something like:  


```
const baseConfig = {
  username: process.env.DB_USERNAME,
  password: process.env.DB_PASSWORD,
  database: process.env.DB_NAME,
  host: process.env.DB_HOSTNAME,
  port: process.env.DB_PORT,
  dialect: 'postgres',
}
```

Since all the envs are using the same params

harishv7 (approved):
lgtm, 1 minor comment

PR #2442, opened 2026-01-13
refactor: remove find or create

seaerchin:
## Problem
refer to [notion](https://www.notion.so/opengov/Postgres-engine-upgrade-for-Go-2c477dbba788803b8761cfe5e8550c26?source=copy_link#2d277dbba788807f8b91e8c114284b3c) for both an overview of the problem as well as the solution.

This is a temporary fix and should be reverted once the blue/green is done

### DIFF
--- a/src/server/repositories/TagRepository.ts
+++ b/src/server/repositories/TagRepository.ts
@@ -50,22 +50,36 @@ export class TagRepository implements TagRepositoryInterface {
               // This is acceptable because the max number of tags is small (max of 4),
               // the table itself is not large (38k rows on go)
               // and it's indexed (tagString, tagKey, id)
+              const tagWhere = {
+                tagString: tag,
+                tagKey: tag.toLowerCase(),
+              }
+
+              // First, try to find existing tag
               const possibleTag = await Tag.findOne({
                 transaction: t,
-                where: {
-                  tagString: tag,
-                  tagKey: tag.toLowerCase(),
-                },
+                where: tagWhere,
               })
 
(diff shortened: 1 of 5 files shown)


src/server/repositories/TagRepository.ts
```diff
@@ -43,20 +43,34 @@ export class TagRepository implements TagRepositoryInterface {
...
+              // of `findOrCreate`.
+              // This is acceptable because the max number of tags is small (max of 4),
+              // the table itself is not large (38k rows on go)
+              // and it's indexed (tagString, tagKey, id)
+              const possibleTag = await Tag.findOne({
                 where: {
                   tagString: tag,
                   tagKey: tag.toLowerCase(),
                 },
-                transaction: t,
```
dcshzj, line 52: question: should we pass in the transaction here also?

Seems like [sequelize's own implementation](https://github.com/sequelize/sequelize/blob/17f73737bc01bb1d215b06210170f7262c504835/packages/core/src/model.js#L1992-L1994) passes in the same transaction for both the `findOne` and `create` and I think it would be good to retain the same behaviour?

seaerchin: updated - i did a find/create/find pattern for 2 reasons:

1. we have a unique constraint (hence, we need to do a try/catch for the `create` in case another process already creates the record)
2. opted for `find/create/find` over `create/find` because we are read heavy rather than write heavy

dcshzj (approved):

PR #2448, opened 2026-01-29
chore: skip empty lines while parsing csv

seaerchin:
## Problem
- currently, trailing empty lines create an error for the user. however, we can and should handle this on our end

## Solution
- set `skipEmptyLines: greedy` so that we skip whitespace (` `) and empty lines. 
- further, we `.trim` to ensure no whitespace makes it true (jic)

### DIFF
--- a/src/server/modules/bulk/services/BulkService.ts
+++ b/src/server/modules/bulk/services/BulkService.ts
@@ -35,7 +35,7 @@ export class BulkService implements interfaces.BulkService {
 
     return new Promise((resolve, reject) => {
       Papa.parse(dataString, {
-        skipEmptyLines: false,
+        skipEmptyLines: 'greedy',
         delimiter: ',',
         complete: () => {
           // check for empty file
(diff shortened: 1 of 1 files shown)


seaerchin:
see https://github.com/opengovsg/GoGovSG/pull/2389

src/server/modules/bulk/services/BulkService.ts
```diff
@@ -92,7 +92,7 @@ export class BulkService implements interfaces.BulkService {
                     counter + 1
                   }: ${rowData} contains more than one column of data`,
                 )
-              case !isNotEmpty:
+              case isEmpty:
                 dogstatsd.increment(BULK_VALIDATION_ERROR, 1, 1, [
                   `${BULK_VALIDATION_ERROR_TAGS.isNotEmpty}`,
                 ])
```
adriangohjw, line 98: just a bit confused why this is tagging as `BULK_VALIDATION_ERROR_TAGS.isNotEmpty` but seems like unrelated to this change

adriangohjw (approved):

src/server/modules/bulk/services/BulkService.ts
seaerchin: yeea the case naming is v odd

PR #2449, opened 2026-01-29
chore: add 3 new migrations for indexes to optimise runtime

seaerchin:
refer to https://www.notion.so/opengov/Rework-rawDirectory-search-2f777dbba7888071839ef3f8d491402e?source=copy_link for a write-up on this


adriangohjw (approved):
why the title says "3 migrations" ah?

seaerchin:
> why the title says "3 migrations" ah?

i deleted the others - we had migrations to create a trigram index + label tree (with index) + createdAt index but since we only creating the `clicks` index i deleted the additional migratoins

PR #2455, opened 2026-02-02
chore: update workflow

seaerchin:
## Problem
our workflow is very complicated and difficult to read. additionally, the branches are not named conventionally 

## Solution
extract workflows into a resuable one (`build_and_deploy`) then use a matrix strategy to deploy


adriangohjw (approved):
disclaimer: i didnt scrutinize the whole thing line-by-line too carefully (i assume copy-pasting is correct)

PR #2459, opened 2026-02-09
chore: remove `useMaster` true from read queries

seaerchin:
## Problem
our select and count statements are reading from the writer instances and using alot of compute at present. we want to minimize CPU spikes on writer and offload to reader since this isn't time sensitive

## Solution
remove `useMaster: true`

### DIFF
--- a/test/integration/api/user/Urls.test.ts
+++ b/test/integration/api/user/Urls.test.ts
@@ -48,7 +48,8 @@ describe('Url integration tests', () => {
   const longUrl = 'https://example.com'
 
   beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


test/integration/api/user/Urls.test.ts
adriangohjw, line 1: question: why are the tests commented out?

seaerchin: i'm not entirely sure - i think the previous engineers just didn't implement/fix the test

adriangohjw (approved):

PR #2460, opened 2026-02-11
chore: track domain search

seaerchin:
## Problem
we want to track domain search to see how many ppl are using it

## Solution
1. increment metric in dd + track the domains that are searched


src/server/repositories/UrlRepository.ts
```diff
@@ -275,6 +281,34 @@ export class UrlRepository implements UrlRepositoryInterface {
     // split email/domains by space into tokens, also reduces injections
     const likeQuery = emails.map(sanitiseQuery)
 
+    const domains = emails.filter((email) => {
+      return !validator.isEmail(email)
+    })
```
adriangohjw, line 286: if people search by non-domain and non-email e.g. `firstname`, then it's technically not "domain"?

src/server/repositories/UrlRepository.ts
```diff
@@ -275,6 +281,34 @@ export class UrlRepository implements UrlRepositoryInterface {
...
+      dogstatsd.increment(DIRECTORY_SEARCH_DOMAIN, 1, [
+        `environment:${process.env.DD_ENV}`,
+        `env:${process.env.DD_ENV}`,
+      ])
+      domains.forEach((domain) => {
+        dogstatsd.set(`${DIRECTORY_SEARCH_UNIQUE_DOMAINS}`, domain, 1, [
+          `environment:${process.env.DD_ENV}`,
+          `env:${process.env.DD_ENV}`,
+        ])
+      })
```
adriangohjw, line 302: question: im assuming using `set` set the value instead of incrementing. What's the thought process behind setting it v.s. incrementing it?

seaerchin: from the [docs](https://github.com/bdeitte/hot-shots): 

>   // Set: Counts unique occurrences of a stat (alias of unique)
  client.set('my_unique', 'foobar');
  client.unique('my_unique', 'foobarbaz');

i did this bercause i wanted to see what domains are searched

src/server/repositories/UrlRepository.ts
seaerchin: sorry i don't quite get what you mean but this is meant as a catch all - if they're not inputting an email `blah@test.com`, i assume they're searching for domain (only other thing that can search on anw)

adriangohjw: oh i was thinking someone searching for the prefix of the email e.g. `firstname` instead of `firstname@domain.com` - technically its neither "domain" or "validEmails" but its classified as "domain" for now

but thanks for clarifying! a bit misleading but okay to approve since the tracking is a short-term one

adriangohjw (approved):

src/server/repositories/UrlRepository.ts
seaerchin: the prefix won't work cos we search for suffix `like %blah` so i think this case won't be considered!

PR #2530, opened 2026-08-05
chore(deps): bump typescript to v7 + adopt faster swc packages

adriangohjw:
## Summary

Stacked on #2529 (oxc migration). Bumps TypeScript 5.9.3 → 7.0.2 (the Go-native compiler). TS 7.0 ships no programmatic compiler API at all, so this required replacing every tool in the build/test/dev path that depends on it, and fixing real module-evaluation-order bugs that swap surfaced.

## Commits (read in order)

1. **`chore(deps): replace ts-jest, ts-loader and ts-node(-dev) with swc/tsx equivalents`** — closes every compiler-API consumer *before* the bump, so this commit is independently verifiable on TypeScript 5.9.3:
   - `ts-jest` → `@swc/jest` (`.swcrc` added, with `decoratorMetadata`/`legacyDecorator` enabled for inversify's DI to keep working). We already ran Jest with `isolatedModules` + `diagnostics: warnOnly`, so we were never relying on ts-jest's type-checking.
   - `ts-loader` → `swc-loader`. Also restores IE11-safe (`target: es5`) output for `.tsx` files in the client bundle — that silently regressed to ES2020 during the earlier TypeScript 5 migration (tsconfig's `target` had to move off `es5` for tsgolint, but the webpack loader's own target should have stayed pinned).
   - `ts-node`/`ts-node-dev` → `tsx`. `@swc-node/register` was the obvious pick for loading `webpack.config.ts`, but its peer range on `typescript` is `>=4.3 <7` — it would have blocked this very bump. `tsx` (esbuild-based) has zero dependency on the `typescript` package.
   - Note: `tsx watch` has no equivalent of `ts-node-dev`'s `--poll` flag (added originally for Docker filesystem-event reliability) — worth watching if `docker-dev` hot-reload gets flaky.

2. **`fix: resolve module-evaluation-order bugs the compiler swap exposed`** — real, pre-existing fragility in application/test code, not compiler config. TSC's specific CommonJS emit style (getter-based re-exports using `var`, not hoisting imports above interspersed statements) happened to paper over these; SWC's emit (spec-closer to real ESM) does not:
   - A genuine circular import in the `redirect` module (extracted `RedirectType`/`RedirectResult` into a new `redirect/types.ts`; import siblings directly instead of through a self-referencing barrel).
   - `jest.spyOn` on a named export — SWC compiles exports as non-configurable getters (real ESM semantics), so `spyOn` can never redefine them. Rewrote via `jest.mock` + `jest.requireActual`.
[...]

PR #2531, opened 2026-08-06
chore(deps): migrate package manager from npm to pnpm@11.20.0, bump Node 18 to 24

adriangohjw:
## ⚠️ Sequencing callout

**This PR bumps the app's Node target from 18 to 24 (`engines.node`, `.nvmrc`, Dockerfiles, CI, Gitpod) — do not merge until infra (Elastic Beanstalk / Lambda runtime, any other Node-18-pinned infra) is upgraded to run Node 24.** pnpm@11.20.0 requires Node >=22.13, which is why the Node bump is bundled in here rather than a separate PR — pnpm 10.x (Node 18-compatible) was tried first and reverted once we decided to bump Node instead. See commit history for that back-and-forth.

Also noticed in passing, **not changed here**: `serverless.yml` pins `runtime: nodejs14.x` for the Lambda functions — already older than even the previous Node 18 target. Worth its own follow-up.

## Summary
- Replaces `package-lock.json` with `pnpm-lock.yaml`, pins `pnpm@11.20.0` via `packageManager`, and adds `pnpm-workspace.yaml` with explicit `allowBuilds` entries for native modules (`bcrypt`, `sharp`, `esbuild`, `@swc/core`, `unix-dgram`) that pnpm v10+ blocks by default.
- `nodeLinker: hoisted` in `pnpm-workspace.yaml`: the app relies on npm's classic flat/hoisted `node_modules` in a few undeclared places (an ambient `@types/react` resolution, webpack's bare `querystring` import) — hoisted linking matches npm's resolution instead of pnpm's stricter isolated default, so the migration doesn't need to hunt down every such case individually. Also declares `@types/react`, `@types/history`, `@types/body-parser` explicitly as devDependencies since they were previously only ever pulled in transitively.
- Bumps Node 18 → 24 across `engines.node`, `.nvmrc`, `Dockerfile`/`Dockerfile.dev` (`node:24-alpine3.24`), `.gitpod.Dockerfile`, and CI (drops the now-redundant Node 22 override on the lint/format jobs, bumps the deploy job's `setup-node`).
- CI: composite setup action now installs via `pnpm/action-setup` and caches the pnpm store; replaces the `lockfile-lint` job (npm/yarn-only, can't parse `pnpm-lock.yaml`) with `pnpm audit signatures` (needs pnpm >=11.1.0, satisfied now that we're on 11.20.0); replaces the deploy job's ad-hoc unpinned `npm install -f serverless-plugin-include-dependencies@5.0.0` with a frozen-lockfile `pnpm install` (that plugin is already a pinned devDependency).
- Updates husky pre-push, `docker-compose.yml`, Gitpod config, and README instructions to use pnpm.

Stacked on `chore/migrate-to-typescript-7` since that branch was already the up-to-date head with ~20 unmerged commits at the time of this work.

## Test plan
- [x] `pnpm install --frozen-lockfile`
- [x] `pnpm run typecheck`
- [x] `pnpm run lint` / `pnpm run format:check`
- [x] `pnpm run build`
- [x] `pnpm run test` (58 suites, 481 passed / 4 skipped)
- [ ] `docker build -f Dockerfile.dev .` — **not verified**, local Docker daemon was broken in the environment this was built in. Please confirm the Alpine/musl native module builds (bcrypt, sharp) succeed before merging.
- [ ] Confirm infra (EB/Lambda) is on Node 24 before merging — see callout above
- [ ] CI green on this PR (lint/format/typecheck/build/test/testcafe/integration/audit-signatures jobs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

This PR migrates dependency management from npm to pnpm and raises the application toolchain from Node 18 to Node 24.
- Replaces the npm lockfile and install commands with pnpm equivalents across CI, Docker, Gitpod, local tooling, and documentation.
- Pins the package manager, configures hoisted dependency linking and native dependency builds, and updates runtime targets.
- Replaces lockfile linting with pnpm signature auditing and installs deployment dependencies from the frozen lockfile.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

The PR appears safe to merge once the explicitly documented Node 24 infrastructure prerequisite is satisfied.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| .github/actions/setup/action.yml | Migrates shared CI setup to pnpm with frozen-lockfile installation and store caching through setup-node v4. |
| .github/workflows/build-and-deploy.yml | Migrates deployment installation to pnpm and Node 24. |
| .github/workflows/ci.yml | Converts CI commands to pnpm and replaces npm-specific lockfile linting with pnpm signature auditing. |
| Dockerfile | Moves the production image to Node 24 and installs, builds, prunes, and starts the application through pnpm. |
| Dockerfile.dev | Moves the development image to Node 24 and pnpm-based dependency installation. |
| package.json | Updates the Node engine, pins pnpm, and declares dependencies needed under the new package-manager layout. |
| pnpm-workspace.yaml | Configures hoisted linking and permits required native package build scripts. |
| pnpm-lock.yaml | Introduces the pnpm dependency lockfile replacing package-lock.json. |

</details>

<sub>Reviews (5): Last reviewed commit: ["feat(deps): bump Node 18 to 24, restore ..."](https://github.com/opengovsg/gogovsg/commit/46a03a8167ec582c0b8378fa64b7731b1f0881ed) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50865480)</sub>


seaerchin (commented):

**Verdict: LOW-MODERATE RISK** — foundational and honestly sequenced, but leaves a known gap that's already biting another PR.

This is the PR that most of the rest of this queue's Node-version-sensitive bumps depend on (react-router-dom v7 needs Node ≥20, connect-redis@10 needs ≥22, nanoid@6 needs ^22/^24/≥26, etc.). The author is explicit and correct about sequencing: *"do not merge until infra (Elastic Beanstalk / Lambda runtime...) is upgraded to run Node 24."* `pnpm@11.20.0` requiring Node ≥22.13 is the actual reason the Node bump is bundled with the package-manager migration rather than split out — also disclosed, not hidden (pnpm 10.x was tried first and reverted).

- **Real, still-open gap:** the PR body itself notes *"not changed here: `serverless.yml` pins `runtime: nodejs14.x` for the Lambda functions... worth its own follow-up."* That gap is exactly what's now causing a real production blocker in the sibling PR #2533, whose `aws-sdk` v3 migration requires Node ≥20 in that same Lambda function. These two PRs need coordinated handling — either this PR (or a fast-follow immediately after it) needs to bump the Lambda runtime too, or #2533 needs to hold until that's done.
- **Not verified locally, flagged by the author:** Docker build wasn't exercised (broken daemon in the sandbox) — confirm the Alpine/musl native module builds (`bcrypt`, `sharp`) actually succeed on Node 24 before merging.
- `nodeLinker: hoisted` is a deliberate, reasonable choice to avoid hunting down every phantom-dependency case individually during the migration, rather than an oversight.

**Before merging:** confirm all downstream infra (EB, Lambda) is actually on Node 24 first, coordinate the Lambda runtime bump with #2533, and run a real `docker build` to confirm native module compilation on Alpine.

Dockerfile
```diff
@@ -33,18 +33,19 @@ RUN wget https://github.com/IBM/plex/blob/master/packages/plex-sans/fonts/comple
...
 COPY . ./
 
 RUN { \
   echo "Building..."; \
-  npm run build; \
+  pnpm run build; \
   echo "Removing devDependencies for production..."; \
-  npm prune --production; \
+  pnpm prune --prod; \
   }
```
dcshzj, line 48: **Blocking — the production image cannot be built from this branch.**

`pnpm prune` delegates to the install routine and re-runs the root `prepare` script, *after* it has just removed husky as a devDependency. From an actual `docker build` of 16e7ba2:

```
#14 23.96 devDependencies: skipped
#14 24.91 . prepare$ husky install
#14 24.93 . prepare: sh: husky: not found
#14 24.97 [ELIFECYCLE] Command failed.
ERROR: failed to build: ... exit code: 1
```

This is specifically an npm→pnpm behaviour change: `npm prune --production` removes husky and never runs `prepare`, which is why the old Dockerfile was fine.

Worth flagging that PR CI being green is misleading here — `build-and-deploy.yml` builds and pushes this image, so this fails every staging/production deploy, not just a local build.

The `set -e` below is a second, independent issue: `sh -c '{ false; true; }'` exits 0, so only the *last* command in this block decides the layer's status. A failing `pnpm run build` currently wouldn't fail the build, and the image would ship with a stale or missing `build/`.

Verified fix — prune runs to completion, husky is removed, prod deps (including the built sharp/bcrypt binaries) stay intact:

```suggestion
RUN set -e; { \
  echo "Building..."; \
  pnpm run build; \
  echo "Removing devDependencies for production..."; \
  pnpm prune --prod --ignore-scripts; \
  }
```

.github/workflows/ci.yml
```diff
@@ -14,54 +14,44 @@ jobs:
...
-      # Only needs package-lock.json on disk, not node_modules -- skip the
-      # full setup action's npm ci.
-      - uses: actions/setup-node@v4
-        with:
-          node-version-file: 'package.json'
-      - run: npx lockfile-lint --type npm --path package-lock.json --validate-https --allowed-hosts npm
+      # lockfile-lint doesn't understand pnpm-lock.yaml's format (only npm/yarn) --
+      # pnpm's own signature verification is the native equivalent.
+      - uses: ./.github/actions/setup
+      - run: pnpm audit signatures
```
dcshzj, line 54: Two independent regressions relative to the `lockfile-lint` job this replaces.

**1. Verification now runs after installing.** `./.github/actions/setup` runs `pnpm install --frozen-lockfile`, which executes dependency build scripts (bcrypt, sharp, esbuild, @swc/core) before anything has been checked. The old job deliberately skipped the install for exactly this reason. Good news: it's unnecessary — `pnpm audit signatures` reads the lockfile only. I ran it in a clean checkout with no `node_modules` and it audited all 1751 entries.

**2. Coverage isn't equivalent.** `lockfile-lint --validate-https --allowed-hosts npm` enforced *where* packages may come from. Signature auditing verifies provenance of whatever is listed, and skips registries that publish no signing keys — so a dependency pointing at an unapproved host can pass.

The obvious like-for-like replacement isn't available, which is worth knowing before someone tries: `pnpm-lock.yaml` v9 records no tarball URLs for registry dependencies. In 16,283 lines there are exactly 7 URL occurrences, all inside `deprecated:` message strings — there is no host to allowlist. The equivalent invariant is that every entry resolves to a *registry integrity hash*; git, tarball, directory and file sources all produce structurally different `resolution:` shapes.

Suggestion below. I tested the check both ways: it passes 1751/1751 on this lockfile, and fails with an annotation when I inject a `type: git` source or an `http://` tarball.

```suggestion
  supply-chain:
    name: Supply Chain
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v2
      # Deliberately does NOT use ./.github/actions/setup: verifying the lockfile
      # must happen before any install executes dependency build scripts.
      - uses: pnpm/action-setup@v4
      - uses: actions/setup-node@v4
        with:
          node-version-file: 'package.json'

      # Replaces lockfile-lint's --validate-https --allowed-hosts npm, which read
      # package-lock.json's `resolved` URLs. pnpm-lock.yaml v9 has no such URLs, so
      # the equivalent invariant is that every entry resolves to a registry
      # integrity hash -- git/tarball/directory/file sources fail this check.
      - name: Enforce registry-only dependency sources
        run: |
          total=$(grep -cE '^ +resolution: \{' pnpm-lock.yaml)
          registry=$(grep -cE '^ +resolution: \{integrity: sha(512|256|1)-' pnpm-lock.yaml)
          echo "$registry/$total lockfile entries resolve to a registry integrity hash"
          if [ "$total" -ne "$registry" ]; then
            grep -nE '^ +resolution: \{' pnpm-lock.yaml \
              | grep -vE 'resolution: \{integrity: sha(512|256|1)-' \
              | sed 's/^/::error::non-registry dependency source: /'
            exit 1
          fi

      # Verifies each lockfile entry's registry signature. Lockfile only -- no
      # node_modules required.
      - run: pnpm audit signatures
```

Two notes on that suggestion:
- To get the *approved host* property back as well, add a committed `.npmrc` containing `registry=https://registry.npmjs.org/`, so the registry those integrity hashes are fetched from is pinned rather than ambient. I left the corresponding assertion out of the suggestion so that applying it as-is doesn't fail on a missing file.
- The job rename is safe: `develop` currently has no required status checks configured, so nothing is pinned to the `Lockfile Lint` or `Audit Signatures` context.

.github/workflows/build-and-deploy.yml
```diff
@@ -106,10 +106,13 @@ jobs:
...
       - uses: actions/checkout@v2
 
+      - uses: pnpm/action-setup@v4
+
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: '18.x'
+          node-version: '24.x'
+          cache: 'pnpm'
```
dcshzj, line 115: `cache: 'pnpm'` is silently ignored here. `actions/setup-node@v1` accepts only `always-auth`, `node-version`, `registry-url`, `scope` and `version` — there is no `cache` input, so this job gets no caching at all.

v1 also still declares the `node12` action runtime. The runners are already force-upgrading `node20` actions (this run warns about `actions/checkout@v2`, `actions/setup-node@v4` and `pnpm/action-setup@v4` being pushed onto Node 24), so v1 is well past that line.

Since `engines.node` is now the single source of truth, this can also stop hardcoding the version:

```suggestion
      - name: Use Node.js
        uses: actions/setup-node@v4
        with:
          node-version-file: 'package.json'
          cache: 'pnpm'
```

pnpm-workspace.yaml
```diff
@@ -0,0 +1,18 @@
...
+  '@datadog/native-appsec': false
+  '@datadog/native-iast-taint-tracking': false
+  '@datadog/native-metrics': false
+  '@swc/core': true
+  aws-sdk: false
+  bcrypt: true
+  core-js: false
+  esbuild: true
+  sharp: true
+  unix-dgram: true
```
dcshzj, line 18: `unix-dgram: true` can never succeed in either Alpine image — there's no python3, so node-gyp fails during every build:

```
node_modules/unix-dgram install$ node-gyp rebuild
node_modules/unix-dgram install: gyp ERR! find Python
node_modules/unix-dgram install: gyp ERR! configure error
```

It's an optional dependency of hot-shots, and `src/server/util/dogstatsd.ts` constructs `new StatsD({ prefix: 'go.', mock: DEV_ENV })` with no `protocol: 'uds'`, so the Unix-socket transport is never used. Allowing the build only adds a wall of gyp errors to every image build and CI install.

```suggestion
  unix-dgram: false
```

dcshzj (commented):
Reviewed this by building it rather than reading it — ran the production `docker build`, `serverless package`, and the supply-chain checks against 16e7ba2. One blocking issue, two CI regressions and one no-op setting, all inline.

**Things that check out, for the record:**
- Alpine/musl native builds succeed: `bcrypt install: Done`, `sharp install: Done` (libvips downloaded, integrity check passed). That's the unchecked box in the test plan.
- `pnpm run build` succeeds inside the image (webpack compiled, 3 size warnings only).
- Lambda packaging is unaffected by `nodeLinker: hoisted` — `serverless package` produces all four zips and `serverless-plugin-include-dependencies` resolves cleanly (only the usual missing-optional-dependency warnings), with sharp's files included.
- `pnpm audit signatures` verifies 1751/1751 packages; nothing is skipped today.

**Two items I can't anchor inline because the files aren't in this diff:**

1. `serverless.yml` — all four `package.patterns` blocks still list `'package-lock.json'`, which this PR deletes. I confirmed the packaged zips now contain root `package.json` but no lockfile, so the pattern is dead config. Worth switching to `'pnpm-lock.yaml'` here since it's a direct consequence of the rename.

2. `runtime: nodejs14.x` — agree with keeping it out of this PR. For the follow-up: create has been blocked since 2024-01-09, but *update* isn't blocked until 2027-03-03, so existing functions still deploy today — this isn't on fire. Target `nodejs24.x` (supported to 2028-04-30); not `nodejs20.x`, which is already deprecated as of 2026-04-30.

**On the failing check:** `End To End Tests` fails identically on base PR #2530 — same `Directory Page` assertions, only the generated slugs differ — and `Integration Tests` passes on both. It's inherited from the base, not caused by this PR.

.github/workflows/ci.yml
dcshzj: Correction to my own suggestion above — **do not apply it as written, it had a false-negative hole.** Fixed in 486874a7.

I had an independent pass done over these comments, and it found that the invariant I proposed ("every entry resolves to a registry integrity hash") is *not* sufficient. A non-registry tarball keeps its integrity hash and just gains a `tarball:` key:

```
resolution: {integrity: sha512-..., tarball: https://codeload.github.com/evil/pkg/tar.gz/deadbeef}
```

Because `integrity:` is still the first key, my `\{integrity: sha512-` pattern matched it and counted it as clean — I measured this directly against the real lockfile: the injected codeload entry passed, `1751` matched-as-ok. Probing pnpm for the other source types gives `{commit:, repo:, type: git}` for git deps, `{directory:, type: directory}` for local ones, and `{gitHosted: true, integrity:, tarball:}` for GitHub shorthand.

So the check has to assert the resolution is *exactly* an integrity hash and nothing else, rather than that it starts with one. What landed anchors the whole line:

```
ok=$(grep -cE '^ +resolution: \{integrity: sha(512|256|1)-[A-Za-z0-9+/]+={0,2}\}$' pnpm-lock.yaml)
```

Verified: passes 1751/1751 on the current lockfile, and fails on all four evasions above (integrity-first tarball, gitHosted tarball, git source, local directory).

Related, and the reason this check is worth having separately from the audit step: `pnpm audit signatures` reports that codeload tarball as `1 package has a verified registry signature`, exit 0. Non-registry sources are counted as **verified**, not skipped — so its pass message is actively misleading, which is a bit worse than the "narrower coverage" I described in the original comment.

dcshzj:
Pushed fixes for the review comments above (41e74901..8258f484), split one concern per commit so anything you disagree with can be dropped on its own:

| Commit | What |
|---|---|
| 41e74901 | `pnpm prune --prod --ignore-scripts` + `set -e` — unbreaks the production image build |
| 486874a7 | lockfile source check before any install, `.npmrc` registry pin, prod image built in PR CI |
| c54e0e00 | deploy job to `setup-node@v4` so `cache: 'pnpm'` actually applies |
| ec731ce9 | `unix-dgram: false` |
| 8258f484 | `serverless.yml` packages `pnpm-lock.yaml` |

I had an independent pass run over my own comments before pushing. It confirmed the substance of all of them, but corrected me on two points:

**1. The `set -e` part was mis-scoped.** I flagged it in the Dockerfile comment without noting that the base branch has the identical `RUN { … }` structure — it's pre-existing, not something this migration introduced. It's also currently unreachable, because the block's last command fails today regardless; it only becomes a live risk once the prune step succeeds. Still worth having, but it isn't a defect of this PR and I shouldn't have implied it was.

**2. I overstated the audit-signatures ordering severity.** Seven other jobs in this same `ci.yml` run the same `pnpm install` in the same CI run, both before and after this PR, so the install-before-verify ordering never actually protected anything at the workflow level. Worth fixing on its own merits — verification that doesn't need `node_modules` shouldn't ask for it — but it wasn't the hole I made it sound like. The genuinely important half is the coverage gap, which I've corrected in the thread above.

**Three things that turned up while verifying, none of which I'd raised:**

1. **`pnpm prune --prod` is not a local prune.** It performs a registry install — I measured 23 packages added and ~120 tarballs downloaded, including wrong-platform optional bindings, plus a transient socket error with retries. So the production image build now needs npm-registry access at that step, where `npm prune --production` needed none. Worth knowing for build reliability; it also makes that layer noticeably slower to commit.

2. **Lambda package size regression:** `bulk-qrcode-generation.zip` goes 21.41 MB → 30.47 MB (4561 → 5025 files, 109.82 → 130.50 MB unzipped). This was verified to be a hoisting/dedup artifact rather than version drift — both trees contain the same 1424 package names with exactly one version difference. Still inside Lambda's 50 MB zipped / 250 MB unzipped limits, but the headroom roughly halves. Not blocking, worth being aware of before adding anything else to that function.

3. **Two Gitpod risks I could not confirm without launching a workspace**, so treat as unverified: `pnpm add -g maildev` in `.gitpod.yml` fails in a clean corepack pnpm environment with no `PNPM_HOME` (`the configured global bin directory is not in PATH / Run "pnpm setup"`), and the follow-on `$(which maildev)` depends on the same thing; and `.gitpod.Dockerfile` activates corepack as root before `USER gitpod`, while corepack's version cache is per-user.

Two notes on what I pushed:
- The `.npmrc` pin assumes registry.npmjs.org. If OGP fronts npm with a mirror or proxy anywhere, that line needs adjusting rather than the check being dropped.
- I did not get a full production image build to complete locally, so the new `Docker build (production image)` job on this PR is the first end-to-end validation. What was verified in-container is the specific fix: `pnpm prune --prod --ignore-scripts` exits 0, top-level `node_modules` drops 1088 → 455 with husky/oxlint/webpack/typescript/jest removed, and `sharp`, `bcrypt`, `express` and `hot-shots` all still load afterwards.

Unrelated to your changes, restating for the record: `End To End Tests` fails identically at the base commit, so it isn't yours to fix here.

seaerchin:
2 issues flagged by adversarial review:

> - High — Node 24 migration leaves deployed Lambdas on Node 14. [package.json:50](https://github.com/opengovsg/GoGovSG/blob/8258f4845036585b5825b6217bbcaa1e75a2395b/package.json#L50) declares Node 24 while [serverless.yml:6](https://github.com/opengovsg/GoGovSG/blob/8258f4845036585b5825b6217bbcaa1e75a2395b/serverless.yml#L6-L8) still deploys nodejs14.x. The deployment job installs and packages the newly resolved dependency tree under Node 24, then deploys it to Node 14. Dependencies or generated code may therefore load successfully in CI but fail during Lambda initialization or invocation. The PR’s own sequencing warning recognizes this incompatibility but leaves it in the shipped configuration. Update the Serverless runtime and Lambda infrastructure as part of the coordinated rollout, or separate the Node/package-manager migration until every runtime is compatible. Add an invocation smoke test against the actual target runtime. Confidence: 0.99

doing it in a follow up PR

> - Medium — The replacement supply-chain check permits any attacker-controlled HTTPS registry. [.github/workflows/ci.yml:67](https://github.com/opengovsg/GoGovSG/blob/8258f4845036585b5825b6217bbcaa1e75a2395b/.github/workflows/ci.yml#L67-L86) claims to preserve the old npm-host allowlist, but line 85 only checks that .npmrc begins with registry=https://. A dependency-changing PR can point .npmrc at an arbitrary HTTPS registry and regenerate integrity hashes against packages served there. Registry signatures do not restore the trust boundary if verification metadata is also obtained from that registry. Assert the exact approved registry URL—or a small explicit allowlist—and reject scoped registry overrides as well. Confidence: 0.91

valid, but i think no risk for now

seaerchin (approved):

PR #2533, opened 2026-08-06
chore(deps): fix Node-18-blocked advisories 

adriangohjw:
## Problem

Three dependencies had open security advisories that couldn't be fixed while this repo was pinned to Node 18 (resolved on `feat(deps): bump Node 18 to 24`, commit `935db8b5`):

- `sharp` (GHSA-f88m-g3jw-g9cj, libvips CVEs) — the fix landed in `0.35.0`, which requires Node ≥20.9.0
- `file-type` (GHSA-5v7r-6r5c-r473, infinite-loop DoS) — the fix landed in `21.3.1`, which requires Node ≥20
- `aws-sdk` (GHSA-j965-2qgj-vjmq, region-validation) — this one was never actually blocked by Node 18; the advisory's "patched version ≥3.0.1" refers to the entirely different, non-drop-in `@aws-sdk/*` package family. `aws-sdk` (the v2 package) never published a v3 and is now fully end-of-support with no further fixes coming.

## Solution

**`sharp` 0.32.6 → 0.35.3**: only `resize().png()/.jpeg().toBuffer()` are used — none of the removed/renamed APIs across 0.33/0.34/0.35 (trim options, background metadata, sharpen properties, the `jp2k`→`jp2` rename) apply here.

**`file-type` 16.5.4 → 21.3.1**: `file-type` went pure ESM at v17 (dropped the default export and CommonJS entirely). `FileTypeFilterService` now uses a dynamic `import('file-type')` with the renamed `fileTypeFromBuffer` export — confirmed working via a plain Node smoke test and the `tsc`-compiled build output. Jest's resolver can't load a pure-ESM package with no `require` export condition, so the test now virtual-mocks `file-type`'s `fileTypeFromBuffer` instead of exercising the real library — the test's job is verifying our fallback-to-extension logic, not `file-type`'s own detection heuristics.

**`aws-sdk` v2 → `@aws-sdk/*` v3** (S3 + SQS, the only two AWS surfaces this repo uses):

- `.putObject()`/`.putObjectAcl()`/`.upload()` + `.promise()` → `.send(new PutObjectCommand(...))`/`.send(new PutObjectAclCommand(...))` and `@aws-sdk/lib-storage`'s `Upload` class (same `httpUploadProgress` event, `.done()` in place of `.promise()`)
- `.sendMessage().promise()` → `.send(new SendMessageCommand(...))`
- `s3ForcePathStyle` → `forcePathStyle`; `httpOptions: { timeout }` → `requestHandler: { requestTimeout }`
- Removing `aws-sdk` also removed its transitive `querystring` shim, which the **client** bundle depended on incidentally — two client action files `import querystring from 'querystring'` directly. Added an explicit `querystring-es3` fallback in `webpack.config.ts` — this was caught by an actual production build, not by tests, so it's worth noting as the one surprising side effect of this migration.
- Test mocks for S3 moved from spying on named v2 methods to spying on `.send()` and dispatching by command class. This surfaced a real, unrelated bug in the existing test file: two `describe` blocks each called `jest.spyOn(s3Client, 'send').mockImplementation(...)` at their top level, which runs during Jest's *collection* phase — the second call silently overwrote the first for the whole file, so one block's assertions were checking calls that never reached its own mock. Moved both into `beforeEach` so each block's mock is actually the active one when its own tests run.

## Security advisories (`pnpm audit --prod`)

**Before this branch (i.e. on `chore/harden-supply-chain`) → after this branch:**

| Severity | Before | After |
|---|---|---|
| High | 1 (`sharp`) | 0 |
| Moderate | 2 (`file-type`, transitive `uuid` — already fixed on the parent branch via override) | 0 |
| Low | 1 (`aws-sdk`) | 0 |
| **Total** | **4** | **0** |

`pnpm audit --prod` now reports **"No known vulnerabilities found."**

## Tests

- `pnpm run typecheck`, `pnpm run build`, `pnpm run lint`, `pnpm run format:check` — all pass
- `pnpm run test` — 58 suites / 481 tests pass, including the updated `aws.test.ts` and `UrlRepository.test.ts`
- `pnpm audit --prod` — 0 vulnerabilities (down from 4)
- Manual smoke tests: `sharp` PNG render, `file-type` real detection via plain Node `import()`, `QrCodeService`'s existing test (exercises `sharp` end-to-end)
- **Recommend a staging deploy check** before promoting to prod for the S3/SQS surfaces specifically — file uploads (`uploadFileToS3`, the bulk QR-code zip upload) and the OTP/bulk-QR SQS queue are worth exercising for real, since AWS SDK v3's request signing/behavior can differ subtly from v2 in ways unit tests with mocked clients won't catch.

## Before & After Screenshots

N/A — no UI changes.

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any new dependencies, new scripts, etc._

**New dependencies**:

- `@aws-sdk/client-s3` : replaces `aws-sdk`'s `S3` client
- `@aws-sdk/client-sqs` : replaces `aws-sdk`'s `SQS` client
- `@aws-sdk/lib-storage` : replaces `aws-sdk`'s `S3.upload()` multipart helper

**Removed dependencies**:

- `aws-sdk` : fully removed, no longer maintained upstream

**New dev dependencies**:

- `querystring-es3` : webpack browser-bundle polyfill for Node's `querystring` module, previously satisfied only incidentally by `aws-sdk` v2's own transitive dependency of the same package name

**Config changes**:

- `webpack.config.ts` : new `resolve.fallback.querystring` entry
- `.github/dependabot.yml` : new `aws-sdk` group (`@aws-sdk/*` — shares one version number across the family, same pattern as the existing `webpack`/`babel` groups)

The PR upgrades the vulnerable `sharp` and `file-type` packages and migrates the application's S3 and SQS integrations from AWS SDK v2 to modular AWS SDK v3 clients.
- Replaces v2 S3 and SQS operations with v3 commands and multipart uploads.
- Adapts file-type detection to the package's pure-ESM API.
- Updates dependency metadata, mocks, and the browser query-string fallback.
- The previously reported bulk-worker runtime incompatibility remains unresolved.
</details>

<details><summary><h3>Confidence Score: 4/5</h3></summary>

The PR is not safe to merge until the bulk QR-code Lambda runtime is upgraded to support its newly imported dependencies.

The bulk worker remains configured for Node 14 while its cold-start import graph now includes AWS SDK v3 and sharp packages requiring Node 20 or newer, so queued bulk QR jobs cannot initialize the handler.

**Files Needing Attention:** serverless.yml and src/server/serverless/bulk-qrcode-generation/s3.js
</details>

| Filename | Overview |
|----------|----------|
| src/server/serverless/bulk-qrcode-generation/s3.js | Migrates bulk-worker uploads to AWS SDK v3, but the handler remains deployed on an incompatible Node 14 runtime. |
| src/server/services/aws.ts | Migrates application S3 operations to AWS SDK v3 commands and the multipart Upload helper. |
| src/server/services/sqs.ts | Migrates queue dispatch to SQSClient and SendMessageCommand. |
| src/server/modules/threat/services/FileTypeFilterService.ts | Uses dynamic import and the renamed pure-ESM `fileTypeFromBuffer` API. |
| package.json | Replaces AWS SDK v2 and upgrades sharp and file-type, introducing Node 20+ requirements relevant to the separately deployed worker. |
| webpack.config.ts | Explicitly supplies the browser querystring fallback previously available through a transitive dependency. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant App as Main application
    participant SQS as AWS SQS
    participant Worker as Bulk QR Lambda
    participant S3 as AWS S3
    App->>SQS: SendMessageCommand(job batch)
    SQS->>Worker: Invoke on Node 14
    Worker--xWorker: Load Node 20+ dependencies
    Note over Worker: Cold start fails before processing
    Worker--xS3: Artifacts are not uploaded
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(deps): disable clock-skew correction..."](https://github.com/opengovsg/gogovsg/commit/03301e956ed11c14559b6e9df769bad2a28995ba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50865447)</sub>

**Context used:**

- Knowledge Base — [Background jobs and bulk operations](https://app.greptile.com/opengovsg/-/custom-context/knowledge-base/opengovsg/gogovsg/-/docs/background-jobs-and-bulk-operations.md)


seaerchin (commented):

**Verdict: BREAKING CHANGES LIKELY — real, currently-unresolved production blocker.**

**Top concern, already flagged by another automated reviewer (Greptile) and appears unresolved:** the bulk-QR-code Lambda function (`src/server/serverless/bulk-qrcode-generation/s3.js`, `serverless.yml`) still runs on `nodejs14.x` with no runtime override, but this PR's `aws-sdk` v2→v3 migration means that Lambda's cold-start import path now loads `@aws-sdk/client-s3`/`@aws-sdk/lib-storage`, both of which require Node ≥20. That function will fail at cold start in production unless the Lambda runtime is bumped alongside this change (or before it).

Setting that blocker aside, the rest of the migration work is genuinely careful:
- `file-type` 16→21 correctly handles the package going pure-ESM at v17 — uses a dynamic `import()` with the renamed `fileTypeFromBuffer` export, confirmed via a plain Node smoke test; since Jest can't load a pure-ESM package with no `require` condition, the test was changed to virtual-mock `file-type` rather than exercise the real library (a reasonable, disclosed trade-off).
- `aws-sdk` v2→v3: found a genuinely subtle, production-only bug during a real build — removing `aws-sdk` also removed its transitive `querystring` shim that the **client** bundle depended on incidentally; fixed with an explicit `querystring-es3` webpack fallback. Also fixed a real pre-existing test bug (two `describe` blocks' top-level `jest.spyOn(s3Client, 'send')` silently overwrote each other during Jest's collection phase).
- `sharp` 0.32→0.35: only uses `resize().png()/.jpeg().toBuffer()`, none of the removed APIs — low risk on its own (see also the separate sharp-focused PR #2505, which correctly flags the same Node ≥20.9 requirement).
- `pnpm audit --prod` goes from 4 vulnerabilities (1 high, 2 moderate, 1 low) to 0 — real security value here.

**Before merging:** fix the Lambda runtime version (or add a runtime override) before this can safely ship — this isn't optional, it's a production cold-start failure waiting to happen. Also do the author's own recommended staging deploy check for the S3/SQS surfaces (file uploads, bulk QR zip, OTP/bulk-QR SQS queue) since AWS SDK v3's signing behavior can differ subtly from v2 in ways mocked unit tests won't catch.

seaerchin (approved):
lgtm - this seems like a bump to compat version so i didn't check for breaking changes apart from those addressed (prop style)

seaerchin (changes requested):
need to update serverless to node 24+

seaerchin (approved):
lgtm again, the serverless issue should be addressed in another PR

